PR #54, opened 2023-06-09
Add information about republishing messages

kriszyp:
This updates the configuration documentation for the new setting. Presently we are only treating this as a potential option for clustering performance issues, but if further testing reveals more of a need for this, could potentially add the clustering instructions themselves.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -221,6 +221,10 @@ Path to the private key file.
 
 When true, will skip certificate verification. For use only with self-signed certs.
 
+`republishMessages` - _Type_: boolean; _Default_: true
+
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -221,6 +221,10 @@ Path to the private key file.
 
 When true, will skip certificate verification. For use only with self-signed certs.
 
+`republishMessages` - _Type_: boolean; _Default_: true
+
+When true, all transactions that are received from other nodes are republished to this node's stream. When pub/sub routes are not fully connected between all nodes, this ensures that messages are routed to all nodes through intermediate nodes. This also ensures that all writes, whether local or remote, are written to the NATS transaction log. However, there is additional overhead with republishing, and setting this is to false can provide better clustering performance. When false, you need to ensure all pub/sub routes are fully connected between every node to every other node, and be aware that the NATS transaction log will only consist of local writes.
```
DavidCockerill, line 226: Do you using 'routes' could be confusing here? routes are also used to describe/define nats networking (not sure how to best describe them) connections, that are separate from subscriptions.

kriszyp: "When subscriptions are not fully connected between all nodes"
Does that sound right?

DavidCockerill: Yes, I think that is better. Maybe also change the "When false, you need to ensure all pub/sub routes" part to subscriptions also

kylebernhardy: can we change: `can provide better clustering performance` to `can provide better data replication performance`?

PR #55, opened 2023-06-15
DOCS-11: added clustering.logLevel to the config page

terraHDB:
this pr adds a newer, previously undocumented config setting to the config file page in our documentation: clustering.logLevel. I also took this opportunity to remove deprecated log levels: fatal and notify from the logging section.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -433,7 +433,7 @@ Control the verbosity of logs.
 logging:
   level: error
 ```
-There exists a log level hierarchy in order as `trace`, `debug`, `info`, `warn`, and `error`. When the level is set to `trace` logs will be created for all possible levels. Whereas if the level is set to `warn`, the only entries logged will be `warn` and `error`. The default value is `error`.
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -422,7 +433,7 @@ Control the verbosity of logs.
 logging:
   level: error
 ```
-There exists a log level hierarchy in order as `trace`, `debug`, `info`, `warn`, `error`, `fatal`, and `notify`. When the level is set to `trace` logs will be created for all possible levels. Whereas if the level is set to `fatal`, the only entries logged will be `fatal` and `notify`. The default value is `error`.
+There exists a log level hierarchy in order as `trace`, `debug`, `info`, `warn`, and `error`. When the level is set to `trace` logs will be created for all possible levels. Whereas if the level is set to `warn`, the only entries logged will be `warn` and `error`. The default value is `error`.
```
DavidCockerill, line 436: What was your thinking for changing this paragraph? We still have fatal and notify logs in hdb. This is not the same for clustering log level - there aren't as many log levels for that.

terraHDB: oh I was looking in the natsTerms.js, instead of hdbTerms.js. my fault. will change back

DavidCockerill (approved):

PR #70, opened 2023-09-11
CORE-2252 Add docs for clone node fully connected

DavidCockerill:


docs/cloning.md
```diff
@@ -65,6 +65,13 @@ logging:
 
 *Note: any required configuration needed to install/run HarperDB will be default values or auto-generated unless it is provided in the config file.* 
 
+#### Fully connected clone
+A fully connected clone is when in addition to setting up replication between the leader and the clone,
```
kriszyp, line 69: How about let's define fully-connected topology first:

A fully connected topology is when all nodes are replicating (publish and subscribing) with all other nodes. A fully connected clone maintains this topology with addition of the new node. In addition to setting up replication between....

DavidCockerill: I like that, Thank you!

DavidCockerill: done

PR #71, opened 2023-09-11
DOCS-22 Break Up Config Page

terraHDB:
This PR breaks up the large configuration page into more navigable page. 

These changes are easily viewable with this link https://docs.harperdb.io/copy-of-harperdb-1/


DavidCockerill:
I understand "split things up over multiple pages" point of view... conversely, I also like the one scrollable page. I find it easy to do finds on. Do you think there is any chance gitbook supports something like this, where is has links to sections on the page?
https://www.mongodb.com/docs/manual/reference/configuration-options/

DavidCockerill:
Did someone request it get broken up?

One thing to note is the work I am currently doing (removing CF section + combining some params) will affect this.

PR #73, opened 2023-10-04
CORE-2164 Updates to 4.2 docs after 'schema' to 'database' naming cha…

DavidCockerill:
I think there are probably more locations in the docs where schema should be updated to 'database' but i figured this is a good chunk of them.


kriszyp (approved):
Lot's of great updates!

PR #75, opened 2023-10-18
CORE-2167 doc updates after adding support for defining table with pr…

DavidCockerill:
…imary_key


DavidCockerill:
@kriszyp I think we need to make some updates to # Dynamic Schema Tables in dynamic-schema.md but looking at it im not sure its as easy and replacing all the 'schema' for 'database'. Can you please take a look at this section. Thanks

PR #77, opened 2023-10-27
Release 4.2 reorg

kriszyp:


kriszyp (commented):
I don't necessarily think we should be using the top level categories for folders at all. Currently this is kind of inconsistent; a lot of the standard categories (Clustering, Jobs, SQL Guide) are top level folders, but there is a bunch inside of /developers. I think we should eliminate the /developers folder and just use top level categories (like "Developers") as headers in SUMMARY.md for consistency, what do you think?

Intro page:  Cool!
I think the Application entry page probably need a little more direction on how to install and REST entry page a little more info on getting schema set up. I kinda of originally wrote those pages with the idea of a sequentially reading Getting Started -> Applications -> REST page. However, your set up is great, and adding a little more to intros rather than assuming this sequence is an improvement. I can certain ly do this.

What is your intention for https://docs.harperdb.io/docs/custom-functions/host-static and its sibilings?
https://docs.harperdb.io/harperdb-4.2-docs-reorg/OjUcvHiTvuINkwuqpXqR/custom-functions/host-static 404s; there are a few of the custom-functions pages that logically redict to component, but I had originally intended to leave the other pages (without reasonale redirects) in place (although there has to be some link to them in SUMMARY.md, even if buried)

jcohen-hdb:
My thoughts with the top level folders is to embrace GitBook's folder structure approach. When creating subtitles, pages, etc in GitBook that's how it handles them. I like sticking with that where possible because GitBook will automatically update all the files whenever paths have to change. The only reason I pulled the other ones out is to not break existing links. I do see your point in being consistent though. But I'd vote for categorized folders per GitBook's organization method for consistency, then add in a ton of redirects to go from 4.1 unorganized to 4.2 organized. Thoughts here? 

Totally agreed on the adding info part. My goal so far has been to get what we have organized. Then next week we can edit existing content and add new pages, etc. To me, that's easier and won't come with breaking changes like moving things around.

For the Custom Functions pages, I wasn't sure if they should stay or go. In this repo I deleted them, because I didn't have that context. Would they fall under the [Fastify routes pages](https://docs.harperdb.io/harperdb-4.2-docs-reorg/OjUcvHiTvuINkwuqpXqR/developers/applications/define-routes)? They're good content, I just wasn't sure if they were still current, so if we see use for them, we should certainly keep them.

kriszyp:
For the Custom Functions pages, I had set up redirects (they are still there) for the pages I thought could go to the new components pages. Moving the remaining content into the fastify routes page sounds good to me, I think. And then we just need to add redirects for those remaining CF pages (to the fastify routes page, presumably).

jcohen-hdb:
Awesome, thanks for updating that. I think we're in good shape then. As long as SUMMARY.md is correct, they will still render in GitBook, so we're good there. It makes sense to me that people are working out of their IDEs, so in that case I think everything is fine.

I'm going to continue updating things today, but those are more minor changes. I'll work in the release_4.2 branch for those, since it'll be mostly spelling fixes, small rewording, etc. If I find I need larger changes I'll make a new branch. Thanks again!

PR #79, opened 2023-11-07
Add release notes for patches

kriszyp:


jcohen-hdb:
I think there are a couple of extra things that have to happen before these will show up. We need to update the Release Notes landing page to link to these and we need to update SUMMARY.md so the links show up in the navigation. I'm happy to do that on this branch. I'll also make a note to make this more clear in the Docs Best Practices guide, because it's definitely not right now.

kriszyp:
Yeah, I actually did know that. And yet I was lazy enough to file this PR anyway. Thank you for finishing that! (I can certainly do it, if I get a chance to do it before you)

jcohen-hdb:
I just made the updates, I'm going to go ahead an approve myself (which is probably not the best practice.

kriszyp:
Thank you!

PR #82, opened 2023-11-15
Clean up API docs

DavidCockerill:


jcohen-hdb (approved):
I made a couple of quick corrections and I think we caught everything. I'll approve now and we can do a quick review on GitBook to validate.

PR #97, opened 2024-03-13
Resource api updates

kriszyp:


jcohen-hdb (approved):
All of this makes sense and is very informative, thank you!

PR #102, opened 2024-03-27
CORE-2333 update clone node docs to reflect clone changes

DavidCockerill:


jcohen-hdb (approved):
I made a few spelling/grammar updates. The makes sense and reads well. Good work!

PR #106, opened 2024-04-03
Updating old Studio docs for Applications

jcohen-hdb:
Our Studio docs were out of date which often lead to confusion for users who were editing Applications via Studio. My goal for this page was to tailor it to a user who will be approaching Applications editing from the Studio perspective instead of the typical developer perspective that the Applications docs cover. Feel free to make any update or just let me know if there's stuff you'd like me to change. Thank you!


jcohen-hdb:
Making the executive decision that this is much better than what we have right now, so I'm going to merge it. Take a look if you to verify. Thank you!

PR #110, opened 2024-05-15
Add docs for configuring multi-certificate SNI TLS config

kriszyp:


kylebernhardy (approved):

docs/deployments/configuration.md
```diff
@@ -792,20 +797,21 @@ Path to the private key file.
...
-`user` - _Type_: string; _Default_: Common Name
-
-This configures a specific username to authenticate as for mTLS connections. If a `user` is defined, any authorized mTLS connection (that authorizes against the certificate authority) will be authenticated as this user.
-This can also be set to `null`, which indicates that no authentication is performed based on the mTLS authorization. When combined with `required: true`, this can be used to enforce that users must have authorized mTLS _and_ provide credential-based authentication.
+```yaml
+tls:
+  - certificate: ~/hdb/keys/certificate1.pem
+    certificateAuthority: ~/hdb/keys/ca1.pem
+    privateKey: ~/hdb/keys/privateKey1.pem
+    host: example.com # the host is optional, and if not provided, this certificate's common name will be used as the host name.
```
mbranyon, line 807: Did we want to make this a list as well, or do we just have multiple entries of the same cert if it is to handle more than one hostname?

kriszyp: Oh, totally forgot about that. I will add that in a second PR.

kriszyp: @mbranyon If no host is specified, should we read the alt names in addition to the common name as the default list of matching hosts? I assume we should.

mbranyon: I would agree with that

docs/deployments/configuration.md
```diff
@@ -792,20 +797,21 @@ Path to the private key file.
...
-You can also define specific mTLS options by specifying an object for mtls with the following (optional) properties which may be included:
+If you want to define multiple certificates that are applied based on the domain/host requested via SNI, you can define an array of TLS configurations. Each configuration can have the same properties as the root TLS configuration, but can (optionally) also have an additional `host` property to specify the domain/host that the certificate should be used for:
 
-`user` - _Type_: string; _Default_: Common Name
-
-This configures a specific username to authenticate as for mTLS connections. If a `user` is defined, any authorized mTLS connection (that authorizes against the certificate authority) will be authenticated as this user.
-This can also be set to `null`, which indicates that no authentication is performed based on the mTLS authorization. When combined with `required: true`, this can be used to enforce that users must have authorized mTLS _and_ provide credential-based authentication.
+```yaml
+tls:
+  - certificate: ~/hdb/keys/certificate1.pem
```
mbranyon, line 804: I just tested this specific bit out. HarperDB did not like using `~`. I'm worried that there may be confusion in the docs around having to use the full path since they differ only slightly (`/home/harperdb` in container vs. `/home/ubuntu` on the host).

it looks like the relative path seems to work:
```
  - certificate: hdb/keys/certificate1.pem
```

a-ronjohnson (approved):

docs/deployments/configuration.md
kriszyp: PR for that https://github.com/HarperDB/harperdb/pull/2130

PR #112, opened 2024-08-02
Add ssh key operation docs

Devin-Holland:


docs/developers/operations-api/components.md
```diff
@@ -285,3 +285,170 @@ _Operation is restricted to super_user roles only_
...
+{
+  "operation": "list_ssh_keys"
+}
+```
+
+### Response: 200
+```json
+[
+    {
+        "name": "harperdb-private-component.key"
```
kriszyp, line 406: In this example, the name listed here ("harperdb-private-component.key") doesn't match the name of the SSH key in the add_ssh_key ("harperdb-private-component"). Is that typo? Or are they actually inconsistent with each other?

docs/developers/operations-api/components.md
```diff
@@ -285,3 +285,170 @@ _Operation is restricted to super_user roles only_
...
+    },
+    ...
+]
+```
+
+## Set SSH Known Hosts
+Sets the SSH known_hosts file. This will overwrite the file.
```
kriszyp, line 413: Should we document that this operation is not necessary for github.com since we auto-add the right known hosts in that case?

docs/developers/operations-api/components.md
Devin-Holland: This was because of a typo by me in my core PR. I updated that which makes it output without `.key` so i updated this to reflect. Both should be without the `.key`

docs/developers/operations-api/components.md
Devin-Holland: It's documented here:

https://github.com/HarperDB/documentation/pull/112/files#diff-76288cf2a040af5d1685e726557b17dfa51f8a7a1cd089d61be80b871de7a7f4R300

But I can also add it there if you think it makes sense to be in both.

PR #117, opened 2024-09-18
created debugging-harperdb.md file

Keyology:
Hey @kriszyp  @jcohen-hdb, 

I created a PR to push this forward please let me know if there are any changes you would like made.

### DIFF
--- a/docs/deployments/debugging-harperdb.md
+++ b/docs/deployments/debugging-harperdb.md
@@ -1,0 +1,44 @@
+# Debugging HarperDB
+
+If your HarperDB servers are experiencing performance issues, memory leaks, or errors, enabling debugging can help diagnose the problem. To start debugging, ensure that the HarperDB configuration has debugging enabled.
+
+To learn more, please visit: [HarperDB Configuration - Threads](https://docs.harperdb.io/docs/deployments/configuration#threads)
+
+In most deployments, debugging is already enabled. If it's set up, you can connect to HarperDB using Node DevTools, available via `chrome://inspect` in Chrome (click "Open dedicated DevTools for Node").
+
+## Port Configuration for Debugging
+
+Since DevTools can only connect to one thread per port, we assign specific ports for debugging:
+
+- Main thread: Port 9229
+- Worker 1: Port 9230
+- Worker 2: Port 9231
+- And so on for additional worker threads.
+
+To debug multiple threads, you'll need to register each port in DevTools:
+
+1. In the "Connections" tab, click "Add connection."
+2. Add `localhost:9230`, `localhost:9231`, and any other necessary ports.
+
+## Connecting via SSH with Tunneling
+
+When debugging a specific HarperDB instance, you’ll need to SSH into the instance using port tunneling. Here's how to connect to worker threads 1 and 2 (typically, debugging the main thread isn't needed since it doesn’t handle traffic):
+
+```bash
+ssh -L 9230:localhost:9230 -L 9231:localhost:9231 root@cool-customer.harperdbcloud.com
+```
+
+Once logged in, your DevTools should be able to connect to these threads for debugging.
+
+## Debugging Tips
+
+- Enable Pretty Printing: When debugging, be sure to enable pretty printing (bottom left of the code screen). This makes it easier to set breakpoints, even in HarperDB’s obfuscated code.
+
+- Breakpoints in Application Code: If you're debugging your own application, just open the relevant file and set breakpoints as usual.
+
+
+## Investigating Performance
+
+- Run a Performance Profile: To investigate performance issues, start by running a profile for about 10-30 seconds. This will help you identify bottlenecks when analyzing the results.
+
+- Check Memory Usage: For memory-related issues, take a heap snapshot to spot any large memory allocations. You can also compare heap usage over time to track memory growth or leaks.


docs/deployments/debugging-harperdb.md
```diff
@@ -0,0 +1,44 @@
+# Debugging HarperDB
```
jcohen-hdb, line 1: I think somewhere up here we need to add some context around what kinds of debugging this is referring to. @kriszyp This information would be specifically for self-managed deployments, correct?

docs/deployments/debugging-harperdb.md
```diff
@@ -0,0 +1,44 @@
...
+
+If your HarperDB servers are experiencing performance issues, memory leaks, or errors, enabling debugging can help diagnose the problem. To start debugging, ensure that the HarperDB configuration has debugging enabled.
+
+To learn more, please visit: [HarperDB Configuration - Threads](https://docs.harperdb.io/docs/deployments/configuration#threads)
+
+In most deployments, debugging is already enabled. If it's set up, you can connect to HarperDB using Node DevTools, available via `chrome://inspect` in Chrome (click "Open dedicated DevTools for Node").
```
jcohen-hdb, line 7: Would this be true in "most deployments" or is that referring to how we do our deployments? Not sure if this conflicts with what is in our config or not.

docs/deployments/debugging-harperdb.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Port Configuration for Debugging
+
+Since DevTools can only connect to one thread per port, we assign specific ports for debugging:
+
+- Main thread: Port 9229
```
jcohen-hdb, line 13: https://docs.harperdb.io/docs/deployments/configuration#threads
This shows debug.port as 9249? Also it mentions startingPort, which is not in the sample config. Not sure if that's directly relevant to this, but seems like it.

docs/deployments/debugging-harperdb.md
```diff
@@ -0,0 +1,44 @@
...
+
+1. In the "Connections" tab, click "Add connection."
+2. Add `localhost:9230`, `localhost:9231`, and any other necessary ports.
+
+## Connecting via SSH with Tunneling
+
+When debugging a specific HarperDB instance, you’ll need to SSH into the instance using port tunneling. Here's how to connect to worker threads 1 and 2 (typically, debugging the main thread isn't needed since it doesn’t handle traffic):
+
+```bash
+ssh -L 9230:localhost:9230 -L 9231:localhost:9231 root@cool-customer.harperdbcloud.com
```
jcohen-hdb, line 28: I don't think we'd want `harperdbcloud.com` examples in the public docs, but that may be my misunderstanding.

jcohen-hdb (changes requested):
@kriszyp I'd definitely like your eyes on this as well. A lot of this stuff isn't really my wheelhouse, hence all the questions.

@Keyology In order to make this page visible in the navigation, you'll need to add it to the [SUMMARY.md](https://github.com/HarperDB/documentation/blob/debugging-Harper-db/docs/SUMMARY.md) file

docs/deployments/debugging-harperdb.md
kriszyp: Yes, this is for remote connection to servers that you have SSH access to and access to the ports.

docs/deployments/debugging-harperdb.md
kriszyp: You are correct Jake, this is definitely not true for most deployments, it is only for our deployments, for our engineers.
In 4.4, with dev defaults on install, debug is enabled, but not with a starting port range, as is assumed in this doc.
If you are running inside a docker container, you have to enable a range of ports, which is somewhat of an elaborate process itself.

docs/deployments/debugging-harperdb.md
kriszyp: We should update those docs to be 9229, that is much more standard. But this should be clarified that this is connects to the ports configured in the starting port range.

kriszyp (commented):
I would think that potentially more interesting content/approach would be looking at different IDEs (VSCode, WebStorm, Devtools), and how to configure them, with follow-up steps for what needs to be configured for each (for example, you don't need starting range for VSCode and Webstorm, but you do for Devtools).
Unfortunately, taking a post that was intended for a complete different audience, and turning into docs, isn't a great way to write docs, sorry you got dragged into this.

docs/deployments/debugging-harperdb.md
```diff
@@ -0,0 +1,44 @@
...
+
+- Main thread: Port 9229
+- Worker 1: Port 9230
+- Worker 2: Port 9231
+- And so on for additional worker threads.
+
+To debug multiple threads, you'll need to register each port in DevTools:
+
+1. In the "Connections" tab, click "Add connection."
+2. Add `localhost:9230`, `localhost:9231`, and any other necessary ports.
```
Ethan-Arrowood, line 21: How would someone know how many threads they are using? Is there a way to get dev tools to connect to _all_ threads without manually adding each connection?

docs/deployments/debugging-harperdb.md
```diff
@@ -0,0 +1,44 @@
...
+- Enable Pretty Printing: When debugging, be sure to enable pretty printing (bottom left of the code screen). This makes it easier to set breakpoints, even in HarperDB’s obfuscated code.
+
+- Breakpoints in Application Code: If you're debugging your own application, just open the relevant file and set breakpoints as usual.
+
+
+## Investigating Performance
+
+- Run a Performance Profile: To investigate performance issues, start by running a profile for about 10-30 seconds. This will help you identify bottlenecks when analyzing the results.
+
+- Check Memory Usage: For memory-related issues, take a heap snapshot to spot any large memory allocations. You can also compare heap usage over time to track memory growth or leaks.
```
Ethan-Arrowood, line 44: Lets include some links here about how to analyze a performance profile or heap snapshot results.

PR #120, opened 2024-10-07
GraphQL Querying Docs

Ethan-Arrowood:
Adding some docs for the upcoming graphql querying feature (https://github.com/HarperDB/harperdb/pull/2226) 

Important note here, the Coming Soon! sections are not going to be completed in the first feature PR. We will continue to add to this over time. Furthermore, this is distinctly separate from the GraphQL schema declaration feature.

### DIFF
--- a/docs/technical-details/reference/graphql.md
+++ b/docs/technical-details/reference/graphql.md
@@ -41,7 +41,7 @@ Accept: application/graphql-response+json
 
 > Tip: For the best user experience, include the `Accept: application/graphql-response+json` header in your request. This provides better status codes for errors.
 
-The HarperDB GraphQL querying system is strictly limited to HarperDB Resources. Queries can only specify HarperDB Resources and their attributes in the selection set. Queries can filter using [arguments](https://graphql.org/learn/queries/#arguments) on the top-level Resource field. HarperDB provides a short form pattern for simple queries, and a long form pattern based off of the [Resource Query API](./resource.md#query) for more complex queries.
+The HarperDB GraphQL querying system is strictly limited to exported HarperDB Resources. For many users, this will typically be a table that uses the `@exported` directive in its schema. Queries can only specify HarperDB Resources and their attributes in the selection set. Queries can filter using [arguments](https://graphql.org/learn/queries/#arguments) on the top-level Resource field. HarperDB provides a short form pattern for simple queries, and a long form pattern based off of the [Resource Query API](./resource.md#query) for more complex queries.
 
 Unlike REST queries, GraphQL queries can specify multiple resources simultaneously:
 
@@ -153,7 +153,7 @@ fragment sharedFields on Any {
 
 ### Short Form Querying
 
-Any Resource attribute can be used as an argument for a query. In this short form, multiple arguments is treated as multiple equivalency conditions with the default `and` operation.
+Any attribute can be used as an argument for a query. In this short form, multiple arguments is treated as multiple equivalency conditions with the default `and` operation.
 
 For example, the following query requires an `id` variable to be provided, and the system will search for a `Dog` record matching that id.
 
--- a/docs/technical-details/reference/graphql.md
+++ b/docs/technical-details/reference/graphql.md
@@ -22,11 +22,11 @@ query GetDogs {
 }
 ```
 The `GET` request would look like:
-```
+```sh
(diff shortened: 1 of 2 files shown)


Ethan-Arrowood:
Note: make edits to this doc: https://docs.harperdb.io/docs/developers/applications#querying

docs/technical-details/reference/graphql.md
```diff
@@ -0,0 +1,235 @@
...
+curl 'http://localhost:9926/graphql?query=query+GetDogs+%7B+Dog+%7B+id+name+%7D+%7D'
+```
+And the `POST` request would look like:
+```sh
+curl 'http://localhost:9926/graphql' \
+	-H 'Content-Type: application/json' \
+	-d '{ "query": "query GetDogs { Dog { id name } } }" '
+```
+
+The HarperDB GraphQL querying system is strictly limited to HarperDB Resources. Queries can only specify HarperDB Resources and their attributes in the selection set. Queries can filter using [arguments](https://graphql.org/learn/queries/#arguments) on the top-level Resource field. HarperDB provides a short form pattern for simple queries, and a long form pattern based off of the [Resource Query API](./resource.md#query) for more complex queries.
```
DavidCockerill, line 33: Is `resources` the best word here? Applications/components? or maybe say any table that is defined with a graphql schema?

docs/technical-details/reference/graphql.md
```diff
@@ -0,0 +1,235 @@
...
+
+Unlike REST queries, GraphQL queries can specify multiple resources simultaneously:
```
DavidCockerill, line 35: Same kind of thinking as above. Would it make more sense to use 'table' rather than 'resources `Unlike REST queries, GraphQL queries can specify multiple tables simultaneously`

docs/technical-details/reference/graphql.md
```diff
@@ -0,0 +1,235 @@
...
+
+fragment sharedFields on Any {
+	id
+	name
+}
+```
+
+### Short Form Querying
+
+Any Resource attribute can be used as an argument for a query. In this short form, multiple arguments is treated as multiple equivalency conditions with the default `and` operation.
```
DavidCockerill, line 145: Is a resource attribute different from a table attribute?

docs/technical-details/reference/graphql.md
```diff
@@ -0,0 +1,235 @@
...
+query GetDogs {
+	Dog {
+		id
+		name
+	}
+}
+```
+The `GET` request would look like:
+```sh
+curl 'http://localhost:9926/graphql?query=query+GetDogs+%7B+Dog+%7B+id+name+%7D+%7D'
```
kriszyp, line 24: I think it is a little easier to read HTTP requests in standard HTTP format, and less tech-specific (curl). I'd expect most modern front-end devs to actually be more familiar with `fetch` than curl. And plain HTTP is what we have used elsewhere in the REST docs.

docs/technical-details/reference/graphql.md
kriszyp: "Resource" is technically a correct definition of what can be queried here. However, it is more specifically the _exported_ resources  (a table is one type of resource, but other resources can be exported). And it might be helpful to include a note or parenthetical that this includes `@export`'ed tables from the schema or resources.js.

docs/technical-details/reference/graphql.md
kriszyp: No, since a table is a resource. However, it may be less likely to confuse to just say "Any attribute", or "Any field".

kriszyp (commented):
Looking great, nice work!

docs/technical-details/reference/graphql.md
DavidCockerill: out of curiosity, how would you query a resource that isn't a table resource?

docs/technical-details/reference/graphql.md
Ethan-Arrowood: Scratch that, I like the look of http format better

docs/technical-details/reference/graphql.md
Ethan-Arrowood: I've included a new line here that may provide some additional clarity.

docs/technical-details/reference/graphql.md
Ethan-Arrowood: Fixed

DavidCockerill (approved):

kriszyp (approved):

PR #123, opened 2024-10-30
Keoni edits to resource md changes 10 25 24

Keyology:
@kriszyp  Please review my changes and let me know your thoughts


docs/deployments/debugging-harperdb.md
```diff
@@ -0,0 +1,44 @@
...
+
+If your HarperDB servers are experiencing performance issues, memory leaks, or errors, enabling debugging can help diagnose the problem. To start debugging, ensure that the HarperDB configuration has debugging enabled.
+
+To learn more, please visit: [HarperDB Configuration - Threads](https://docs.harperdb.io/docs/deployments/configuration#threads)
+
+In most deployments, debugging is already enabled. If it's set up, you can connect to HarperDB using Node DevTools, available via `chrome://inspect` in Chrome (click "Open dedicated DevTools for Node").
+
+## Port Configuration for Debugging
+
+Since DevTools can only connect to one thread per port, we assign specific ports for debugging:
```
kriszyp, line 11: This is delivery specific, this is not the default config for HarperDB.

docs/deployments/debugging-harperdb.md
```diff
@@ -0,0 +1,44 @@
...
+
+1. In the "Connections" tab, click "Add connection."
+2. Add `localhost:9230`, `localhost:9231`, and any other necessary ports.
+
+## Connecting via SSH with Tunneling
+
+When debugging a specific HarperDB instance, you’ll need to SSH into the instance using port tunneling. Here's how to connect to worker threads 1 and 2 (typically, debugging the main thread isn't needed since it doesn’t handle traffic):
+
+```bash
+ssh -L 9230:localhost:9230 -L 9231:localhost:9231 root@cool-customer.harperdbcloud.com
```
kriszyp, line 28: Customer of who?

kriszyp (changes requested):
The resources.md updates look to good to me, this seems like some helpful reorganization.
The debugging-harperdb.md simply does not belong in this repository, the document is for entirely different audience with an entirely different set of assumptions than HarperDB users.
If you want to add more information about debugging, I would add some information to the existing debugging.md documentation (is there an intuitive reason why users would expect to have look in two different debugging docs?), and describe _when_ and _how_ they should configure HarperDB to use port ranges (when you need to connect to different threads in devtools), and how debugging differs across different IDEs.
We can provide a link to Node.js's documentation for SSH tunneling which is much better for general purpose usage. The tips and performance profiling info could go in that document as well. This all should be in a separate PR.

kriszyp:
Also, for the resource.md documentation, a great follow-up would be to see if we could break it up into two documents.

PR #146, opened 2025-02-25
update resource extension options reference

Ethan-Arrowood:
Rel: https://github.com/HarperDB/harperdb/pull/2472


docs/developers/components/built-in.md
```diff
@@ -104,22 +104,57 @@ rest:
...
+For example, with the configuration option: `files: 'web/**/*'`, and a file structure like:
+
+```
+web/
+|- a.txt
+|- foo/
+|  |- b.txt
+```
+
+The pattern will match `['web/a.txt', 'web/foo', 'web/foo/b.txt']`. The extension will serve `/a` file, and then the `/foo` directory. When it receives the `/foo/b` file, it will determine it is already serving all files within `/foo` and skip it. However, a better configuration value would be `files: 'web/**/*.txt'`.
```
kriszyp, line 139: Does `web/**/*` have an observable negative impact? If you wanted all file types to be served, that would be correct (or just as correct as `web/**`)?

kriszyp (approved):

docs/developers/components/built-in.md
Ethan-Arrowood: Hm I need to revisit this section. Those patterns are actually equivalent as far as I can tell.

kriszyp (approved):

kriszyp (approved):

PR #179, opened 2025-06-06
initial data loader documentation

heskew:


kriszyp (approved):
Fantastic!
Would you mind adding a link to this from 4.6.0.md release notes? (and if there is anything else you want to say about it in the release notes)
(And we have a release_4.6 branch now, so we would like to target that for 4.6 stuff now, but we can just cross-merge for the time being)

PR #186, opened 2025-06-26
Improve component documentation

Ethan-Arrowood:
This is a small step towards improving component documentation. It mainly reorganizes pages to properly put the reference docs under the technical-details/reference section. It has special language intended for 4.6 specifically detailing how there is a new experimental Plugin API. It updates the Application guide section so it includes important distinctions between Components, Applications, and Extensions.

I've complete removed the "Components" section from the Developers part of the docs. I think encompassing it in Applications is great as it immediately details the differences and then links out to technical reference as needed. 

I think we should add a "Plugin" section under Developers that walks the developer through building a custom plugin, and then integrating it with the application they built.

### DIFF
--- a/.gitbook.yaml
+++ b/.gitbook.yaml
@@ -70,11 +70,11 @@ redirects:
   custom-functions: developers/applications/README.md
   custom-functions/define-routes: developers/applications/define-routes.md
   custom-functions/using-npm-git: developers/custom-functions/create-project.md
-  custom-functions/custom-functions-operations: developers/components/operations.md
+  custom-functions/custom-functions-operations: developers/operations-api/README.md
   custom-functions/debugging-custom-function: developers/applications/debugging.md
   custom-functions/example-projects: developers/applications/example-projects.md
-  add-ons-and-sdks: developers/components/README.md
-  add-ons-and-sdks/google-data-studio: developers/components/google-data-studio.md
+  add-ons-and-sdks: developers/applications/README.md
+  add-ons-and-sdks/google-data-studio: developers/miscellaneous/google-data-studio.md
   sql-guide: developers/sql-guide/README.md
   sql-guide/features-matrix: developers/sql-guide/features-matrix.md
   sql-guide/insert: developers/sql-guide/README.md


heskew:
> I think we should add a "Plugin" section under Developers that walks the developer through building a custom plugin, and then integrating it with the application they built.

I would have loved that when I first dug into harper

docs/technical-details/reference/components/configuration.md
```diff
@@ -0,0 +1,85 @@
+# Component Configuration
+
+> For information on the distinction between applications and extensions, refer to beginning of the [Applications](../../../developers/applications) documentation section.
+
+Harper components are configured with a `config.yaml` file located in the root of the component module directory. This file is how an components configures other components it depends on. Each entry in the file starts with a component name, and then configuration values are indented below it.
+
+```yaml
+name:
```
cb1kenobi, line 8: Is there a general `config.yml` schema that we could show here? I think it would be helpful if we said here's the blueprint for a component config file. Maybe something like this?

```yaml
rest: true/false
extension: "path/to/extension.js"
"<component_name>":
	package: "<npm_package>"
	files: "<pattern>"
```

docs/technical-details/reference/components/managing-applications.md
```diff
@@ -1,20 +1,22 @@
-# Managing
+# Managing Applications
```
cb1kenobi, line 1: I wonder if this should be just "Applications"? We have "Extensions" and "Plugins" under this "Components" section, so seeing "Applications" could drive home that they are just another type of component.

docs/technical-details/reference/components/plugins.md
```diff
@@ -0,0 +1,475 @@
...
+This is a functional example of how the `handleComponent()` method and `scope` argument can be used to create a simple static file server plugin. This example assumes that the component has a `config.yaml` with the `files` option set to a glob pattern that matches the files to be served.
+
+> This is a simplified form of the [static](./built-in.md#static) built-in extension.
+
+```js
+export function handleComponent(scope) {
+  const staticFiles = new Map();
+
+  scope.options.on('change', (key, value, config) => {
+	if (key[0] === 'files' || key[0] === 'urlPath') {
```
kriszyp, line 32: Indentation is at 2, 3, 5 spaces, fun :).

kriszyp (approved):
This is fantastic, great job. I will merge this tomorrow for the release, happy to have any further edits come later.

docs/developers/applications/README.md
```diff
@@ -1,33 +1,81 @@
...
+  ApolloExt --> GraphqlSchema
+  ApolloExt --> Networking
+
+  GraphqlSchema --> Database
+  JsResource --> Database
+  Rest --> Networking
+
 ```
 
+> As of Harper v4.6, a new, **experimental** component system has been introduced called **plugins**. Plugins are a **new iteration of the existing extension system**. They are simultaneously a simplification and an extensibility upgrade. Instead of defining multiple methods (`start` vs `startOnMainThread`, `handleFile` vs `setupFile`, `handleDirectory` vs `setupDirectory`), plugins only have to define a single `handleComponent` method. Plugins are **experimental**, and complete documentation is available on the [plugin API](../../technical-details/reference/components/plugins.md) page. In time we plan to deprecate the concept of extensions in favor of plugins, but for now, both are supported.
```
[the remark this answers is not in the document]
Ethan-Arrowood: The nuance is that we aren't _just_ renaming extensions to plugins. Plugins are a new API. And for at least v4.6 (and likely rest of v4) we'll have both extensions and plugins.

docs/developers/applications/README.md
```diff
@@ -1,33 +1,81 @@
...
+
+> Harper maintains a number of custom components that are available on `npm` and `GitHub`, such as the [`@harperdb/nextjs`](https://github.com/HarperDB/nextjs) extension or the [`@harperdb/status-check`](https://github.com/HarperDB/status-check) application.
+
+Harper does not currently include any built-in applications, making "custom applications" a bit redundant. Generally, we just say "application". However, there is a multitude of both built-in and custom extensions, and so the documentation refers to them as such. A complete list of built-in extensions is available in the [Built-In Extensions](../../technical-details/reference/components/built-in.md) documentation page, and the list of custom extensions and applications is available below.
```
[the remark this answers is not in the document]
Ethan-Arrowood: https://github.com/HarperDB/harperdb/blob/main/server/operationsServer.ts#L162 this doesn't use Harper's static plugin - its just using fastify-static.

docs/technical-details/reference/components/built-in.md
```diff
@@ -1,26 +1,24 @@
-# Built-In Components
+# Built-In Extensions
```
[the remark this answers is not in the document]
Ethan-Arrowood: Nope you're not crazy, looks to me like GitBook uses the file name as the header instead of the provided text 🤦‍♂️

cb1kenobi (approved):

docs/technical-details/reference/components/configuration.md
Ethan-Arrowood: I think the difference in configurations between like root config, extensions/plugins, and applications is enough to just keep this generic. I'm trying to essentially explain yaml without explaining yaml haha. The point of this is that each "entry" is tab-delimited. Other parts of the docs actually dig into the specifics of what goes in to the file.

PR #204, opened 2025-07-24
Clarified host value explanation for add and update ssh key

dillardm89:
Attempted to deploy component using api and cli following docs and ran into github public key permission errors. Found that `package.json` used `git@github.com` when attempting to authenticate based on package url provided in `deploy_component` request. However, original `host` value instructions for `add_ssh_key` direct user to enter `private-repo-name.github.com` to create the config file, which would require calling `git@private-repo-name.github.com` instead in order to access the associated private key in config.

I can recreate and demo if needed for clarification.

If we specifically need to support `private-repo-name.github.com` as host in config then package name in `package.json` would need to be modified to match `git@private-repo-name.github.com` or a custom git script added to ensure proper key is used.

_Also removed a redundant section of docs for clarity_

### DIFF
--- a/docs/developers/operations-api/components.md
+++ b/docs/developers/operations-api/components.md
@@ -47,10 +47,12 @@ or this can be shortened to:
 HarperDB/application-template#semver:v1.0.0
 ```
 
-You can also install from private repository if you have an installed SSH keys on the server:
+You can also install from private repository if you have installed SSH keys on the server. Ensure the `host` portion of the url exactly matches the `host` used when adding ssh keys to ensure proper authentication.
 
 ```
 git+ssh://git@github.com:my-org/my-app.git#semver:v1.0.0
+
+git+ssh://git@harperdb-private-component.github.com:my-org/my-app.git#semver:v1.0.0
 ```
 
 Or you can use a Github token:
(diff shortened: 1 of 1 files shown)


dillardm89:
Thanks that makes perfect sense but was definitely not clear from the docs as-is.

I rolled back the ssh key changes to keep the harperdb-private-component host, kept some clarification in key documentation though, and add clarification in deploy_component to ensure user knows to match the host with the ssh key.

Devin-Holland (approved):

Devin-Holland:
Re-posting my message from before stripping some stuff out:

The reason we need to specify a different host than just `github.com` because it uses the `config` that the `add_ssh_key` operation builds to translate what it sees in the `package` of `deploy_component` to the correct key. That's why we use a custom `host` that should then get translated back to `github.com` (the `hostname`).

If you use `github.com` as the `host` then you can only deploy 1 private repository on an instance.

Can you give an example of what you used for the `add_ssh_key` and the `deploy_component` that didn't work? I think you are referencing a POC which i believe the commands should look something like:

```
{
    "operation": "add_ssh_key",
    "name": "my-poc",
    "key": "my\nssh\nkey\n",
    "host": "my-poc.github.com",
    "hostname": "github.com"
}
```

and then

```
{
    "operation": "deploy_component",
    "project": "my-poc",
    "package": "git+ssh://git@my-poc.github.com:HarperDB/my-poc.git#semver:v1.0.0"
}
```

I tested those 2 commands (i added a test deploy key to the repo) on a fresh install locally.

PR #214, opened 2025-08-07
get site:build running without errors and adding a tmp site:reset for simple cleanup

heskew:
`npm run site:dev` and `npm run site:build` should both run without errors with the latter having some left over link hash warnings that can be resolved later.


Ethan-Arrowood (approved):
works locally for me 👍

PR #223, opened 2025-08-11
Fix name of sort attribute key

cap10morgan:
This should be singular, right?


kriszyp (approved):

kriszyp:
I don't know where we are in the doc transition; typically we want this on the release_4.5 branch so it applies to 4.5+, but maybe don't need that now.

cap10morgan:
> docs content hasn't been migrated yet so it would need to be applied to each individual branch. if it can wait until after the content migration it could touch all relevant versions in a single pr...

Yeah, this is not super urgent stuff, so let me know what's easiest for you all w/r/t the platform transition.

PR #261, opened 2025-08-18
demo idea: restructure docs into section-based sidebars

ldt1996:
Changed the docs navigation so that each section has its own sidebar instead of everything being in one big list.

- added separate sidebars for:
  - getting started
  - developers
  - administration
  - deployments
  - technical details
- updated config so navbar items load the right sidebar for each section

Now when you click a top-level section, the sidebar only shows that section’s content. 

Please note I didn't update the sidebar for any other version than `4.6`. This is just a demo.

image


heskew:
@Ethan-Arrowood sidebar(s) can be expanded by default iirc

@ldt1996 I don't recall. :) I'll give it another clean try shortly

heskew:
just tried this out with the build artifact. still have the homepage problem. guessing it's not in a sidebar so there's no sidebar shown. might make sense to put the homepage in the getting started sidebar and removing the getting started index. 

I do like this direction. we'd have to work on the ux since the top nav gets too crammed before we flip to the nave menu. we'll also have `release notes` up there...

PR #267, opened 2025-08-19
s/search_(attribute|value)/$1/ in replication docs

cap10morgan:
H/T to @cb1kenobi for flagging


cb1kenobi (approved):

Ethan-Arrowood:
I believe this needs to be done to the `docs/` files too so it carries over to eventual 4.7 docs

heskew:
> I believe this needs to be done to the `docs/` files too so it carries over to eventual 4.7 docs

right... any thoughts on improving that workflow? just having versioned docs, i.e. cutting 4.7 from the 4.6 docs, would get us part of the way there...

PR #270, opened 2025-08-20
Update search_by_value props in docs/dev/replication too

cap10morgan:
Applying #267 to docs/ hierarchy too

### DIFF
--- a/docs/developers/replication/index.md
+++ b/docs/developers/replication/index.md
@@ -276,7 +276,7 @@ You can also check the configuration of the replication system, including the cu
 	"operation": "search_by_value",
 	"database": "system",
 	"table": "hdb_nodes",
-	"search_attribute": "name",
-	"search_value": "*"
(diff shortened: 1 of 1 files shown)


Ethan-Arrowood (approved):
Just to be clear, this change is only happening on 4.6 and eventually 4.7? Or is backported to 4.5? (If so, please update 4.5 docs as well).

cap10morgan:
> Just to be clear, this change is only happening on 4.6 and eventually 4.7? Or is backported to 4.5? (If so, please update 4.5 docs as well).

So there's no real change happening here. We've supported both sets of props for some time, but an amount of time I don't know exactly. So the further back in versions we go, the less confident I am about the change. All I'm really trying to do is encourage users of recent versions to use the newer, better, faster, stronger props. So IMO it's fine to just have this in 4.6+. But I'm happy to be overruled there. :)

PR #274, opened 2025-08-21
remove versioned release notes content, replacing with redirects to the non-versioned release notes

heskew:


Ethan-Arrowood:
Nice! So now each minor version has its appropriate release notes files? And then you are using redirects to map to them respectively? GitHub is having a hard time rendering so I'll have to checkout locally later or tomorrow to properly review

heskew:
annoyingly the local 'serve' mechanism won't behave the same way as github pages so it's not a perfect test. 

this is a cleanup followup pr for putting all release notes outside of /docs/ in a single place and working towards only needing to add a new file for any given version. most of this pr is to put the redirects in place of where the release notes used to be. the new location: https://github.com/HarperDB/documentation/tree/main/release-notes. should be easier to get towards automating nearly all of this part (if not completely)

Ethan-Arrowood (approved):

PR #310, opened 2025-09-10
Remove some out-of-date pages and reorganize relevant ones.

Ethan-Arrowood:
This deletes the `applications/example-projects` and `developers/miscellaneous` pages from 4.4 docs and onwards. 

The `developers/miscellaneous/query-optimization.md` was moved to the `reference` section and within the `Resource` section when present (4.6+).

Also removes some redirects no longer needed. 

The deleted files contained information that is severely out of date and no longer relevant. We did not maintain any of the example projects (many wouldn't even run on Harper today), google data studio is no longer maintained, and we don't maintain any of the "SDKs" listed; though we may consider adding something like that back in later on when we actually maintain things like that.


kriszyp (approved):
Yeah, outside of the Go SDK preservation that Wes noted, this looks good to me.

PR #316, opened 2025-09-15
Docs for caching page

nenharper:
The current caching page is more guide-like than a reference page, but still not relevant to the application guide and narrative at hand. 

This PR breaks this out into a documentations page and a reference page, and make the docs more narrative.

### DIFF
--- a/versioned_docs/version-4.6/developers/applications/caching.md
+++ b/versioned_docs/version-4.6/developers/applications/caching.md
@@ -1,292 +1,85 @@
 ---
 title: Caching
 ---
-
 # Caching
 
-Harper has integrated support for caching data from external sources. With built-in caching capabilities and distributed high-performance low-latency responsiveness, Harper makes an ideal data caching server. Harper can store cached data in standard tables, as queryable structured data, so data can easily be consumed in one format (for example JSON or CSV) and provided to end users in different formats with different selected properties (for example MessagePack, with a subset of selected properties), or even with customized querying capabilities. Harper also manages and provides timestamps/tags for proper caching control, facilitating further downstreaming caching. With these combined capabilities, Harper is an extremely fast, interoperable, flexible, and customizable caching server.
+In the [quickstart guide](../../getting-started/first-harper-app.md), you built a working Dog API in just a few minutes. That API is now ready to store and query data. But what happens when your application also needs to pull in data from other systems—say, a third-party service that provides dog breed information?
 
-## Configuring Caching
+If you hit that external API every time a user makes a request, you’ll pay the cost in speed, reliability, and maybe even money. That’s where Harper’s built-in caching comes in. Harper lets you cache external data in the same tables and APIs you’re already using, so your app feels fast, reliable, and cost-efficient without you writing glue code.
 
-To set up caching, first you will need to define a table that you will use as your cache (to store the cached data). You can review the [introduction to building applications](./) for more information on setting up the application (and the [defining schemas documentation](./defining-schemas)), but once you have defined an application folder with a schema, you can add a table for caching to your `schema.graphql`:
+## Step 1: Add a Cache Table
+Caching in Harper works just like creating any other table. Open up your `schema.graphql` and add a cache table alongside your `Dog` table:
 
 ```graphql
-type MyCache @table(expiration: 3600) @export {
-	id: ID @primaryKey
+type BreedCache @table(expiration: 3600) @export {
+  id: ID @primaryKey
+}
+```
+Here, `expiration: 3600` means cached entries expire after an hour. Harper will automatically evict old data and refresh it when needed. By exporting this table, you instantly get a REST API for it at `http://localhost:9926/BreedCache/`.
+
+## Step 2: Connect to an External Source
+Now let’s say you want to enrich your Dog records with breed details from an external API. Instead of hitting that API directly every time, we’ll connect it to the `BreedCache` table.
+
+Open `resources.js` and define a resource:
+
+```javascript
+class BreedAPI extends Resource {
+  async get() {
+    return (await fetch(`https://dog-api.com/${this.getId()}`)).json();
+  }
+}
+
+const { BreedCache } = tables;
+BreedCache.sourcedFrom(BreedAPI);
+```
+
+That’s it. When your app calls /BreedCache/husky, Harper will:
+
+- Check if “husky” is already in the cache.
+- If not (or if it’s expired), fetch it from https://dog-api.com/husky.
+- Store it in BreedCache so the next request is instant.
+
+Harper even prevents “cache stampedes”: if multiple users request the same breed at the same time, only one fetch goes out to the source.
+
+## Step 3: Use the Cache in Your API
+Now you can combine your Dogs with cached breed info. For example, imagine you added a “breed” attribute to your Dog table earlier:
+
+```graphql
+type Dog @table @export {
+  id: ID @primaryKey
+  name: String
+  breed: String @indexed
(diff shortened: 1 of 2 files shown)


versioned_docs/version-4.6/developers/applications/caching.md
```diff
@@ -1,292 +1,85 @@
...
-	async post(data) {
-		if (data.invalidate)
-			// use this flag as a marker
-			this.invalidate();
-	}
+type BreedCache @table(expiration: 3600) @export {
+  id: ID @primaryKey
 }
 ```
+Here, `expiration: 3600` means cached entries expire after an hour. Harper will automatically evict old data and refresh it when needed. By exporting this table, you instantly get a REST API for it at `http://localhost:9926/BreedCache/`.
```
Ethan-Arrowood, line 18: Harper just marks data as expired. It will not optimistically refresh data. The next request from a user to an expired record will result in the refetch and cache being restored.

versioned_docs/version-4.6/developers/applications/caching.md
```diff
@@ -1,292 +1,85 @@
...
-	subscribe() {
-		const subscription = super.subscribe();
-		setupListeningToRemoteService().on('update', (event) => {
-			subscription.send(event);
-		});
-		return subscription;
-	}
+class BreedAPI extends Resource {
+  async get() {
+    return (await fetch(`https://dog-api.com/${this.getId()}`)).json();
```
Ethan-Arrowood, line 28: Lets break this up into a couple lines for clarity

Ethan-Arrowood: Does dog-api actually exist? Remember this is meant to be a functional follow-along guide. Can we find a simple 3rd party resource to use here? Or is this copied from somewhere else?

versioned_docs/version-4.6/developers/applications/caching.md
```diff
@@ -1,292 +1,85 @@
...
+BreedCache.sourcedFrom(BreedAPI);
 ```
 
-### Subscribing to Caching Tables
+That’s it. When your app calls /BreedCache/husky, Harper will:
 
-You can subscribe to a caching table just like any other table. The one difference is that normal tables do not usually have `invalidate` events, but an active caching table may have `invalidate` events. Again, this event type gives listeners an opportunity to choose whether or not to actually retrieve the value that changed.
+- Check if “husky” is already in the cache.
+- If not (or if it’s expired), fetch it from https://dog-api.com/husky.
+- Store it in BreedCache so the next request is instant.
```
Ethan-Arrowood, line 40: Aye this is the correct behavior i was talking about above.

versioned_docs/version-4.6/developers/applications/caching.md
```diff
@@ -1,292 +1,85 @@
...
-Here both the update to the post and the update to the comments will be atomically/transactionally committed together with the same timestamp.
+You can enrich a Dog API request by querying both Dog and BreedCache. The Dog table is your source of truth, while BreedCache ensures you never block on slow or flaky external APIs.
 
-## Cache-Control header
+## Step 4: Keep Your Cache Fresh
+By default, caches are passive: they refresh only when requested and expired. Sometimes that’s fine. Other times, you’ll want them to stay in sync as soon as data changes at the source. Harper supports both:
 
-When interacting with cached data, you can also use the `Cache-Control` request header to specify certain caching behaviors. When performing a PUT (or POST) method, you can use the `max-age` directive to indicate how long the resource should be cached (until stale):
+- Passive caching: great for data that doesn’t change often (like dog breed characteristics).
+- Active caching: connect to a subscription or webhook so Harper gets notified when the source changes. Your cache updates immediately, and your users always see fresh data.
```
Ethan-Arrowood, line 62: We do? How?

versioned_docs/version-4.6/developers/applications/caching.md
```diff
@@ -1,292 +1,85 @@
...
+You can even invalidate records manually, or implement write-through caching so updates flow both ways.
 
-You can use the `only-if-cached` directive on GET requests to only return a resource if it is cached (otherwise will return 504). Note, that if the entry is not cached, this will still trigger a request for the source data from the data source. If you do not want source data retrieved, you can add the `no-store` directive. You can also use the `no-cache` directive if you do not want to use the cached resource. If you wanted to check if there is a cached resource without triggering a request to the data source:
+## Step 5: Layer Caches for Your Users
 
-```http
-GET /my-resource/id
-Cache-Control: only-if-cached, no-store
-```
+Caching doesn’t stop at Harper. Because every Harper cache table is a REST API, your clients can cache responses downstream too. Harper automatically tags responses with ETag headers so browsers or edge caches can hold onto data longer. That means your app is fast all the way down—from the source, to Harper, to the client.
```
Ethan-Arrowood, line 68: Not longer but for the "appropriate amount of time"

versioned_docs/version-4.6/developers/applications/caching.md
Ethan-Arrowood, line 1: This is good progress. But what about a non-third-party caching example? Like what if I have expensive work to do on the requested resource and I want to be caching that? This page needs to be functional so the user can follow along and actually try things out. Can you show more requests demonstrating the caching? Inspecting status codes and things like that? This content mentions some cool topics too like Etags; but fails to mention that the requests have to send those same tags back to the client for subsequent requests to actually be cache hits. https://github.com/HarperDB/react-ssr-example/blob/151685affcf84e50263720ad8f1e0321d2c87fed/caching-test.js#L13-L21 without them you'll get a 200 every time. 

This page barely scratches the surface for caching without ever showing off a real example. This doesn't have to go over _everything_ or go deep into everything. I'd expect a really solid example or two and then some more high level information on the other capabilities with links out to either more examples or reference docs. If those examples don't exist yet, then put a note like "example coming soon!" so we can come back around to it. If reference docs don't exist definitely let us know so we can write them!

versioned_docs/version-4.6/reference/caching.md
```diff
@@ -0,0 +1,272 @@
+---
+title: Caching
+---
+
+# Caching
+Harper provides integrated caching features for external data sources. This page documents all schema directives, configuration options, and resource APIs relevant to caching.
```
Ethan-Arrowood, line 6: Our caching system is not just for external data sources. In fact we'd rather our users use Harper as the data source. But we support integration with 3rd party data sources because that is an important pattern for customers. Also why is this page documenting Resource api methods? Those have their own reference page. If you're trying to demonstrate how they integrate with a caching example I think thats better set for the guide page or its own example.

Ethan-Arrowood (changes requested):

versioned_docs/version-4.6/developers/applications/caching.md
nenharper: We have been using the dog example. Not necessarily dog-api.com but it's just a placeholder

Ethan-Arrowood (commented):
okay this is really strong; I'd like to get another engineer to review the reference doc for completeness and correctness but it looks good from my review. Otherwise, I really like the new example too. Nice work with the ETag headers too.

docs/reference/index.md
Ethan-Arrowood, line 1: Why are we adding back the TOC? This was intentionally removed in #313

Ethan-Arrowood (commented):
Other than randomly adding back in the table of contents reference index pages, I think this PR is good to go. Can you revert the changes to those files and just retain the `caching.md` changes? And then can you replicate to the relevant versions too?

PR #322, opened 2025-09-17
New fastify page

nenharper:
This PR makes the fastify page more narrative and adds a reference for it


Ethan-Arrowood (commented):
I wish we could've consulted the outline before you spent time rewriting this section. We are moving away from recommending fastfy routes and rather want users to be implementing custom resources instead (which has only been briefly mentioned previously). I think we need to discuss as a team if it makes sense to even have this part of the guide still in play. I agree some reference for it will always be necessary, but it could be potentially omitted from this guide.

versioned_docs/version-4.5/developers/components/built-in.md
```diff
@@ -20,19 +19,6 @@ Harper provides extended features using built-in components. They do **not** nee
 
 <!-- ## clustering -->
 
-## fastifyRoutes
```
Ethan-Arrowood, line 23: We likely want to keep this part as we'll still support fastify through this built in and there are customers actively using it on this version too. We aren't removing fastify support yet; just don't want to promote its use anymore

kriszyp (commented):
This PR is _only_ for changing 4.5 and 4.6 docs, but not touching current/4.7?

PR #326, opened 2025-09-24
Purge SQL pages

nenharper:
Purge SQL section (based on Kyle & Drew's suggestion)

### DIFF
--- a/redirects.ts
+++ b/redirects.ts
@@ -150,6 +150,13 @@ function generateDocsRedirects(basePath: string): RedirectRule[] {
 
 		// SQL Guide
 		{ from: withBase('/sql-guide'), to: withBase('/') },
+		{ from: withBase('/developers/sql-guide'), to: withBase('/') },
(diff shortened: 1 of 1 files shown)


Ethan-Arrowood (commented):
Needs to be removed from `docs/` too.

Also need to check for dead links throughout rest of docs referencing these pages. 

And need to consider appropriate redirects. These pages are likely heavily indexed by search engines, and may even exist throughout other posts and things. I think redirecting to like another querying page or even just the home page of the docs is fine.

redirects.ts
```diff
@@ -149,7 +149,7 @@ function generateDocsRedirects(basePath: string): RedirectRule[] {
 		},
 
 		// SQL Guide
-		{ from: withBase('/sql-guide'), to: withBase('/developers/sql-guide/') },
+		{ from: withBase('/sql-guide'), to: withBase('/') },
```
Ethan-Arrowood, line 153: We also need to create redirects for all these pages being deleted (to redirect to `/` or else where) so that these deleted pages do not become dead links.

PR #327, opened 2025-09-29
Update defining schemas page

nenharper:
This PR updated the defining schemas page


Ethan-Arrowood (approved):
LGTM please copy to other relevant versions and lets get this merged 🚀

kriszyp (approved):
I think generally looks good. I assume we may need to copy to the correct versions.

PR #352, opened 2025-10-14
add ga4 gtag

Ethan-Arrowood:


Ethan-Arrowood:
I added the secret; not sure why CI failed first run. Checking again before merging to be certain.

Ethan-Arrowood:
Ah that is definitely what is missing; thought I pushed that commit. Whoops

PR #353, opened 2025-10-15
Fix: Add read_log log_name attr

cap10morgan:
This has been an optional attribute [since forever](https://github.com/HarperFast/harperdb/blob/ded3df3cce4d7b297b03ec7f9c24ad3e22b29db8/validation/readLogValidator.js#L35), but was missing from the docs.


cap10morgan:
> But we recently added the ability to have distinct log files for different components. So it actually is kinda useful again.

So maybe change it to something like `component` and have it default to `harper` (or similar) when omitted?

cb1kenobi (approved):

PR #384, opened 2025-11-05
Removed hrefs for redirect in Docs

T-Brougham:


Ethan-Arrowood (approved):
I don't see any links using the `#install` anchor so I think this is good to go. 

Thanks for fixing!

PR #396, opened 2025-11-25
Hide version dropdown on non-versioned paths

Ethan-Arrowood:
Hides the version dropdown in the navbar when on a non-versioned path such as `/release-notes` or `/fabric`. Works on all subpaths too. 

Before:
Screenshot 2025-11-25 at 9 26 21 AM

After: 
Screenshot 2025-11-25 at 9 25 20 AM

> Search bars are different due to different search components on the live site and local development. 

Found solution for this here: https://github.com/facebook/docusaurus/issues/4389#issuecomment-1962954701

### DIFF
--- a/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
+++ b/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
@@ -3,12 +3,12 @@ import DocsVersionDropdownNavbarItem from '@theme-original/NavbarItem/DocsVersio
 import { useLocation } from '@docusaurus/router';
 import type { Props } from '@theme/NavbarItem/DocsVersionDropdownNavbarItem';
 
-function checkPathname(pathname: string) {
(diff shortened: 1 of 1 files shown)


src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import DocsVersionDropdownNavbarItem from '@theme-original/NavbarItem/DocsVersionDropdownNavbarItem';
+import { useLocation } from '@docusaurus/router';
+import type { Props } from '@theme/NavbarItem/DocsVersionDropdownNavbarItem';
+
+function checkPathname(pathname: string) {
```
dawsontoth: 💡 What if you named this `isNonVersionedDocs` or something? Just an idea. The logic of it all looks good to me, though.

PR #397, opened 2025-11-26
add detailed scanInterval description

hdbjeff:
Provided a detailed description of how to determine when the eviction cycle will run based on the scanInterval setting.

### DIFF
--- a/docs/developers/applications/caching.md
+++ b/docs/developers/applications/caching.md
@@ -38,14 +38,12 @@ You can provide a single expiration and it defines the behavior for all three. Y
 
 `expiration` = 1 hour with default `scanInterval` (15 minutes, one quarter of `expiration`). This creates the following fixed eviction schedule:
 
-```
-00:00  
-00:15  
-00:30  
-00:45  
-01:00  
-... continuing every 15 minutes ...
-```
+> 00:00  
+> 00:15  
+> 00:30  
+> 00:45  
+> 01:00  
+> ... continuing every 15 minutes ...
 
 If the server starts at 12:05 it does not run eviction at 12:20 or “15 minutes after startup.” Instead, the next scheduled anchor is 12:15, then 12:30, 12:45, 13:00, etc. The schedule is clock-aligned, not startup-aligned.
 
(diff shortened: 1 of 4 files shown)


docs/developers/applications/caching.md
```diff
@@ -26,6 +26,8 @@ You can provide a single expiration and it defines the behavior for all three. Y
 - `eviction` - The amount of time after expiration before a record can be evicted (defaults to zero).
 - `scanInterval` - The interval for scanning for expired records (defaults to one quarter of the total of expiration and eviction).
 
+How often records are evicted is based on the `scanInterval` setting. Using Javascript's `setHours` method, we divide days and years by the interval in the local timezone of the Harper server to find the specific times to run at, regardless of the start time of the Harper server. For example, if the Harper server started at 1205 and the table `expiration` was set to 1 hour, with the default `scanInterval` (here being 15 minutes, one quarter of the TTL) eviction would run at 1215, 1230, 1245, 1300 etc. If the Harper server started at 1205pm and the `expiration` was set to 1 day, with the default `scanInterval` (now of 6 hours) the eviction would run at 1800 the same day, 0000 and 0600 the next day, etc.
```
Ethan-Arrowood: What is "JavaScript's `setHours` method"? This: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setHours ?

Furthermore, I feel like this description is missing something. I understand the start time doesn't have any impact, so then what does? Would a server that started at _any_ time and an `expiration` of 1 day also have an eviction at 1800 the same day? What if my server started after 1800?

heskew: Something more like this (easier to scan and remove impl detail)?

```
- `scanInterval` - The interval for scanning for expired records, in seconds (defaults to 1/4 of the total of expiration and eviction).

**Scan Alignment Examples**:
- 1-hour expiration, default interval (15 min): :00, :15, :30, :45 each hour
- 6-hour expiration, default interval (90 min): 00:00, 01:30, 03:00, 04:30, 06:00, 07:30... 
- 1-day expiration, default interval (6 hours): 00:00, 06:00, 12:00, 18:00 daily
- 1-day expiration, custom 8-hour interval: 00:00, 08:00, 16:00 daily

Scans occur at these clock times in the server's local timezone, regardless of when the server started.
```

Ethan-Arrowood (approved):
amazing. love it!

PR #408, opened 2025-12-17
Preview PRs using Harper Fabric ✨

Ethan-Arrowood:
This PR adds a new workflow for deploying PRs to a Harper Fabric cluster for previewing. You can see how it works in this PR itself.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -18,8 +18,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
-        with:
-          fetch-depth: 0
 
       - uses: actions/setup-node@v6
         with:
--- a/.github/workflows/pr-preview.yaml
+++ b/.github/workflows/pr-preview.yaml
@@ -74,7 +74,7 @@ jobs:
 
       - name: Install HarperDB CLI
         run: |
-          npm install harperdb
+          npm install -g harperdb
 
       - name: Deploy to HarperDB
         env:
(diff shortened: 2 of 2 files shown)


heskew:
👀

Ethan-Arrowood:
I'm going to attempt the close and reopen flows now.

heskew (approved):
sweet. love the prod workflow cleanup too. 

is there a risk of someone cutting a pr we don't notice and putting something malicious up on an official lookin (though clearly not 'production') harper site? can we put it on a manual gate to start, eventually auto deploy trusted groups/individuals?

.github/workflows/deploy.yaml
```diff
@@ -1,43 +1,27 @@
...
 
 jobs:
   build:
     name: Build Docusaurus
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0
```
cb1kenobi: Generally, I don't think you want `fetch-depth: 0` as it will download the entire git history when you probably only need the most recent commit which is faster.

kriszyp (approved):
So cool!

.github/workflows/pr-preview.yaml
```diff
@@ -0,0 +1,203 @@
...
+
+          # Copy build directory
+          cp -r build "$PR_DIR/"
+
+          echo "Contents of $PR_DIR:"
+          ls -la "$PR_DIR"
+
+      - name: Install HarperDB CLI
+        run: |
+          npm install harperdb
```
cb1kenobi: Why not install harper globally? Then I would think you wouldn't need `npx` on the next step, right?

PR #475, opened 2026-04-16
Document replication settings, working directory, and latency metric

kriszyp:


reference/components/javascript-environment.md
```diff
@@ -133,3 +133,7 @@ export class MyResource extends Resource {
 ### `getResponse()`
 
 Returns the outgoing `Response` object for the current request, or `undefined` if called outside a request context. Use this to set response headers or inspect the response mid-handler. Equivalent to `getContext().response`.
+
+### Current Working Directory
+
+Harper using a multi-threaded server architecture and uses the harper data root path as the current working directory. Components should not and cannot change the current working directory.
```
cb1kenobi:
```suggestion
Harper has a multi-threaded server architecture and uses the harper data root path as the current working directory. Components should not and cannot change the current working directory.
```

Or better yet?
```suggestion
Harper has a multi-threaded server architecture and uses the harper data root path as the current working directory. Components should not and cannot change the current working directory, so don't even think about using `process.chdir()` or any package that uses `process.chdir()`.
```

kriszyp: Love the second, but kinda went with hybrid, polite version :).

cb1kenobi (approved):
Looks good!

PR #495, opened 2026-05-25
Fix Resource post/put/patch data type + document search sort + content-type table

kriszyp:
## Summary

Three corrections / clarifications to `reference/resources/resource-api.md` based on real friction I hit while building harper-celebrity-match and harper-roadmap demos.

### 1. `post` / `put` / `patch` `data` type is wrong

The page documents the signature as `data: Promise<object>` and the example shows `let data = await promisedData;`. That's not what the runtime actually passes. In [server/REST.ts](https://github.com/HarperFast/harper/blob/main/server/REST.ts#L96-L121) `request.data` is assigned the synchronous return value of the deserializer — for `application/json` requests that's the parsed value, not a Promise.

Result: every user initially writes `const data = await promisedData`, sees it work because `await` of a non-thenable just resolves to itself, then copy-pastes that into other handlers. Eventually they find `Search.js` / `Import.js` in our example components doing `const { foo } = data` and rewrite — the docs should match the example.

Fixed signature: `data: object`. Example updated to drop the `await`.

### 2. Content-type → `data`-shape table

Added a small table explaining what `data` is for the built-in deserializers (JSON, CBOR, MessagePack, NDJSON, plain text).

### 3. `sort` requires a matching indexed condition

The current sort docs imply `sort: { attribute: 'id' }` should just work because `id` is the primary key. It doesn't — Harper's query planner requires the sort attribute to be both `@indexed` AND narrowed by a `conditions` entry. The error message you get is "id is not indexed and not combined with any other conditions" which is misleading (id IS indexed). Documented the actual requirement + the workaround.

Tracks:
* HarperFast/harper#773 — improve the error message itself
* HarperFast/harper#774 — separate long-running-fetch abort docs

### DIFF
--- a/reference/resources/resource-api.md
+++ b/reference/resources/resource-api.md
@@ -152,12 +152,12 @@ class MyResource extends Resource {
 
 The shape of the resolved value depends on the request's `Content-Type`. The built-in deserializers produce:
 
-| `Content-Type` | resolved `data` |
-|---|---|
-| `application/json` | parsed JSON value |
-| `application/cbor`, `application/x-msgpack` | decoded value |
+| `Content-Type`                               | resolved `data`       |
+| -------------------------------------------- | --------------------- |
+| `application/json`                           | parsed JSON value     |
+| `application/cbor`, `application/x-msgpack`  | decoded value         |
 | `application/x-ndjson`, `application/ndjson` | array of parsed lines |
-| `text/plain` | string |
+| `text/plain`                                 | string                |
 
 Custom content types registered through `contentTypes.set(...)` receive whatever value the deserializer returns.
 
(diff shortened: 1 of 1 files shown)


dawsontoth (approved):

Ethan-Arrowood (approved):
LGTM — verified all three sections against harper main (the promise-of-data behavior in `REST.ts`, every row of the content-type table, and the sort/index logic in `Table.ts`) and they all check out.

Two non-blocking notes: the PR description still argues the pre-revision thesis (drop the `await`) — worth refreshing so the merge commit doesn't memorialize it; and the `greater_than`/`''` workaround example only works for string primary keys (numeric ids won't compare above `''`). Neither blocks — merging.

PR #496, opened 2026-05-26
docs(storage): document storage.rocks.* memory config options

kriszyp:
Documents four new storage configuration parameters being added in v5.1.0 (source: [HarperFast/harper#780](https://github.com/HarperFast/harper/pull/780)).

## What's added

**`reference/configuration/options.md`** — four new bullets under the existing `storage` section:
- `rocks.blockCacheSize`
- `rocks.writeBufferManagerSize`
- `rocks.writeBufferManagerCostToCache`
- `rocks.writeBufferManagerAllowStall`

**`reference/database/storage-tuning.md`** — a new \"RocksDB Memory\" detail section between \"Read & Write Behavior\" and \"Storage Reclamation\":

- Intro framing the section as RocksDB-specific (no effect under \`lmdb\`)
- A subsection on how RocksDB reads are cached — laying out the three-tier hierarchy (block cache → OS page cache → disk) with concrete latency expectations, so operators understand why a smaller block cache isn't necessarily worse on memory-constrained hosts
- Per-param entries following the established \`Type:\` / \`Default:\` / prose / \`yaml\` block pattern
- Practical guidance: when to lower the block cache vs. when to enable the WriteBufferManager
- Explicit note that the block cache does not shrink on idle — important for operators reasoning about long-running deployments
- Soft cap vs. hard cap (\`allowStall: true\`) trade-off

## Scope decision

This PR is intentionally general guidance for any RocksDB deployment. The **Fabric-specific multi-tenant tuning rationale** (why we picked 15% block cache + 5% WBM in host-manager, the canopy memory data, the dependency chain across the three implementation PRs) is captured in [host-manager's DESIGN.md](https://github.com/HarperFast/host-manager/pull/100) rather than the public docs, since it's our infrastructure concern rather than something operators of self-hosted Harper need to reason about.

## Where to focus review

- **Accuracy** of the cache-hierarchy description and latency claims (1 µs / 5-20 µs / 100 µs-10 ms tiers) — these align with empirical testing on the rocksdb-js binding side ([HarperFast/rocksdb-js#584](https://github.com/HarperFast/rocksdb-js/pull/584)).
- **Accuracy** of the cost-to-cache description: pinned WBM charges show up in \`block-cache-usage\` but LRU cannot evict them, so the cache cannot \"shrink\" to make room for writes — verified in the rocksdb-js tests in #584.
- **Tone**: matches the existing storage-tuning.md voice (problem-first, parameter description folded into guidance). Happy to restructure if a different shape is preferred.
- **Version annotation** uses \`<VersionBadge version=\"v5.1.0\" />\` per the existing convention in \`replication/\`.

## What's NOT changed

- No release notes entry — that would live in \`release-notes/v5.1.0.md\` once that file exists.
- No changes to v4 docs in \`reference_versioned_docs/version-v4/\` (these options are v5+ only).

## Build / format

- \`npm run format:write\` — clean
- \`npm run build\` — clean (the one broken-anchor warning is pre-existing in \`resources/resource-api.md\` referencing \`javascript-environment#transaction\`, unrelated)

Generated by Claude. — *Claude Sonnet 4.7*

### DIFF
--- a/reference/configuration/options.md
+++ b/reference/configuration/options.md
@@ -241,7 +241,7 @@ storage:
 - `reclamation.evictionFactor` — Heuristic factor for early eviction under disk pressure; _Default_: `100000`. See [Storage Tuning — Reclamation](../database/storage-tuning.md#storage-reclamation)
 - `rocks.blockCacheSize` — RocksDB shared block cache size in bytes; _Default_: 25% of constrained memory. See [Storage Tuning — RocksDB Memory](../database/storage-tuning.md#rocksdb-memory) (Added in: v5.1.0)
 - `rocks.writeBufferManagerSize` — Process-wide cap (bytes) on RocksDB memtable memory across all databases. `0` disables; _Default_: one third of `blockCacheSize` (enabled). See [Storage Tuning — RocksDB Memory](../database/storage-tuning.md#rocksdb-memory) (Added in: v5.1.0)
-- `rocks.writeBufferManagerCostToCache` — Charge memtable memory against the block cache so both share a single accounting pool; _Default_: `true`. Has no effect when `writeBufferManagerSize` is `0`. (Added in: v5.1.0)
+- `rocks.writeBufferManagerCostToCache` — When enabled, memtable memory and the block cache share a single, unified memory pool. During heavy write bursts, the block cache dynamically shrinks to give memtables more room. Once those memtables flush to disk, the block cache automatically reclaims that memory space. _Default_: `true`. Has no effect when `writeBufferManagerSize` is `0`. (Added in: v5.1.0)
 - `rocks.writeBufferManagerAllowStall` — Stall writes when memtable memory exceeds `writeBufferManagerSize` (hard cap) instead of allowing brief overshoot with more aggressive flushing (soft cap); _Default_: `true`. (Added in: v5.1.0)
 
 ---
--- a/reference/configuration/options.md
+++ b/reference/configuration/options.md
@@ -240,9 +240,9 @@ storage:
 - `reclamation.interval` — Free-space check interval; _Default_: `1h`
 - `reclamation.evictionFactor` — Heuristic factor for early eviction under disk pressure; _Default_: `100000`. See [Storage Tuning — Reclamation](../database/storage-tuning.md#storage-reclamation)
 - `rocks.blockCacheSize` — RocksDB shared block cache size in bytes; _Default_: 25% of constrained memory. See [Storage Tuning — RocksDB Memory](../database/storage-tuning.md#rocksdb-memory) (Added in: v5.1.0)
-- `rocks.writeBufferManagerSize` — Process-wide cap (bytes) on RocksDB memtable memory across all databases. `0` disables; _Default_: `0`. See [Storage Tuning — RocksDB Memory](../database/storage-tuning.md#rocksdb-memory) (Added in: v5.1.0)
-- `rocks.writeBufferManagerCostToCache` — Charge memtable memory against the block cache so both share a single accounting pool; _Default_: `false`. Has no effect when `writeBufferManagerSize` is `0`. (Added in: v5.1.0)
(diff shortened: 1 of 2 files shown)


reference/configuration/options.md
```diff
@@ -239,6 +239,10 @@ storage:
 - `reclamation.threshold` — Free-space ratio below which reclamation begins evicting from caching tables; _Default_: `0.4` (Added in: v4.5.0)
 - `reclamation.interval` — Free-space check interval; _Default_: `1h`
 - `reclamation.evictionFactor` — Heuristic factor for early eviction under disk pressure; _Default_: `100000`. See [Storage Tuning — Reclamation](../database/storage-tuning.md#storage-reclamation)
+- `rocks.blockCacheSize` — RocksDB shared block cache size in bytes; _Default_: 25% of constrained memory. See [Storage Tuning — RocksDB Memory](../database/storage-tuning.md#rocksdb-memory) (Added in: v5.1.0)
+- `rocks.writeBufferManagerSize` — Process-wide cap (bytes) on RocksDB memtable memory across all databases. `0` disables; _Default_: one third of `blockCacheSize` (enabled). See [Storage Tuning — RocksDB Memory](../database/storage-tuning.md#rocksdb-memory) (Added in: v5.1.0)
+- `rocks.writeBufferManagerCostToCache` — Charge memtable memory against the block cache so both share a single accounting pool; _Default_: `true`. Has no effect when `writeBufferManagerSize` is `0`. (Added in: v5.1.0)
```
cb1kenobi: The word "charge" is kinda confusing. It steps from the description in rocksdb-js which also is confusing.


```suggestion
- `rocks.writeBufferManagerCostToCache` — When enabled, memtable memory and the block cache share a single, unified memory pool. During heavy write bursts, the block cache dynamically shrinks to give memtables more room. Once those memtables flush to disk, the block cache automatically reclaims that memory space. _Default_: `true`. Has no effect when `writeBufferManagerSize` is `0`. (Added in: v5.1.0)
```

reference/database/storage-tuning.md
```diff
@@ -138,6 +138,109 @@ Default: `true`
 
 In-memory record caching of decoded records. Disable to reduce heap usage when records are large and unlikely to be re-read in the same process.
 
+## RocksDB Memory
+
+RocksDB exposes two large native memory pools that Harper makes tunable: a shared **block cache** for hot SST blocks, and a **WriteBufferManager** (enabled by default) that caps total memtable memory across every database in the process. These options apply only when `storage.engine` is `rocksdb`.
```
cb1kenobi: I believe the comma is unnecessary.
```suggestion
RocksDB exposes two large native memory pools that Harper makes tunable: a shared **block cache** for hot SST blocks and a **WriteBufferManager** (enabled by default) that caps total memtable memory across every database in the process. These options apply only when `storage.engine` is `rocksdb`.
```

reference/database/storage-tuning.md
```diff
@@ -138,6 +138,109 @@ Default: `true`
...
+
+### `storage.rocks.writeBufferManagerSize`
+
+<VersionBadge version="v5.1.0" />
+
+Type: `number` (bytes)
+
+Default: one third of `blockCacheSize` (enabled). Set to `0` to disable.
+
+Harper attaches a single RocksDB `WriteBufferManager` to every opened database in the process. Total memtable memory — including active memtables, immutable memtables awaiting flush, and the maintain-window history that RocksDB's OptimisticTransactionDB retains for conflict checking — is capped at this size across the entire process.
```
cb1kenobi: Consider using monospace font for `OptimisticTransactionDB`.

```suggestion
Harper attaches a single RocksDB `WriteBufferManager` to every opened database in the process. Total memtable memory — including active memtables, immutable memtables awaiting flush, and the maintain-window history that RocksDB's `OptimisticTransactionDB` retains for conflict checking — is capped at this size across the entire process.
```

cb1kenobi (approved):
This is awesome, thank you for documenting this!

cb1kenobi (approved):

PR #508, opened 2026-06-01
docs(schema): document HNSW int8 quantization option

kriszyp:
## Summary
Documents the new `quantization: "int8"` option for HNSW vector indexes — adds a row to the HNSW parameter table and a short example.

## Context
Pairs with HarperFast/harper#894 (optional int8 vector quantization for the HNSW index): ~5× smaller index, substantially faster search, ~1% recall cost; opt-in, the record's full-precision vector is unchanged. Making it the default is tracked separately (HarperFast/harper#932).

_Generated with the assistance of an LLM (Claude Opus 4.8)._

### DIFF
--- a/reference/database/schema.md
+++ b/reference/database/schema.md
@@ -227,6 +227,34 @@ If the field value is an array, each element in the array is individually indexe
 
 Null values are indexed by default (added in v4.3.0), enabling queries like `GET /Product/?category=null`.
 
+### `@embed`
+
+<VersionBadge version="v5.1.0" />
+
+Automatically computes an embedding vector for the attribute whenever the source field is written, using a configured [embedding model](../models/overview):
+
+```graphql
+type Document @table {
+	id: Long @primaryKey
+	text: String
+	embedding: [Float] @embed(source: "text", model: "default")
+}
+```
+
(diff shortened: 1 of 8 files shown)


reference/database/schema.md
```diff
@@ -453,6 +422,16 @@ let results = Document.search({
 | `optimizeRouting`      | `0.5`             | Heuristic aggressiveness for omitting redundant connections (0 = off, 1 = most aggressive)          |
 | `mL`                   | computed from `M` | Normalization factor for level generation                                                           |
 | `efSearchConstruction` | `50`              | Max nodes explored during search                                                                    |
+| `quantization`         | _(full precision)_ | `"int8"` stores each indexed vector as 8-bit scalar-quantized values plus a per-vector scale instead of float32 — roughly a 5× smaller index and substantially faster search, at a small recall cost (~1%). Omit for full-precision float32. Only the index is quantized; the full-precision vector on the record is unchanged. |
```
Ethan-Arrowood: I think default should be `"float32"` for this option to make more sense. `_(full precision)_` is not a value. I believe right now this option only supports two values, but will there be more in the future? If not, and this is truly a binary configuration, should it not reflect that better? Like maybe making it a toggleable `int8Quantization: boolean` ?

PR #526, opened 2026-06-11
Document vm-current-context as the default module loader

kriszyp:
## Summary

Update the v5 migration guide for the new default `applications.moduleLoader: vm-current-context` (HarperFast/harper#1248).

- Rewrite the **VM Module Loader** intro to describe shared intrinsics as the default and how scoped values are obtained (`import ... from 'harper'`).
- Add a **Module Loader Modes** reference table covering `vm-current-context`, `vm`, `native`, and `compartment`.
- Note that the SES-constrained (https-only) `fetch` applies only in `vm` mode.

## Purpose

Keep the migration docs aligned with the changed default so app authors understand the intrinsics-sharing behavior and when to opt back into `vm`.

_Generated with assistance from Claude (Opus 4.7); reviewed by Kris._

### DIFF
--- a/release-notes/v5-lincoln/v5-migration.md
+++ b/release-notes/v5-lincoln/v5-migration.md
@@ -118,14 +118,14 @@ The `blob.save()` method has been removed. Please use the `saveBeforeCommit` fla
 
 ## VM Module Loader
 
-Harper v5 loads application modules through Node.js's VM module API, giving each application its own module cache and execution context. This provides per-application context: the `logger` global/export is automatically tagged with the application name, and `config` reflects that application's own configuration. Each application's module graph is isolated from other applications and from Harper internals.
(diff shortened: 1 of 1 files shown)


Ethan-Arrowood (approved):
Do we have something updating this information wherever its detailed in the /reference/ section? Not just the release notes?

kriszyp (commented):
Good catch on the migration guide. One follow-up: `reference/configuration/options.md` line 330 still describes `moduleLoader` with `vm` as the default and doesn't mention `vm-current-context` as an option. Should that be updated in this PR or a follow-on?

Also worth noting: the table in this PR could link to or cross-reference the options page so readers can find the full config reference. Otherwise LGTM — the migration guide update is the right place to explain the behavioral change.

PR #529, opened 2026-06-11
docs(configuration): document HARPER_CONFIG and the `$union` array directive

heskew:
## Summary

Documents the two config env-var features shipped in HarperFast/harper#1217 (`$union`) and #1238 (`HARPER_CONFIG`), both sub-issues of HarperFast/harper#1097. All changes are in `reference/configuration/overview.md`.

- **`HARPER_CONFIG`** — new, documented as the *recommended* way to set configuration from the environment: merge semantics (sets only the keys it names), reasserts its keys on every restart, yields only to `HARPER_SET_CONFIG`, with individual env vars/CLI args still winning for the keys they set, and original-value restoration on key removal.
- **`$union` array directive** — how to compose an array (idempotent, order-preserving, non-destructive) instead of the default wholesale replace, across all three config env vars; plus the note that only recognized `$`-directives (`$union`) are special so JSON-Schema `$schema`/`$ref` in component config pass through untouched.
- **Updated** the intro (two → three variables), the precedence ladder, the "combining" example (platform-pins-via-SET + app-sets-via-CONFIG), and the state-tracking and backup notes to cover all three variables.

Targets `v5.1.0` version badges, matching the existing `v5.0.0` convention in these docs.

## Notes

- Single file touched; `prettier --check` clean. Built on a worktree off `main` (not the in-progress `docs/models-5.1` branch).
- Closes the documentation deliverable in HarperFast/harper#1097.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) (agent: Claude Opus 4.8)

### DIFF
--- a/reference/configuration/overview.md
+++ b/reference/configuration/overview.md
@@ -221,7 +221,7 @@ export HARPER_SET_CONFIG='{"tls":{"uses":{"$union":["server","operations-api"]}}
 - **Idempotent** — reapplying the same `$union` on every restart is a no-op; it never grows duplicates
 - **Non-destructive** — it only adds the items it names and never removes other entries, even under `HARPER_SET_CONFIG`'s drift handling
 
-This lets a platform layer guarantee required array entries (for example `tls.uses`) on every restart without clobbering entries an application adds. `$union` works in all three variables. A bare array (no `$union`) still replaces, which remains the default.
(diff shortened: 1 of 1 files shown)


kriszyp (approved):

kriszyp (approved):
Still approve

PR #536, opened 2026-06-24
docs(sse): document reading the request in static connect()

maurice-harper:
Adds a "Reading the request in connect()" section to the Server-Sent Events reference, explaining how to access query parameters, headers, and the path by defining connect() as a static method that receives the RequestTarget. The section includes a worked example using target.get() and clarifies that connect() reads from the target synchronously before returning the async generator that streams events.

### DIFF
--- a/reference/rest/server-sent-events.md
+++ b/reference/rest/server-sent-events.md
@@ -47,24 +47,38 @@ The default `connect()` behavior subscribes to the resource and streams changes 
 
 ## Reading the request in `connect()`
 
-To read the incoming request (query parameters, headers, the path), define `connect()` as a **`static`** method. Its first argument is the `RequestTarget`, which exposes the query string via `target.get()`:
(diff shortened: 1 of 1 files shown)


kriszyp (commented):
This is a useful addition. One simplification question on the example: why not just make `connect` itself the generator?

```js
export class Search extends Resource {
	static async *connect(target) {
		const query = target.get('q'); // GET /Search?q=harper
		const apiKey = target.get('api_key');
		for (const hit of await runSearch(query)) yield { data: hit };
	}
}
```

The two-method split (`connect` reads the request, then returns a separate `stream` generator) is the pattern you needed back when you had to capture `context`/`request` *synchronously* at the top before the generator body started executing. But this example no longer uses `context` — it uses `target`, which is just a function argument and is therefore naturally captured inside the generator already. So the wrapper doesn't buy anything here, and a single `static async *connect(target)` is the simpler thing to teach. Worth showing the one-method form as the primary example (and maybe keeping the split only if there's a real case where you must do synchronous work before the first yield).

— Claude (Opus 4.8)

kriszyp (approved):
The single-method `static async *connect` form reads well, and keeping the split form for the sync-work/getContext case is the right call. Approving. — Claude (Opus 4.8)

PR #537, opened 2026-06-24
docs(resources): document relationship select and join behavior

Ethan-Arrowood:
## Summary

The Query Object [`select`](reference/resources/resource-api.md) reference documented the nested-select *shape* (`[{ name: 'brand', select: ['id', 'name'] }]`) but not the actual relationship-select behavior — whole vs partial related records, multi-level nesting, to-many array resolution, and join semantics.

This adds a **"Selecting related records"** subsection under `### select` covering:
- Whole related record (list the relationship field by name)
- Partial related record (`{ name, select }`)
- Multi-level nesting
- To-many relationships resolving to arrays (and the `await`-before-iterate nuance)
- **Join behavior:** LEFT JOIN (select without a filter on the relation) vs INNER JOIN (filter on a related attribute)

## Why

This behavior was previously only captured in downstream agent rules (`@harperfast/skills` `programmatic-table-requests`), not the canonical reference. Promoting it to the docs makes the skills rule sourceable from canonical docs (it's being migrated from hand-authored to docs-generated) and gives all readers the same join semantics.

### DIFF
--- a/reference/resources/resource-api.md
+++ b/reference/resources/resource-api.md
@@ -866,6 +866,34 @@ Special properties:
 - `$updatedtime` — Returns the last-updated timestamp
 - `$distance` — When the query ranks or filters by a vector index, returns the computed distance from the target vector. See [Vector Indexing](../database/schema.md#vector-indexing).
 
+#### Selecting related records
+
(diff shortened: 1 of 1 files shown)


kriszyp (approved):
Relationship select/join behavior is documented accurately and the JOIN terminology matches the existing canonical `querying.md`. Approving. — Claude (Opus 4.8)

PR #542, opened 2026-06-24
docs(database): add data-safety note for programmatic writes/deletes

Ethan-Arrowood:
## Summary

Adds a **"Data safety"** warning to the `tables` section of [`reference/database/api.md`](reference/database/api.md): programmatic `update`/`patch`/`delete` operate directly on stored (possibly production) data, an over-broad query can be destructive and isn't easily reversible, so scope with specific conditions, validate the affected set, and gate behind approval/authorization controls.

## Why

This guidance previously lived only in the `@harperfast/skills` `programmatic-table-requests` rule. As that rule migrates from hand-authored to docs-generated ([skills#63](https://github.com/HarperFast/skills/pull/63)), a pure regeneration drops the caution because it had no canonical source. Promoting it here keeps the safety guidance in the canonical reference (where all readers see it) and lets it round-trip through generation rather than being hand-grafted back into the rule.

It lands inside the `tables` section that skills#63 sources, so once this merges the rule regenerates with the caution intact.

### DIFF
--- a/reference/database/api.md
+++ b/reference/database/api.md
@@ -51,6 +51,10 @@ for await (const record of Product.search(query)) {
 
 For the full set of methods available on table classes, see the [Resource API](../resources/resource-api.md).
 
+:::warning Data safety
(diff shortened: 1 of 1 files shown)


kriszyp (approved):
Data-safety note for programmatic writes/deletes is accurate per the resource API. Approving. — Claude (Opus 4.8)

PR #543, opened 2026-06-24
docs: clarify Harper APIs are one live process-wide instance (and SSR data access)

dawsontoth:
## Why

The **JavaScript Environment** page lists both the global form and `import { tables } from 'harper'`, but never says they're the **same instance** — and the page's opening framing ("contextualized module environments ... isolated from other applications") can read as if each app gets its own copy of the data-access objects. With no guidance on reading Harper data during SSR, that leads developers to believe a Vite-loaded SSR entry can't reach Harper data (and to reach for workarounds like fetching their own REST API over loopback).

## What

Added to the **Module Loading** section of `reference/components/javascript-environment.md`:

- A clarification that globals and `harper` imports are the **same live, process-wide objects** (Harper is single-process, so writes are visible everywhere); the automatic link points `harper` at the **running** install, so `import { tables } from 'harper'` resolves to the live runtime from any module Harper loads; application module contexts are *seeded from* the process globals, not given an isolated set.
- A short **SSR** example — read a table in `render` — plus the `ssr.external: ['harper']` note for bundling (symlinked deps aren't reliably auto-externalized for SSR).

Passes `prettier --check`.

### DIFF
--- a/reference/components/javascript-environment.md
+++ b/reference/components/javascript-environment.md
@@ -30,6 +30,21 @@ npm link harper
 
 All installed components have `harper` automatically linked.
 
+Whether you reach them as globals or as `harper` imports, `tables`, `databases`, and the other APIs are the **same live, process-wide objects** — Harper runs as a single process, so a record written through one component is immediately visible to every other. The automatic link points `harper` at the **running** installation (not a separately-installed copy), so `import { tables } from 'harper'` resolves to that live runtime from any module Harper loads. Application module contexts are seeded from the same process globals, not given an isolated set of these objects.
+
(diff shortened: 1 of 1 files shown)


Ethan-Arrowood (approved):
LGTM — this is the canonical home for the process-wide/SSR clarification. I'm deferring my #544 to this. Heads up: I'll wire the `javascript-environment.md` Module Loading section into the skills `programmatic-table-requests` rule sources so it inherits this on regeneration.

kriszyp (approved):
Accurate and consistent with the process-wide-instance docs cluster. Approving. — Claude (Opus 4.8)

PR #547, opened 2026-06-25
docs(rest): clarify PATCH is a shallow top-level merge (nested objects replaced, not deep-merged)

kriszyp:
## What

Adds a warning to the REST **PATCH** reference clarifying that the merge is **shallow** (top-level only): a PATCH whose body contains a nested object **replaces** that whole sub-object rather than deep-merging it, so untouched nested properties are silently dropped.

## Why

The current text ("merging only the provided properties... Unspecified properties are preserved") is accurate for top-level attributes but reads as a deep merge. In practice `PATCH { "settings": { "theme": "dark" } }` drops `settings.notifications` — a silent data-loss footgun, since the request returns 204 with no indication anything was lost. This came up during exploratory QA (qa-explorer): a single nested-field PATCH silently loses sibling nested data on both engines, with no error and no public docs describing the behavior.

The behavior itself is consistent (RFC 7386-style top-level merge) — this PR just documents it and points to the safe pattern (read-modify-write the parent, or send the full nested object). Also notes that dot-path keys are stored literally.

— KrAIs (Claude) on Kris's behalf

### DIFF
--- a/reference/rest/overview.md
+++ b/reference/rest/overview.md
@@ -123,9 +123,11 @@ If the request body includes a nested object, that entire sub-object is **replac
 
 **Example:**
 
+**Example:**
+
 - **Existing record:** `{"settings": {"theme": "light", "notifications": {"email": true}}}`
 - **PATCH request body:** `{"settings": {"theme": "dark"}}`
-- **Resulting record:** `{"settings": {"theme": "dark"}}` (the "notifications" object is lost)
+- **Resulting record:** `{"settings": {"theme": "dark"}}` (the `notifications` object is lost)
 
 To update a single nested field, you must either:
 
--- a/reference/rest/overview.md
+++ b/reference/rest/overview.md
@@ -116,6 +116,10 @@ Content-Type: application/json
 { "status": "active" }
 ```
 
+:::warning
+The merge is **shallow** (top-level only). Preserving "unspecified properties" applies to top-level attributes — if the body includes a nested object, that whole sub-object is **replaced**, not deep-merged, so any nested properties you didn't include are dropped. For example, `PATCH { "settings": { "theme": "dark" } }` against a record whose `settings` is `{ "theme": "light", "notifications": { ... } }` results in `settings` being just `{ "theme": "dark" }` — `notifications` is gone. To update a single nested field, read-modify-write the parent object (or send the full nested object). Dot-path keys (`"settings.theme"`) are stored literally, not interpreted as paths.
+:::
+
 ### DELETE
 
 Delete a specific record or all records matching a query.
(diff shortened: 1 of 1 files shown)


dawsontoth (approved):
Content looks good! I like the example. Once the format is happy, I think this'll be good to go!

reference/rest/overview.md
```diff
@@ -116,6 +116,23 @@ Content-Type: application/json
...
+If the request body includes a nested object, that entire sub-object is **replaced** rather than deep-merged. Any omitted nested properties will be dropped.
+
+**Example:**
+* **Existing record:** {"settings": {"theme": "light", "notifications": {"email": true}}}
+* **PATCH request body:** {"settings": {"theme": "dark"}}
+* **Resulting record:** {"settings": {"theme": "dark"}} (the "notifications" object is lost)
+
+To update a single nested field, you must either:
+1. Read-modify-write the parent object.
+2. Send the full nested object with the updated values.
```
Ethan-Arrowood, line 133: This is what's breaking CI: the bare `{...}` JSON here gets parsed as JSX expressions by MDX (Docusaurus compiles `.md` through MDX), so the site build fails on this file — the Deploy PR Preview log shows `MDX compilation failed for reference/rest/overview.md`. Prettier also wants `-` bullets and blank lines before the lists.

This suggestion fixes both — I verified it passes `prettier --check` and compiles under MDX:

```suggestion
**Example:**

- **Existing record:** `{"settings": {"theme": "light", "notifications": {"email": true}}}`
- **PATCH request body:** `{"settings": {"theme": "dark"}}`
- **Resulting record:** `{"settings": {"theme": "dark"}}` (the `notifications` object is lost)

To update a single nested field, you must either:

1. Read-modify-write the parent object.
2. Send the full nested object with the updated values.
```

Ethan-Arrowood (approved):
Content looks great — this is a real footgun worth documenting, and the example makes it immediately clear. Echoing @dawsontoth: good to go once the format is happy.

The CI failure is caused by this diff (not a flake): bare JSON braces in MDX text. I've attached a one-click suggestion below that fixes the build and the prettier check (verified both locally). Approving on that basis — apply the suggestion and merge at your discretion once it's green.

reference/rest/overview.md
```diff
@@ -116,6 +116,27 @@ Content-Type: application/json
...
 ```
 
+:::warning
+The merge is **shallow** (top-level only). Preserving "unspecified properties" applies only to top-level attributes.
+
+If the request body includes a nested object, that entire sub-object is **replaced** rather than deep-merged. Any omitted nested properties will be dropped.
+
+**Example:**
+
+**Example:**
```
Ethan-Arrowood:
```suggestion
```

Ethan-Arrowood (approved):

PR #553, opened 2026-06-29
docs: clarify that @hidden, @export are not access controls + allowRead trusted-context / filter caveats

kriszyp:
These documentation additions address four behaviors that are easy to mistake for access controls but are not, helping developers avoid building insecure schemas.

- **`reference/database/schema.md`** — Added `:::warning` callout under `@export` clarifying it is a REST routing directive only (omitting it returns 404 but the data remains ops-API/SQL accessible to admins). Added `:::warning` callout under `@hidden` (type directive) that the value is returned on all read surfaces for any role with table `read` permission. Strengthened the `@hidden` field directive description with the same boundary.
- **`reference/users-and-roles/overview.md`** — Added `:::note` under attribute-level rules clarifying that `read: false` blocks the value from response bodies but the attribute can still be used as a filter predicate, enabling boolean inference of exact values via row-count observation.
- **`reference/components/javascript-environment.md`** — Added `:::note` under the `tables` global clarifying that server-side `tables.X.get()`/`.search()` calls execute in a trusted system context and do not re-apply the target table's `allowRead` or role-level `attribute_permissions`; resource code must apply its own authorization when cross-reading protected tables.

### DIFF
--- a/reference/components/javascript-environment.md
+++ b/reference/components/javascript-environment.md
@@ -85,6 +85,10 @@ An object whose properties are the tables in the default database (`data`). Each
 
 See [Database API](../database/api.md) for full reference.
 
+:::note Server-side `tables.*` reads run in a trusted context
+Calls to `tables.X.get()`, `tables.X.search()`, and similar methods from within resource code (or Operations API SQL) execute in a **trusted system context** and do **not** re-apply the target table's `allowRead` guard or role-level `attribute_permissions`. This is by design — server-side code is responsible for its own authorization checks. A computed attribute or custom resource that cross-reads a protected table will pull the raw data into the response regardless of the caller's role. If you expose data from a protected table through a custom resource, apply authorization explicitly (e.g. check `getCurrentUser()` and enforce the relevant restrictions before returning the data).
+:::
+
 ### `databases`
 
 An object containing all databases defined in Harper. Each database is an object of its tables — `databases.data` is always equivalent to `tables`.
--- a/reference/database/schema.md
+++ b/reference/database/schema.md
@@ -174,6 +174,10 @@ type MyTable @table @export(name: "my-table") {
 
 The optional `name` parameter specifies the URL path segment (e.g., `/my-table/`). Without `name`, the type name is used.
 
+:::warning `@export` is a routing directive, not an access control
+Omitting `@export` removes the REST/MQTT route for a table (callers get 404), but it does **not** protect the data. The table still exists in the database and its records remain fully accessible to administrators via the Operations API and SQL. For data confidentiality, use role permissions (`attribute_permissions`, `read: false`) rather than relying on the absence of an export route.
+:::
+
 ### `@sealed`
 
 Prevents records from including any properties beyond those explicitly declared in the type. By default, Harper allows records to have additional properties.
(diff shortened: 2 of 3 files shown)


DavidCockerill (approved):
Approving. The most useful of this batch, because it documents two real disclosure paths rather than only correcting a misconception.

**I verified the `read: false` filter side-channel against the code, and it is real.** `Table.ts:1677-1708` applies `attribute_permissions` to `target.select` — it constrains what is *returned*. I could not find anything, anywhere, that rejects or strips a *condition* on a read-denied attribute. So the value never appears in a response body while the row count still answers questions about it, exactly as described, and binary-search enumeration follows. That also means the mitigation this page offers — reject or ignore such filter conditions in a custom Resource handler — is the only one available, since nothing in core does it. Worth keeping that sentence as strong as it is.

The trusted-context note matches the model too: server-side `tables.X.get()`/`search()` do not re-apply `allowRead` or `attribute_permissions`, so a computed attribute cross-reading a protected table pulls raw data regardless of the caller's role.

Documenting a side channel you are not fixing is the harder call and the more useful one. `@hidden` and "just don't `@export` it" get reached for as confidentiality controls constantly; saying plainly that they are a spec-visibility directive and a routing directive prevents a whole class of accidental exposure.

reference/components/javascript-environment.md
```diff
@@ -85,6 +85,10 @@ An object whose properties are the tables in the default database (`data`). Each
 
 See [Database API](../database/api.md) for full reference.
 
+:::note Server-side `tables.*` reads run in a trusted context
+Calls to `tables.X.get()`, `tables.X.search()`, and similar methods from within resource code (or Operations API SQL) execute in a **trusted system context** and do **not** re-apply the target table's `allowRead` guard or role-level `attribute_permissions`. This is by design — server-side code is responsible for its own authorization checks. A computed attribute or custom resource that cross-reads a protected table will pull the raw data into the response regardless of the caller's role. If you expose data from a protected table through a custom resource, apply authorization explicitly (e.g. check `getCurrentUser()` and enforce the relevant restrictions before returning the data).
```
heskew: The parenthetical “(or Operations API SQL)” conflates external API requests with trusted in-process table calls. External Operations API requests authenticate, and the SQL dispatcher applies AST permissions unless trusted internal dispatch explicitly sets the bypass. Please remove the parenthetical or name the internal-only invocation you mean; as written, this tells readers that external SQL bypasses RBAC.

---

🤖 Posted by Codex (GPT-5) on behalf of @heskew

reference/database/schema.md
```diff
@@ -174,6 +174,10 @@ type MyTable @table @export(name: "my-table") {
 
 The optional `name` parameter specifies the URL path segment (e.g., `/my-table/`). Without `name`, the type name is used.
 
+:::warning `@export` is a routing directive, not an access control
+Omitting `@export` removes the REST/MQTT route for a table (callers get 404), but it does **not** protect the data. The table still exists in the database and its records remain fully accessible to administrators via the Operations API and SQL. For data confidentiality, use role permissions (`attribute_permissions`, `read: false`) rather than relying on the absence of an export route.
```
heskew: The existing Gemini thread covers which roles can reach this table; there is a separate issue in the remediation. `attribute_permissions` controls fields, so it cannot express table-level confidentiality. For a table, recommend omitting that table’s role grant or setting its table-level `read` permission to `false`; reserve `attribute_permissions` with `read: false` for individual fields. Please make the same table/field distinction in the `@hidden` warning below.

---

🤖 Posted by Codex (GPT-5) on behalf of @heskew

reference/components/javascript-environment.md
kriszyp: Addressed in 48e02d55: the Operations API SQL parenthetical was removed; the trusted-context claim is now limited to direct server-side  and  calls.

PR #560, opened 2026-07-02
docs(security): response-header hardening (app-level controls + #1567 gap)

kriszyp:
## Summary

Adds a dedicated **Response Header Hardening** page to the Security reference (`reference/security/response-headers.md`), framed around the controls that actually work on the primary Harper deployment model — Harper-hosted (Fabric), where there is **no customer-controlled reverse proxy / CDN / edge** to configure.

**Framing (revised after reviewer feedback):**

1. **Lead with the app-level mechanism.** A custom [Resource](../resources/resource-api.md) handler can set response headers via `context.responseHeaders.set(name, value)` (from `getContext()` on HTTP-triggered requests). Example sets `X-Content-Type-Options: nosniff` and notes where frame-options / CSP go. This is accurate to responses that flow through a custom Resource handler.
2. **State the gap clearly.** There is currently no built-in/config way to apply these headers to Harper's **auto-generated REST / table / GraphQL endpoints**, and no global default. Tracked in [HarperFast/harper#1567](https://github.com/HarperFast/harper/issues/1567) — "Provide a built-in way to set security/cache response headers across all HTTP surfaces." Framed as: for now, add headers from custom Resources; global/auto-endpoint coverage is #1567.
3. **Edge config demoted to a fallback.** The reverse-proxy/Nginx guidance is kept but flagged as applicable only to self-hosted deployments (not Fabric), with a trimmed 3-line snippet.
4. **Caching of authenticated responses** reframed around app/platform controls (set `Cache-Control: no-store` from a Resource; key on identity if caching), with the missing built-in cache-control also referencing #1567.
5. **Style fixes:** forward-compat note now addresses engineers directly (per Kris's style guide) instead of a meta reader-note; `X-Frame-Options` guidance kept conservative but explicitly app-specific.

**Placement:** split into its own page rather than wedging into the auth-focused Configuration page. Added to the Reference sidebar (`sidebarsReference.ts`) under Security, and cross-linked from the auth Configuration page and `reference/rest/headers.md`.

## Scope / what's deliberately NOT here

Identified during exploratory QA (D-177 / header-hardening theme) but tracked as separate engineering fixes and intentionally excluded — this PR is guidance only:

| Gap | Tracking issue |
|-----|----------------|
| CORS `Vary: Origin` missing | harper#1518 |
| Auth responses need `Cache-Control`/`Vary: Authorization` at origin | harper#1565 |
| Session cookie `SameSite`/TTL hardening | harper#1564 |
| `304 Not Modified` drops `Vary` | harper#1566 |
| Built-in header/cache-control across all HTTP surfaces | harper#1567 |

## Files changed

- `reference/security/response-headers.md` — **new** page
- `reference/security/configuration.md` — cross-link under Related (previous inline section removed)
- `reference/rest/headers.md` — cross-link in See Also
- `sidebarsReference.ts` — sidebar entry

## Test plan

- [ ] Docusaurus renders the new page; sidebar entry appears under Security
- [ ] Confirm the Resource `responseHeaders.set()` example matches current API (`getContext()` → `responseHeaders`)
- [ ] Verify all cross-links resolve
- [ ] Confirm #1567 gap framing reads accurately for auto-generated REST/GraphQL

---

*Drafted by KrAIs (Kris's QA agent) from exploratory QA campaign findings D-177 + the header-hardening theme. Reworked per reviewer feedback (Fabric hosting has no customer edge). Opened as DRAFT for Kris's review before merging.*

### DIFF
--- a/reference/rest/headers.md
+++ b/reference/rest/headers.md
@@ -95,4 +95,4 @@ This is not recommended for production use — prefer the `Accept` header for cl
 - [REST Overview](./overview.md) — HTTP methods and URL structure
 - [Content Types](./content-types.md) — Supported encoding formats
 - [Security Overview](../security/overview.md) — Authentication headers and mechanisms
-- [HTTP Response Header Hardening](../security/configuration.md#http-response-header-hardening) — Setting browser security headers and cache-control at your edge
+- [Response Header Hardening](../security/response-headers.md) — Adding browser security headers and cache-control to Harper responses (from a custom Resource, or at your own edge)
--- a/reference/rest/headers.md
+++ b/reference/rest/headers.md
@@ -95,3 +95,4 @@ This is not recommended for production use — prefer the `Accept` header for cl
 - [REST Overview](./overview.md) — HTTP methods and URL structure
 - [Content Types](./content-types.md) — Supported encoding formats
 - [Security Overview](../security/overview.md) — Authentication headers and mechanisms
+- [HTTP Response Header Hardening](../security/configuration.md#http-response-header-hardening) — Setting browser security headers and cache-control at your edge
--- a/reference/security/configuration.md
+++ b/reference/security/configuration.md
@@ -62,48 +62,9 @@ Password hashing algorithm used when storing user passwords. Replaced the previo
 - **`sha256`** — Default. Good security and excellent performance.
 - **`argon2id`** — Highest security. More CPU-intensive; recommended for environments that do not require frequent password verifications.
 
-## HTTP Response Header Hardening
-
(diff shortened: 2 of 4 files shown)


DavidCockerill (approved):
Approving. No findings.

The deployment framing is what earns this page its place: **on Fabric there is no customer-controlled reverse proxy, CDN or edge**, so "set it at your edge" is not advice a Harper-hosted customer can act on. Leading with the custom-Resource path and treating the edge as the fallback is the right order — most hardening docs do the reverse and send readers off to configure infrastructure they do not have.

Showing `getContext().responseHeaders.set(name, value)` concretely rather than describing it is the difference between a page someone acts on and one they bookmark.

PR #563, opened 2026-07-02
Document loadedFromSource cache disposition on the request context

### LINKED ISSUE #1571: loadedFromSource is unobservable with string-id Table.get() — set on an internal RequestTarget; Context declares it but it is never assigned
## Summary

For a caching table (`sourcedFrom()`), the cache-disposition flag `loadedFromSource` is set on the **RequestTarget of the `get` call** — and when the caller passes a plain string id (the most common calling convention), the static-get dispatch mints an *internal* `RequestTarget` the caller never sees. The flag is faithfully set on that internal object and discarded. There is no way to observe hit/miss from resource code with a string id.

Compounding it: the `Context` interface *declares* the field (`resources/ResourceInterface.ts:98` — `loadedFromSource?: boolean`), which implies it should be readable from the request context, but **no code path ever assigns it to a context**. Reading `context.loadedFromSource` after `Cache.get(stringId)` silently yields `undefined`.

## Impact

Any app computing hit/miss this way records **every lookup as a cache hit** — silently wrong data, no error. In StormScout this was a latent day-one bug: every `Server-Timing … desc=` value and stored cache-status row said `hit`, including 2–3 s upstream source fetches.

## Repro

```ts
// resource method, table wired with sourcedFrom()
await Cache.get(key);                                        // string id
console.log((this.getContext() as any).loadedFromSource);    // undefined — always

import { RequestTarget } from 'harper';
const t = new RequestTarget();
t.id = key;
await Cache.get(t);
console.log(t.loadedFromSource);   // true on source fetch, false on cache hit — works
```

Verbatim probe output (5.1.6 and 5.1.15, identical):

```
"f1_context_flag_after_string_get": null,
"f1_target_flag_after_target_get": false
```

## Code locations (5.1.15)

- `resources/Resource.ts:637` — string-id branch of static `get` dispatch: `query = new RequestTarget(); query.id = id;` (internal target, never returned to the caller)
- `resources/Table.ts:4804` — `target.loadedFromSource = true` (this request acquired the resolution lock and fetched from source)
- `resources/Table.ts:4789` — `= false` after waiting on another request's in-flight fetch (dedupe join — so "hit" with a multi-second duration is expected behavior, worth documenting)
- `resources/Table.ts:730` / `:746` — `= false` on plain cache hits
- `resources/ResourceInterface.ts:98` — declared on `Context`, never assigned

## Suggested fix

Mirror the flag onto the request context (the interface already declares it) when it is set on the target — or document the explicit-`RequestTarget` pattern as the only supported way to observe cache disposition from resource code.

## Environment

Harper 5.1.6, re-verified identical on 5.1.15. Node v24.16.0 (type-stripped TS resources), macOS, single instance, `harper dev`.

kylebernhardy:
Companion docs for HarperFast/harper#1575 (fixes HarperFast/harper#1571), which makes cache disposition observable on caching-table gets via `context.loadedFromSource` / `target.loadedFromSource`.

## Changes

- New **Observing cache disposition** subsection under `sourcedFrom()` in `reference/resources/resource-api.md`, documenting the settled semantics verified in the fix: `true` when the get went to the source (including `staleIfError` fallbacks to stale data); `false` for cache hits, `onlyIfCached`, stale-while-revalidate responses, and dedupe-join waits (with the note that such a "hit" can still take as long as the upstream fetch); last get in a context wins.
- `loadedFromSource` listed in both `getContext(): Context` property references (version-gated "5.1.16+"), cross-linked to the new subsection.
- Per the decision on HarperFast/harper#1577 (closed), `wasLoadedFromSource()` is **not** implemented and its reference section is removed (commit 0862bda by @kriszyp) — the context/target API is the single supported way to observe disposition.

## Note for review

The text says "5.1.16" on the assumption harper#1575 ships in the next 5.1 patch. **Hold merging until harper#1575 lands** and confirm the actual patch version then.

_Generated by an LLM (Claude Fable 5), cross-reviewed by Codex and Gemini; all findings addressed. Includes a commit from @kriszyp._

### DIFF
--- a/reference/resources/resource-api.md
+++ b/reference/resources/resource-api.md
@@ -264,7 +264,7 @@ Within a resource method, the same value is available on the active context via 
 - `false` — the record was served from the cache: fresh hits, `onlyIfCached` requests, stale-while-revalidate responses (the source fetch continues in the background), and requests that waited on another request's in-flight fetch of the same record. This last case means a cache hit can still take as long as an upstream fetch.
 - Each get on a caching table in the same context overwrites the value, so read it after the `get` you are measuring.
 
-Resource instances also expose this as [`wasLoadedFromSource()`](#wasloadedfromsource-boolean). Prior to Harper 5.1.16, the flag was only observable via an explicitly passed `RequestTarget`; `context.loadedFromSource` was never assigned.
+Note that `get()` returns a plain `RecordObject`, not a resource instance — the record itself does not carry cache disposition; read it from the context (or an explicitly passed `RequestTarget`). Prior to Harper 5.1.16, the flag was only observable via an explicitly passed `RequestTarget`; `context.loadedFromSource` was never assigned.
 
 #### Source `get` — controlling timestamp and expiration
 
--- a/reference/resources/resource-api.md
+++ b/reference/resources/resource-api.md
@@ -255,12 +255,12 @@ Each `get` on a caching table records whether the record came from the cache or 
 ```javascript
 const context = {};
 const record = await MyCache.get(recordId, context);
-console.log(context.loadedFromSource); // true = fetched from source, false = served from cache
+console.log(context.loadedFromSource); // true = went to the source, false = served from cache
 ```
 
 Within a resource method, the same value is available on the active context via `getContext().loadedFromSource` after the `get` resolves. The flag settles as follows:
 
-- `true` — the get fetched the record from the source, including when a source error fell back to a stale record (`staleIfError`).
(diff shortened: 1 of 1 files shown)


kriszyp (approved):
Approving — verified line-for-line accurate against harper#1575 / #1577's actual semantics (`loadedFromSource` / `wasLoadedFromSource()`). Since it pairs with those core PRs (v5.1-targeted), just land it alongside them and confirm the version marker. Thanks @kylebernhardy!

— 🤖 KrAIs (Kris's review assistant)

PR #564, opened 2026-07-02
Document config-shaping env vars delivered via loadEnv .env files

### LINKED ISSUE #1513: MCP application config from loadEnv (.env) not applied — config composed/memoized before components run
## Summary

When a component delivers a config-shaping environment variable (`HARPER_CONFIG` / `HARPER_SET_CONFIG` / `HARPER_DEFAULT_CONFIG`) through the documented `loadEnv` (`.env`) mechanism, the value is **read too late to affect the composed configuration**. The root configuration is composed and memoized by `getConfigObj()` during the *root* component load, which runs **before** any component's `loadEnv` plugin sets `process.env`. As a result, config keys delivered via `.env` never reach the config tree.

A concrete, user-visible symptom: a `mcp.application` block delivered via `.env` does **not** mount the MCP application profile (`/mcp`), whereas the byte-identical block delivered as a real process env var **does**.

This is distinct from #803 (`.env` not loaded at all on deployed servers) — here the `.env` **is** read (the `loadEnv` debug line fires), it is simply applied after the config has already been frozen.

## Environment

- Harper **5.1.14** (installed from `@harperfast/harper@5.1.14`, running `dist/`)
- Node v22+ (repro run on macOS, single thread)
- Minimal, hand-written component (no third-party code)

## Minimal reproduction

A trivial component directory (`config.yaml` declares `loadEnv` first, plus one no-op resource so HTTP binds), run via `harper run <appDir>`:

**Variant A — config delivered via `.env` / `loadEnv` (BROKEN):**

`config.yaml`
```yaml
loadEnv:
  files: .env
jsResource:
  files: res/*.js
```
`.env`
```
HARPER_CONFIG={"mcp":{"application":{"mountPath":"/mcp"}}}
```
`HARPER_CONFIG` is **not** present in the spawned process environment — it is delivered only through `.env`.

**Variant B — same block delivered as a real process env var (WORKS):**

`config.yaml`
```yaml
jsResource:
  files: res/*.js
```
With `HARPER_CONFIG={"mcp":{"application":{"mountPath":"/mcp"}}}` exported in the process that launches Harper.

Both variants share an identical base config (via `HARPER_SET_CONFIG` in the process env) that pins ports and sets `rest: true`, so the HTTP listener binds in both cases. The only variable under test is *how* the `mcp` block is delivered.

Probe after boot:
```
POST /mcp   (JSON-RPC "initialize")
```

## Expected vs. actual

| | `GET /` (listener up?) | `POST /mcp` |
|---|---|---|
| **Variant B** (process env var) | `404` (listener up) | **`200`** — MCP profile mounted, `serverInfo: harper-mcp 5.1.14` |
| **Variant A** (`.env` / `loadEnv`) | `404` (listener up) | **`404`** — MCP profile never mounted |

In Variant A the listener is up (`GET /` answers `404`), and the `.env` is provably read — the worker logs:
```
[http/1] [debug]: Loading env file: <appDir>/.env
```
…yet `/mcp` is never registered. **Expected:** Variant A should behave like Variant B (the documented `loadEnv` mechanism should deliver `HARPER_CONFIG` such that `mcp.application` is applied and `/mcp` mounts).

## Root cause (source trace, line numbers from installed `dist/` of 5.1.14)

1. **The config is composed and memoized on the first `getConfigObj()` call.**
   `dist/config/configUtils.js`:
   - module-level cache `let configObj;` (L36)
   - `getConfigObj()` (L885) calls `initConfig()` only when `!configObj`, then returns the cache thereafter
   - `initConfig()` applies the env vars via `applyRuntimeEnvVarConfig(configDoc, configFilePath)` (L326), which reads `process.env.HARPER_CONFIG` (L767)
   - `flattenConfig(obj)` assigns the module-level cache `configObj = obj` (L623), reached from `initConfig` at L331
   So whatever `process.env.HARPER_CONFIG` is **at the moment of the first `getConfigObj()` call** is baked in; later mutations are ignored.

2. **The first `getConfigObj()` happens during the *root* component load, before any user component (and its `loadEnv`) runs.**
   `dist/server/loadRootComponents.js`: `loadComponent(<root>, …, { isRoot: true })` (L28-32) runs first, then `loadComponentDirectories(...)` (L35) loads the app folders afterward.
   `dist/components/componentLoader.js`: for the root, `config = getConfigObj()` (L339) — first call → composes & memoizes.

3. **The MCP application profile reads the memoized config.**
   `dist/components/mcp/index.js`: `handleApplication(scope)` (L96) → `const config = getConfigObj()` (L99) → `if (!config?.mcp?.application) { … return; }` (L100). `mcp` is a built-in trusted plugin (`componentLoader.js` L150 `mcp: mcpComponent`), so this runs against the already-frozen root config. With no `mcp` block present at that point, it returns without registering `scope.server.http(... urlPath: '/mcp' ...)` (L110).

4. **`loadEnv` sets `process.env` during component load — after the config is frozen.**
   `dist/resources/loadEnv.js`: `handleApplication(scope)` (L9) parses the `.env` and does `process.env[key] = value` (L27). This is invoked as part of loading the component that declares it (a non-root load reached via `loadComponentDirectories`), i.e. strictly after step 2/3.

**Causal chain:** `getConfigObj()` memoizes the config at root load (pre-`loadEnv`) → the MCP plugin reads that frozen config and finds no `mcp` block → `loadEnv` then sets `HARPER_CONFIG` in `process.env`, but nothing re-composes the config → `/mcp` is never mounted.

(TypeScript-source equivalents for reference: `config/configUtils.js` `getConfigObj`, `configObj = obj` in `flattenConfig`, `applyRuntimeEnvVarConfig`; `components/componentLoader.ts:339`; `components/mcp/index.ts:122-133`; `resources/loadEnv.ts`; `server/loadRootComponents.js`.)

## Proposed fix

The general issue is that config-shaping env vars delivered via `loadEnv` land after the config is composed. Options, roughly in order of preference:

1. **Apply `loadEnv` before config composition.** If `loadEnv` were processed early (e.g. as a pre-pass over component `.env` files before the first `getConfigObj()` / before the root component loads), then `HARPER_CONFIG` from `.env` would be present when the config tree is composed — making `.env` and process-env delivery equivalent (least surprise, and aligns with #1214 positioning `HARPER_CONFIG` as the recommended mergeable config var).
2. **Re-compose config after `loadEnv` mutates `process.env`.** Detect that a config-relevant env var changed during component load and re-run the env-var application (or invalidate the `configObj` memo and re-`initConfig()`), then let dependent consumers (the MCP profile, etc.) re-read.
3. **At minimum, make MCP application config read lazily / after component composition** so a `mcp` block delivered late still mounts — though this only addresses the MCP symptom, not the general class.

If `loadEnv`-delivered config is intentionally out of scope (i.e. only process-env / config-file delivery is supported for config-shaping vars), a documentation note would help, since `HARPER_CONFIG` via `.env` silently no-ops today rather than erroring.

## Context

We hit this enabling a native MCP surface for a component whose configuration is delivered through the documented `.env` / `loadEnv` mechanism. The block is valid and is read at runtime — it just arrives after the config has been frozen, so the MCP profile never mounts. Happy to provide the full self-contained repro harness if useful.

kylebernhardy:
Companion docs for HarperFast/harper#1580 (fixes HarperFast/harper#1513) — **pivoted with the PR**: config-shaping env vars delivered via a component's `.env` are **not honored** (config is strictly top-down); what 5.2.0 adds is the loud, actionable warning replacing the old silent no-op. The section documents the behavior, the warning, and the supported channels.

**Hold merging until harper#1580 lands.**

### DIFF
--- a/reference/environment-variables/overview.md
+++ b/reference/environment-variables/overview.md
@@ -78,6 +78,19 @@ loadEnv:
 
 Files are loaded in the order specified.
 
+## Config-Shaping Variables
+
+<VersionBadge type="changed" version="v5.2.0" />
+
+The three configuration env vars — `HARPER_DEFAULT_CONFIG`, `HARPER_CONFIG`, and `HARPER_SET_CONFIG` (see [Configuration](../configuration/overview.md)) — shape Harper's root configuration, which is composed once at startup, before components load. When one of these is delivered through a `loadEnv` `.env` file, Harper applies it before composing the configuration, so `.env` delivery behaves the same as setting a real process environment variable. (Prior to 5.2.0, these variables silently had no effect when delivered via `.env`, because they were read after the configuration had already been composed.)
+
+Notes and limitations:
+
+- A real process environment variable still takes precedence over the `.env` value, unless the component declares `override: true`.
+- Encrypted (`enc:v1:`) values cannot shape configuration — secret decryptors are not registered until components load. Such values are skipped with an error log.
+- `componentsRoot` cannot be redirected this way: Harper discovers the `.env` files by scanning the components root from the config file, so a `componentsRoot` override delivered via env var or `.env` cannot change where that scan looks.
+- Only these three variables are applied early. All other `.env` keys load at component-load time as usual.
+
 ## Related
 
 - [Components Overview](../components/overview.md)
--- a/reference/environment-variables/overview.md
+++ b/reference/environment-variables/overview.md
@@ -84,7 +84,7 @@ Files are loaded in the order specified.
 
 The three configuration env vars — `HARPER_DEFAULT_CONFIG`, `HARPER_CONFIG`, and `HARPER_SET_CONFIG` (see [Configuration](../configuration/overview.md)) — shape Harper's root configuration, which is composed once at startup, **before components load**. Delivering one of them through a `loadEnv` `.env` file therefore has **no effect**: configuration is strictly top-down (the instance's configuration controls components, never the reverse), so a component cannot shape instance-wide config.
 
-As of 5.2.0 this is no longer silent: Harper logs a prominent warning at startup for each config-shaping variable found in a component `.env` file — naming the variable and the file — and again at component load time (which also covers components deployed after startup). Earlier versions ignored these variables without any message.
+As of 5.2.0 (and 5.1.x patch releases from 5.1.18) this is no longer silent: Harper logs a prominent warning at startup for each config-shaping variable found in a component `.env` file — naming the variable and the file — and again at component load time (which also covers components deployed after startup). Earlier versions ignored these variables without any message.
 
 To configure the instance, use the supported channels instead: set the variable in the process environment (container env, service unit, shell), or put the equivalent keys in the instance's `harper-config.yaml`.
 
--- a/reference/environment-variables/overview.md
+++ b/reference/environment-variables/overview.md
@@ -78,18 +78,15 @@ loadEnv:
 
 Files are loaded in the order specified.
 
-## Config-Shaping Variables
+## Config-Shaping Variables Are Not Honored
 
 <VersionBadge type="changed" version="v5.2.0" />
 
-The three configuration env vars — `HARPER_DEFAULT_CONFIG`, `HARPER_CONFIG`, and `HARPER_SET_CONFIG` (see [Configuration](../configuration/overview.md)) — shape Harper's root configuration, which is composed once at startup, before components load. When one of these is delivered through a `loadEnv` `.env` file, Harper applies it before composing the configuration, so `.env` delivery behaves the same as setting a real process environment variable. (Prior to 5.2.0, these variables silently had no effect when delivered via `.env`, because they were read after the configuration had already been composed.)
+The three configuration env vars — `HARPER_DEFAULT_CONFIG`, `HARPER_CONFIG`, and `HARPER_SET_CONFIG` (see [Configuration](../configuration/overview.md)) — shape Harper's root configuration, which is composed once at startup, **before components load**. Delivering one of them through a `loadEnv` `.env` file therefore has **no effect**: configuration is strictly top-down (the instance's configuration controls components, never the reverse), so a component cannot shape instance-wide config.
 
-Notes and limitations:
+As of 5.2.0 this is no longer silent: Harper logs a prominent warning at startup for each config-shaping variable found in a component `.env` file — naming the variable and the file — and again at component load time (which also covers components deployed after startup). Earlier versions ignored these variables without any message.
 
-- A real process environment variable still takes precedence over the `.env` value, unless `loadEnv` is configured with `override: true`.
-- Encrypted (`enc:v1:`) values cannot shape configuration — secret decryptors are not registered until components load. Such values are skipped with an error log.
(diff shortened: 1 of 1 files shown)


reference/environment-variables/overview.md
```diff
@@ -78,6 +78,19 @@ loadEnv:
...
 
+## Config-Shaping Variables
+
+<VersionBadge type="changed" version="v5.2.0" />
+
+The three configuration env vars — `HARPER_DEFAULT_CONFIG`, `HARPER_CONFIG`, and `HARPER_SET_CONFIG` (see [Configuration](../configuration/overview.md)) — shape Harper's root configuration, which is composed once at startup, before components load. When one of these is delivered through a `loadEnv` `.env` file, Harper applies it before composing the configuration, so `.env` delivery behaves the same as setting a real process environment variable. (Prior to 5.2.0, these variables silently had no effect when delivered via `.env`, because they were read after the configuration had already been composed.)
+
+Notes and limitations:
+
+- A real process environment variable still takes precedence over the `.env` value, unless the component declares `override: true`.
```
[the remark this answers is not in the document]
kylebernhardy: Applied in 938746d — now reads "unless `loadEnv` is configured with `override: true`".

reference/environment-variables/overview.md
```diff
@@ -78,6 +78,16 @@ loadEnv:
 
 Files are loaded in the order specified.
 
+## Config-Shaping Variables Are Not Honored
```
kriszyp, line 81: This guidance is now already present on `origin/main` in the `loadEnv` warning added by #579 (current main lines 49-57). A three-way merge keeps both insertions, so the page would explain the same rule and workaround twice. Could we rebase and fold only the uniquely useful warning-timing detail into the existing warning, or close this PR as superseded?

reference/environment-variables/overview.md
```diff
@@ -78,6 +78,16 @@ loadEnv:
...
+
+<VersionBadge type="changed" version="v5.2.0" />
+
+The three configuration env vars — `HARPER_DEFAULT_CONFIG`, `HARPER_CONFIG`, and `HARPER_SET_CONFIG` (see [Configuration](../configuration/overview.md)) — shape Harper's root configuration, which is composed once at startup, **before components load**. Delivering one of them through a `loadEnv` `.env` file therefore has **no effect**: configuration is strictly top-down (the instance's configuration controls components, never the reverse), so a component cannot shape instance-wide config.
+
+As of 5.2.0 (and 5.1.x patch releases from 5.1.18) this is no longer silent: Harper logs a prominent warning at startup for each config-shaping variable found in a component `.env` file — naming the variable and the file — and again at component load time (which also covers components deployed after startup). Earlier versions ignored these variables without any message.
```
kriszyp, line 87: The v5.1 backport first shipped in `v5.1.19`, not `v5.1.18`: merge `2fff82731` is not contained by the `v5.1.18` tag, while it is contained by `v5.1.19` (`componentEnvPrepass.ts` is likewise absent in 5.1.18 and present in 5.1.19). Please change this to `from 5.1.19` so 5.1.18 operators are not told to expect warnings their version cannot emit.

kriszyp (commented):
I think you are filing a duplicate PR?

PR #565, opened 2026-07-03
docs: make target.loadedFromSource canonical, drop context mirror

kriszyp:
Follow-up to #563 (merged). Companion to core PR HarperFast/harper#1590.

#563 documented cache disposition as living on **both** the request context and the `RequestTarget`, and led with a `context.loadedFromSource` example. That framing doesn't hold up: `loadedFromSource` is a **per-`get()` result**, and the context is shared for the whole request — so `context.loadedFromSource` is last-write-wins (a nested caching `get()` inside a resource's own `get()` clobbers the entry get's value). The core PR removes the context mirror and standardizes on `target.loadedFromSource` (which is what the REST handler actually reads).

This reframes the **Observing cache disposition** section:
- Lead with `target.loadedFromSource` as the canonical, per-get signal, with a `RequestTarget`-based example.
- Keep the `true`/`false` settle-semantics (unchanged).
- Remove the `context.loadedFromSource` example, the "available on the active context" paragraph, and the two `getContext()` bullets that listed `loadedFromSource`.

Depends on HarperFast/harper#1590 shipping (which removes `context.loadedFromSource`). Safe to merge alongside it.

_Authored by KrAIs (Claude Opus 4.8), at Kris's direction._

### DIFF
--- a/reference/resources/resource-api.md
+++ b/reference/resources/resource-api.md
@@ -250,21 +250,23 @@ Harper automatically serializes concurrent requests for the same missing or stal
 
 #### Observing cache disposition
 
-Each `get` on a caching table records whether the record came from the cache or from the source, in the `loadedFromSource` property of both the request context and the `RequestTarget`:
+Whether a `get` on a caching table was served from cache or loaded from the source is a **per-`get` result**, recorded on the `RequestTarget` for that call as `target.loadedFromSource`. Because each `get()` has its own target, the value is precise to that one call and unaffected by any nested or subsequent gets — it is what Harper's REST handler reads to set cache response headers.
 
 ```javascript
-const context = {};
-const record = await MyCache.get(recordId, context);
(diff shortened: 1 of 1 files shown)


dawsontoth (approved):

Ethan-Arrowood (approved):
Content looks good — I verified the new example against harper main: `RequestTarget` is exported from the `harper` module, `Table.ts` sets `target.loadedFromSource`, and the REST handler reads exactly that to set cache headers. Approving on the content.

One sequencing flag though: the PR body says this depends on HarperFast/harper#1590, but that PR was **closed without merging** — the live successor is HarperFast/harper#1626, which is still open. Current harper main still assigns `context.loadedFromSource` (`Table.ts:4684`), so please **hold off merging this until #1626 actually lands** (and maybe update the body's reference to point at #1626). Once it ships, this is good to go as-is.

PR #567, opened 2026-07-06
Document MCP custom content resources (mcpResources) and the harper+rest:// descriptor scheme

### LINKED ISSUE #1609: MCP application profile: no hook to expose custom content resources (only Resource descriptors + harper:// metadata)
## Summary
The MCP **application** profile has no component-author hook to expose **custom content resources** — arbitrary `resources/read`-able `text`/`blob` content under an author-chosen URI. This blocks a real use case: a public docs server exposing each documentation page as an MCP resource (`docs:///{+path}` → the page Markdown).

The 2025-06-18 resources spec explicitly supports this: a resource may be "files, database schemas, or **application-specific information**"; `resources/read` returns `text` or `blob`; and "implementations are always free to use additional, custom URI schemes."

## Current behaviour (`harper@5.1.15`)
Resources are "discovered, not registered" (`components/mcp/resources.ts`). On a live `application`-profile instance, `resources/list` returns **only**:
- `harper://about`, `harper://openapi`, `harper://schema/{db}/{table}` — server metadata; and
- `http(s)://<host>/<path>` — a **descriptor** for each exported Resource (returns `{ uri, path, database, table, hint }`, *not* content; the hint just points back to a `get_*`/`search_*` tool).

There is no `resources` equivalent of `static mcpTools`. Modelling hundreds of content pages as Resource classes is a poor fit, and the descriptor never carries the body.

## Ask
A component-author hook to register custom resource URIs / templates with an author-supplied `read` returning `text`/`blob` content — parallel to how `mcpTools` opens the tool surface (e.g. a static `mcpResources` / `mcpResourceTemplates`, or a registration API).

## Minor, related
Exported-Resource descriptors use the `http(s)://` scheme, but the spec says use `https://` "only when the client is able to fetch and load the resource directly from the web on its own." These are in-process, RBAC-gated descriptors, so a custom scheme would fit the spec guidance better.

## Context
Found while evaluating whether a public docs MCP server (tools `search_docs`, `fetch_doc`, LLM-backed `answer`) could move from a hand-rolled JSON-RPC server onto the native `application` profile. The tool surface maps cleanly via `mcpTools` (incl. anonymous access); custom content resources are the one capability with no native path.

kylebernhardy:
Companion docs for HarperFast/harper#1613 (fixes HarperFast/harper#1609).

## Changes to `reference/mcp/tools-and-resources.md`

- New **Custom `mcpResources` opt-in** section (parallel to Custom `mcpTools`): fixed URIs vs `uriTemplate` (`{name}`/`{+name}` semantics), content return shapes (text/blob/JSON), live-class dispatch + RBAC delegation, per-parameter `completions`, reserved-scheme rule, sanitized error behavior.
- **`https://` URIs section → `harper+rest://`** with the rationale (spec reserves https for web-fetchable resources) and the back-compat note for legacy URIs.
- New **Custom content URIs** subsection tying the resources surface together (custom-match precedence).

## Note for review

Version references say "5.1.16+" assuming harper#1613 ships in the next 5.1 patch — confirm at merge time. **Hold merging until harper#1613 lands.**

### DIFF
--- a/reference/mcp/tools-and-resources.md
+++ b/reference/mcp/tools-and-resources.md
@@ -146,7 +146,7 @@ Each entry declares exactly one of `uri` (fixed — listed by `resources/list`) 
 
 Notes:
 
-- Reads dispatch on the **live** registry class, so an exported `resources.js` subclass's method (and its access control) always wins — the same rule as custom `mcpTools`. Authentication is "is the user logged in"; finer-grained gating is the method's responsibility.
+- Reads dispatch on the **live** registry class, so an exported `resources.js` subclass's method (and its access control) always wins — the same rule as custom `mcpTools`. Custom resources are served to **any** MCP session, including anonymous, unauthenticated ones (the public-docs case this feature targets) — the MCP layer performs no auth check for them; to restrict one, check `context.user` in the `read` method and throw.
 - `completions` optionally declares candidate values per template parameter, served by `completion/complete`.
 - Custom URIs must use an author-chosen scheme (`docs:///…` above). The reserved schemes — `harper:`, `harper+rest:`, `http:`, `https:` — are rejected at registration so custom entries cannot shadow the built-in surfaces.
 - A read error from the method surfaces to the client as a sanitized JSON-RPC error; the raw error is written to the server log.
--- a/reference/mcp/tools-and-resources.md
+++ b/reference/mcp/tools-and-resources.md
@@ -110,7 +110,7 @@ The corresponding instance method runs through Harper's normal `transactional()`
 
 ### Custom `mcpResources` opt-in
 
-A component author can expose arbitrary content — documentation pages, rendered reports, any `text` or `blob` payload — as MCP resources under author-chosen URIs by declaring a static `mcpResources` array (5.1.16+):
+A component author can expose arbitrary content — documentation pages, rendered reports, any `text` or `blob` payload — as MCP resources under author-chosen URIs by declaring a static `mcpResources` array (5.1.18+):
 
 ```ts
 class DocsPages extends Resource {
@@ -178,7 +178,7 @@ The schema URIs honor each user's `permission[db].tables[table]` walk — a user
 
 ### `harper+rest://` URIs
 
-The application profile additionally exposes every exported `Resource` (that passes the `exportTypes.mcp` gate **and** the `hasRestVerbs` check) as a `harper+rest://<host>:<port>/<path>` URI (5.1.16+ — earlier releases listed these under `http(s)://`, which the MCP spec reserves for resources a client can fetch directly from the web; legacy `http(s)://` URIs continue to work for `resources/read` and `resources/subscribe`). These resolve in-process via `Resources.getMatch(path, 'mcp')` — there is no outbound HTTP request. The body returned by `resources/read` is a small descriptor:
+The application profile additionally exposes every exported `Resource` (that passes the `exportTypes.mcp` gate **and** the `hasRestVerbs` check) as a `harper+rest://<host>:<port>/<path>` URI (5.1.18+ — earlier releases listed these under `http(s)://`, which the MCP spec reserves for resources a client can fetch directly from the web; legacy `http(s)://` URIs continue to work for `resources/read` and `resources/subscribe`). These resolve in-process via `Resources.getMatch(path, 'mcp')` — there is no outbound HTTP request. The body returned by `resources/read` is a small descriptor:
 
 ```json
 {
--- a/reference/mcp/tools-and-resources.md
+++ b/reference/mcp/tools-and-resources.md
@@ -108,6 +108,49 @@ class Orders extends Tables.orders {
 
 The corresponding instance method runs through Harper's normal `transactional()` envelope, so per-record `allow*` predicates and audit logging behave the same way as regular verb dispatch. Authentication is "is the user logged in" only — finer-grained gating is the method's responsibility.
 
+### Custom `mcpResources` opt-in
+
+A component author can expose arbitrary content — documentation pages, rendered reports, any `text` or `blob` payload — as MCP resources under author-chosen URIs by declaring a static `mcpResources` array (5.1.16+):
+
+```ts
+class DocsPages extends Resource {
+	static mcpResources = [
+		{
+			uri: 'docs:///index',
+			name: 'docs index',
+			description: 'List of all documentation pages',
+			mimeType: 'text/markdown',
+			method: 'readIndex',
+		},
+		{
+			uriTemplate: 'docs:///{+path}',
+			name: 'docs page',
+			description: 'A documentation page by path',
+			mimeType: 'text/markdown',
+			method: 'readPage',
+			completions: { path: ['guides/install.md', 'guides/deploy.md'] },
+		},
+	];
+
+	async readIndex() {
+		return { text: '- docs:///guides/install.md\n…', mimeType: 'text/markdown' };
+	}
+
+	async readPage(params) {
+		return { text: loadPage(params.path), mimeType: 'text/markdown' };
+	}
+}
+```
+
+Each entry declares exactly one of `uri` (fixed — listed by `resources/list`) or `uriTemplate` (listed by `resources/templates/list`). Templates use `{name}` to match a single path segment and `{+name}` to match across segments; `resources/read` extracts the parameters and passes them to the named instance method as its first argument. The method returns a string (text content), `{ text, mimeType? }`, `{ blob, mimeType? }` (base64 binary), or any other object (serialized as JSON).
+
+Notes:
+
+- Reads dispatch on the **live** registry class, so an exported `resources.js` subclass's method (and its access control) always wins — the same rule as custom `mcpTools`. Authentication is "is the user logged in"; finer-grained gating is the method's responsibility.
+- `completions` optionally declares candidate values per template parameter, served by `completion/complete`.
+- Custom URIs must use an author-chosen scheme (`docs:///…` above). The reserved schemes — `harper:`, `harper+rest:`, `http:`, `https:` — are rejected at registration so custom entries cannot shadow the built-in surfaces.
+- A read error from the method surfaces to the client as a sanitized JSON-RPC error; the raw error is written to the server log.
+
 ### `exportTypes` gating
 
 The MCP surface mirrors the public REST surface. A Resource is filtered out of MCP enumeration entirely when its registration sets `exportTypes.mcp = false`:
@@ -133,13 +176,13 @@ Both profiles serve `resources/list`, `resources/read`, and `resources/templates
 
 The schema URIs honor each user's `permission[db].tables[table]` walk — a user with no `read` or `describe` perm on a table gets a "permission denied" response from `resources/read`.
 
-### `https://` URIs
+### `harper+rest://` URIs
 
-The application profile additionally exposes every exported `Resource` (that passes the `exportTypes.mcp` gate **and** the `hasRestVerbs` check) as an `https://<host>:<port>/<path>` URI. These resolve in-process via `Resources.getMatch(path, 'mcp')` — there is no outbound HTTP request. The body returned by `resources/read` is a small descriptor:
+The application profile additionally exposes every exported `Resource` (that passes the `exportTypes.mcp` gate **and** the `hasRestVerbs` check) as a `harper+rest://<host>:<port>/<path>` URI (5.1.16+ — earlier releases listed these under `http(s)://`, which the MCP spec reserves for resources a client can fetch directly from the web; legacy `http(s)://` URIs continue to work for `resources/read` and `resources/subscribe`). These resolve in-process via `Resources.getMatch(path, 'mcp')` — there is no outbound HTTP request. The body returned by `resources/read` is a small descriptor:
 
 ```json
 {
-	"uri": "https://node.example.com:9926/Product",
+	"uri": "harper+rest://node.example.com:9926/Product",
 	"path": "Product",
 	"database": "data",
 	"table": "product",
@@ -148,6 +191,10 @@ The application profile additionally exposes every exported `Resource` (that pas
 ```
 
 Per-record reads go through the tools surface, where each Resource's `allow{Read,…}` predicates run. The `resources/read` descriptor itself is a fast, side-effect-free hint — not a capability.
+
+### Custom content URIs
+
+Author-declared `mcpResources` entries (see [Custom `mcpResources` opt-in](#custom-mcpresources-opt-in)) appear alongside the built-in surfaces: fixed URIs in `resources/list`, templates in `resources/templates/list`. A registered custom URI always wins over the discovered surfaces on `resources/read`.
 
 ## `notifications/*/list_changed`
 


reference/mcp/tools-and-resources.md
```diff
@@ -108,6 +108,49 @@ class Orders extends Tables.orders {
 
 The corresponding instance method runs through Harper's normal `transactional()` envelope, so per-record `allow*` predicates and audit logging behave the same way as regular verb dispatch. Authentication is "is the user logged in" only — finer-grained gating is the method's responsibility.
 
+### Custom `mcpResources` opt-in
+
+A component author can expose arbitrary content — documentation pages, rendered reports, any `text` or `blob` payload — as MCP resources under author-chosen URIs by declaring a static `mcpResources` array (5.1.18+):
```
kriszyp, line 113: Could we use the repository's standard availability annotation here? Please add `<VersionBadge version="v5.1.18" />` on its own line below the heading and remove `(5.1.18+)` from this sentence. That keeps this patch-level addition consistent with the neighboring v5.1 feature documentation.

reference/mcp/tools-and-resources.md
```diff
@@ -108,6 +108,49 @@ class Orders extends Tables.orders {
...
+			name: 'docs page',
+			description: 'A documentation page by path',
+			mimeType: 'text/markdown',
+			method: 'readPage',
+			completions: { path: ['guides/install.md', 'guides/deploy.md'] },
+		},
+	];
+
+	async readIndex() {
+		return { text: '- docs:///guides/install.md\n…', mimeType: 'text/markdown' };
```
kriszyp, line 136: Our content style prefers plain ASCII when typography is not meaningful. Could this sample use `...` instead of the Unicode ellipsis, and make the same replacement in the `docs:///…` example below? That also makes the copied examples easier to type and search.

reference/mcp/tools-and-resources.md
```diff
@@ -133,13 +176,13 @@ Both profiles serve `resources/list`, `resources/read`, and `resources/templates
 
 The schema URIs honor each user's `permission[db].tables[table]` walk — a user with no `read` or `describe` perm on a table gets a "permission denied" response from `resources/read`.
 
-### `https://` URIs
+### `harper+rest://` URIs
 
-The application profile additionally exposes every exported `Resource` (that passes the `exportTypes.mcp` gate **and** the `hasRestVerbs` check) as an `https://<host>:<port>/<path>` URI. These resolve in-process via `Resources.getMatch(path, 'mcp')` — there is no outbound HTTP request. The body returned by `resources/read` is a small descriptor:
+The application profile additionally exposes every exported `Resource` (that passes the `exportTypes.mcp` gate **and** the `hasRestVerbs` check) as a `harper+rest://<host>:<port>/<path>` URI (5.1.18+ — earlier releases listed these under `http(s)://`, which the MCP spec reserves for resources a client can fetch directly from the web; legacy `http(s)://` URIs continue to work for `resources/read` and `resources/subscribe`). These resolve in-process via `Resources.getMatch(path, 'mcp')` — there is no outbound HTTP request. The body returned by `resources/read` is a small descriptor:
```
kriszyp, line 181: Because this is a change to the existing descriptor URI behavior, please add `<VersionBadge type="changed" version="v5.1.18" />` on its own line below the heading and remove the inline `(5.1.18+)` marker. This uses the repository's standard distinction between a newly added surface and changed behavior.

reference/mcp/tools-and-resources.md
```diff
@@ -108,6 +108,49 @@ class Orders extends Tables.orders {
...
+		return { text: loadPage(params.path), mimeType: 'text/markdown' };
+	}
+}
+```
+
+Each entry declares exactly one of `uri` (fixed — listed by `resources/list`) or `uriTemplate` (listed by `resources/templates/list`). Templates use `{name}` to match a single path segment and `{+name}` to match across segments; `resources/read` extracts the parameters and passes them to the named instance method as its first argument. The method returns a string (text content), `{ text, mimeType? }`, `{ blob, mimeType? }` (base64 binary), or any other object (serialized as JSON).
+
+Notes:
+
+- Reads dispatch on the **live** registry class, so an exported `resources.js` subclass's method (and its access control) always wins — the same rule as custom `mcpTools`. Custom resources are served to **any** MCP session, including anonymous, unauthenticated ones (the public-docs case this feature targets) — the MCP layer performs no auth check for them; to restrict one, check `context.user` in the `read` method and throw.
```
heskew: This is not sufficient guidance for protected row-backed content. The page never documents the second `(params, context)` argument, and [Harper #1839](https://github.com/HarperFast/harper/pull/1839) reproduced a data leak when a custom resource fetched through the base `tables.X` class: that bypasses the exported subclass's per-record `allowRead`. Please document the direct `context.user` gate and the current secure row-backed pattern (exported routing Resource + `checkPermission`-bearing `RequestTarget` inside the user-carrying transaction), including its version boundary and the warning not to fetch guarded rows through `tables.X`.

---

🤖 Posted by Codex (GPT-5) on behalf of @heskew

reference/mcp/tools-and-resources.md
```diff
@@ -108,6 +108,49 @@ class Orders extends Tables.orders {
...
+			completions: { path: ['guides/install.md', 'guides/deploy.md'] },
+		},
+	];
+
+	async readIndex() {
+		return { text: '- docs:///guides/install.md\n…', mimeType: 'text/markdown' };
+	}
+
+	async readPage(params) {
+		return { text: loadPage(params.path), mimeType: 'text/markdown' };
```
heskew: Because `{+path}` accepts decoded, multi-segment client input and the adjacent text explicitly allows anonymous reads, `loadPage(params.path)` models a path-traversal footgun: a URI such as `docs:///../../secret` reaches the loader as `../../secret`. Please use an allowlisted map like the core fixture, or show containment/validation before any filesystem-backed load and state that template parameters are client-controlled.

---

🤖 Posted by Codex (GPT-5) on behalf of @heskew

reference/mcp/tools-and-resources.md
heskew: The scheme change is correct here, but three current sibling pages still contradict it: `subscriptions.md` says the `https://`/`http://` form is what `resources/list` advertises (and uses it in every example), `tool-metadata.md` lists `https://{host}/{path}`, and `overview.md` describes only `https://` resource exposure. Please update those canonical references/examples to `harper+rest://` while retaining one explicit note that legacy `http(s)://` inputs still work. This is separate from the existing version-badge thread on this line.

---

🤖 Posted by Codex (GPT-5) on behalf of @heskew

heskew (dismissed):
Requesting changes for three API-contract issues:

- Document the actual custom-resource method signature and the current safe authorization pattern for row-backed content.
- Make the anonymous `{+path}` example validate or allowlist client-controlled paths before loading content.
- Update the sibling MCP overview, metadata, and subscription pages to use the new listed `harper+rest://` scheme.

The v5.1.18 feature/version and compatibility claims otherwise match core, and CI is green. The existing @kriszyp badge/ASCII threads remain separate and should also be resolved.

---

🤖 Posted by Codex (GPT-5) on behalf of @heskew

reference/mcp/tools-and-resources.md
Ethan-Arrowood: Done in b1e7f82a — `<VersionBadge version="v5.1.18" />` on its own line under the heading, inline `(5.1.18+)` removed.

reference/mcp/tools-and-resources.md
Ethan-Arrowood: Fixed in b1e7f82a. Both the sample and the `docs:///...` note bullet are plain ASCII now.

The sample also changed shape while I was in it (see the path-traversal thread): it mirrors the core fixture's allowlist map, so the `readIndex` return that carried the ellipsis is now built from `Object.keys(PAGES)` and doesn't need one.

reference/mcp/tools-and-resources.md
Ethan-Arrowood: Done in b1e7f82a — `<VersionBadge type="changed" version="v5.1.18" />` under the heading, inline marker removed, and the "earlier releases listed these under `http(s)://`" rationale moved into its own sentence rather than a parenthetical.

reference/mcp/tools-and-resources.md
Ethan-Arrowood: Agreed, and fixed in b1e7f82a. Two new subsections under the opt-in:

**The signature.** `(params, context)` is now documented, including what `context` carries (`{ user, profile }`, `user.role.permission` being the RBAC block) and that a fixed-`uri` entry gets `params = {}`.

**Access control for custom resources.** Says outright that the MCP layer runs no authorization check, then covers both cases:

- Author-rendered content: gate on `context.user` and throw. I also called out that an anonymous session arrives as `{ username: '' }`, not `undefined` (`isAuthenticated` in `toolRegistry.ts`), so `if (!context.user)` does not reject it — that seemed like the exact mistake this guidance invites.
- Row-backed content: the secure pattern from harper#1839, worked as a full example — exported routing `Order extends tables.Order` with the per-record `allowRead`, then `RequestTarget` + `target.checkPermission = context.user?.role?.permission ?? true` + single-arg `Order.get(target)`. Plus the "call it single-arg" note, since forwarding `this.getContext()` starts an independent transaction and drops the shared snapshot.
- A `:::warning` for the `tables.X` footgun, linking harper#1735.

On the version boundary: harper#1839's merge commit (`07c2bbcb`) is first tagged **v5.2.0** — no 5.1.x tag contains it, and `v5.1.26:components/mcp/resources.ts` still has the bare `await def.read(params, { user, profile })`. So the split I documented is: the `context.user` gate works from 5.1.18, but the user-carrying transaction (and therefore the guarded row-backed fetch, and `AccessViolation` surfacing as `permission denied`) needs 5.2.0. The warning admonition says that explicitly.

reference/mcp/tools-and-resources.md
Ethan-Arrowood: Good catch — fixed in b1e7f82a.

The example now mirrors the core fixture: a module-level `PAGES` allowlist, `readPage` resolving `PAGES[params.path]` and throwing `no such page` on a miss, and `readIndex` built from `Object.keys(PAGES)`. `loadPage(params.path)` is gone.

I also added a **Template parameters are client-controlled** subsection stating the mechanism rather than just showing the safe shape: `{+name}` matches across `/`, the captured value is percent-decoded before the method sees it, and a read of `docs:///../../secret` calls `readPage({ path: '../../secret' })` — so resolve through an allowlist or validate containment before any filesystem- or URL-backed load.

reference/mcp/tools-and-resources.md
Ethan-Arrowood: All three updated in b1e7f82a, each keeping one explicit legacy note:

- **`subscriptions.md`** — the three `params.uri` examples, the `notifications/resources/updated` example, the "Which URIs are subscribable" sentence, and the record-vs-collection table now use `harper+rest://`. Added a note that `resources/list` advertised these under `http(s)://` before 5.1.18 and that a legacy `http(s)://` URI is still accepted by `resources/subscribe`/`resources/unsubscribe` — which matches `subscribeToResource`, where the scheme check passes `harper+rest:`, `https:`, and `http:`.
- **`tool-metadata.md`** — the resource table row is `harper+rest://{host}/{path}`, the description-prepending sentence follows, and it links to the `harper+rest://` section for the full story.
- **`overview.md`** — the "Resource exposure" bullet now says `harper+rest://`, and mentions that components can publish their own content under author-chosen URIs.

I swept the rest of `learn/`, `reference/`, and `fabric/` for other `https://`-as-MCP-resource references and these three were the only ones.

Ethan-Arrowood:
Picking this up to finish it. All three of @heskew's change requests and all three of @kriszyp's threads are addressed in `b1e7f82a` — details in the individual threads. Summary:

- **Authorization guidance for row-backed content.** The read signature `(params, context)` is now documented, along with a new "Access control for custom resources" subsection: no MCP-layer auth check, the `context.user` gate, and the secure row-backed pattern from [harper#1839](https://github.com/HarperFast/harper/pull/1839) — exported routing Resource plus a `checkPermission`-bearing `RequestTarget` under the user-carrying transaction — with a warning against fetching guarded rows through `tables.X` ([harper#1735](https://github.com/HarperFast/harper/issues/1735)).
- **The `{+path}` footgun.** The example now resolves through an allowlist map like the core fixture, and a new "Template parameters are client-controlled" subsection states the mechanism: `{+name}` matches across `/` and the value is percent-decoded, so `docs:///../../secret` arrives as `readPage({ path: '../../secret' })`.
- **The sibling pages.** `subscriptions.md`, `tool-metadata.md`, and `overview.md` now use `harper+rest://`, each keeping one note that legacy `http(s)://` still works.

Two notes on versions, checked by file content at the tags rather than by `--contains`:

- **The 5.1.18 badges are correct.** `components/mcp/customResourceRegistry.ts` and `HARPER_REST_SCHEME` are both present at `v5.1.18` and absent at `v5.1.17`.
- **The secure row-backed pattern needs 5.2.0, not 5.1.18.** harper#1839's user-carrying transaction is absent through `v5.1.26` (`v5.1.26:components/mcp/resources.ts` still has a bare `await def.read(params, { user, profile })`) and present from `v5.2.0`. So the page splits it: the `context.user` gate works from 5.1.18, the guarded row-backed fetch from 5.2.0. That section carries its own `v5.2.0` badge.

Also in `96a04eb3`: a 5.1.18 release-notes entry, since neither the feature nor the scheme change had one — and the scheme change is observable by any client that re-reads `resources/list`.

Merged `main` (the branch was 127 commits behind). `npm run build` is clean — the two reported broken anchors are pre-existing on main — and `npm run format:check` passes.

One thing for whoever merges: **the PR description still says "Hold merging until harper#1613 lands"**, which is stale. #1613 shipped in 5.1.18; Kyle noted that in a comment above but the description was never updated. There is no hold.

reference/mcp/tools-and-resources.md
```diff
@@ -112,6 +112,144 @@ The MCP transport audits the `tools/call`, but invokes the custom instance metho
...
+An anonymous session arrives as `{ username: '' }`, not `undefined`, so `if (!context.user)` does **not** reject it. Test the field you actually require — a non-empty `context.user.username`, or the specific `context.user.role.permission` entry.
+
+When the content wraps row-level-guarded table data, fetch it through the **exported (routing) Resource** — the subclass that carries the row-level `allow*` predicates — with a `checkPermission`-bearing `RequestTarget` <VersionBadge version="v5.2.0" />:
+
+```ts
+import { RequestTarget } from 'harper';
+
+// The exported subclass carries the per-record guard that REST enforces.
+export class Order extends tables.Order {
+	allowRead(user) {
```
kriszyp: Could we change this to a `static get` to do the auth check?

kriszyp (commented):
Looks good, a couple of fixes.

---
Proposed inline comments (anchors failed):
- `.github/workflows/deploy-fabric.yaml:13`: This production-deploy path filter omits `reference/**` (and the versioned Reference files/config) even though those are the site's primary documentation inputs; the preview workflow includes them at `.github/workflows/pr-preview.yaml:7-10`. A reference-only merge—such as the MCP change this branch started with—will therefore never update the Fabric deployment. Please mirror the Reference path entries here (`reference/**`, `reference_versioned_docs/**`, `reference_versioned_sidebars/**`, and `reference_versions.json`) before enabling this production path.
- `reference/security/secrets.md:234`: This fenced JSON block has spaces before tabs on lines 234-237; `git diff --check b1e7f82a..HEAD` reports each one. Please run the required formatter (or normalize the indentation) so the formatting gate is clean.
- `.gitignore:44`: This new final line has no terminating newline. Please run the required formatter so the whole-repository formatting check remains clean.

reference/mcp/tools-and-resources.md
Ethan-Arrowood: Yes — done in 88b87bd9, and thanks for pushing on it, because the example was teaching a deprecated mechanism.

The exported subclass now gates in `static get`:

```ts
import { getUser } from 'harper';

export class Order extends tables.Order {
	static async get(target, context) {
		const order = await super.get(target, context);
		const user = getUser();
		if (order && order.customerId !== user?.username) throw new Error('not authorized');
		return order;
	}
}
```

and the custom resource just calls `Order.get(params.orderId)`.

Two things fell out of that beyond the mechanical swap:

**The `checkPermission` assignment is gone.** The old example did `target.checkPermission = context.user?.role?.permission ?? true`, which I had lifted from the core fixture in harper#1839. But `resource-api.md`'s `RequestTarget` reference says trusted code may set it to `true` and to *never copy a permission object* into it. With the gate in an override the delegated call needs no `RequestTarget` plumbing at all. I kept one closing sentence for Resources still on the legacy hooks: set `checkPermission = true` only, and never take it from client input — which matters here because a template parameter *is* client input.

**The prose now points at the current model** rather than the row-level-predicate framing: `allow*` are deprecated one-time operation gates, put the check in an operation override, linking [Authorization](https://github.com/HarperFast/documentation/blob/main/release-notes/v5-lincoln/5.2.md#authorization) in the 5.2 notes. `getUser()` is documented as `getContext().user`, and I confirmed it against core — `security/jsLoader.ts` returns `contextStorage.getStore()?.user`, exported from `harper` at v5.2.6.

That last point is also what keeps the v5.2.0 badge honest: harper#1839 is what runs the read inside a `transaction({ ...store, user })`, so `getUser()` only returns the MCP session user from 5.2.0 on. On 5.1.x there is no user in the ambient context for a delegated fetch to authorize against.

The harper#1735 warning is unchanged in substance — the base table class doesn't carry the subclass's override, so `tables.Order.get(id)` still skips the gate.

Ethan-Arrowood:
### On the three proposed comments whose anchors failed

All three are pre-existing on `main`, not changes from this PR — they showed up in the diff range only because I merged `main` into this branch (it was 127 commits behind). Evidence for each, so they don't get lost:

**1. `deploy-fabric.yaml` omits `reference/**` — real bug, and worse than "won't update."** Confirmed on `main`: the `paths` filter lists `docs/**`, `fabric/**`, `learn/**`, `release-notes/**`, `src/**`, `static/**`, `versioned_docs/**`, `versioned_sidebars/**`, `docusaurus.config.ts`, `package-lock.json` — no `reference/**`, and no `reference_versioned_docs/**`, `reference_versioned_sidebars/**`, or `reference_versions.json`. Since `reference/` is the bulk of the site, a reference-only merge silently skips the production fabric deploy. Worth its own PR rather than riding along in an MCP docs change — happy to open one, or to fold it in here if you'd rather it not wait.

**2. `secrets.md:234-237` space-before-tab — pre-existing.** `git show main:reference/security/secrets.md | sed -n '232,238p'` shows the same `   \t"operation": ...` lines. `git diff --check` flags them against `b1e7f82a..HEAD` because the merge commit pulls main's content into that range; they are main's lines. Note the formatting gate does *not* catch these — `npm run format:check` is green on this branch and on main, because Prettier leaves that fenced JSON block inside a numbered list alone. So `git diff --check` is finding something the required formatter genuinely doesn't, which is a gap worth knowing about, but running the formatter won't fix it.

**3. `.gitignore` missing trailing newline — pre-existing.** `git show main:.gitignore | od -c | tail` ends `.claude/` plus trailing spaces with no `\n`. Also not something Prettier touches (`.gitignore` isn't in its file set).

I'd rather not fix main's `.gitignore`, a security doc, and a deploy workflow inside this PR — it would make it harder to review and to revert. Say the word and I'll open a separate cleanup PR for #2 and #3 and a dedicated one for #1, or fold any of them in here if you prefer.

### Also in this push

Rewrote the row-backed authorization example to gate in a `static get` override instead of the deprecated `allowRead` — details in the inline thread.

Separately: I force-pushed to correct the committer email on my three commits (`ethan@harperdb.io` -> `ethan@arrowood.dev`), which is why the SHAs changed. Content is byte-identical to what was there before the rewrite commit; the commits now show as verified.

Ethan-Arrowood:
Both whitespace/CI items from your review are now split out:

- **[#651](https://github.com/HarperFast/documentation/pull/651)** — `ci: deploy on reference/** and historic-redirects.ts changes`. Wider than reported: `deploy.yaml` (Pages, still the canonical deploy) omits the `reference*` paths too, so a reference-only merge deployed to *neither* target. And `historic-redirects.ts` was missing from all three workflows despite `redirects.ts:16` importing it, so editing a historic redirect changed the built site with no deploy and no preview. Purely additive. I left the four stale entries (`docs/**`, `versioned_docs/**`, `versioned_sidebars/**`, `versions.json` — none of which exist any more) in place and flagged them for your call.
- **[#652](https://github.com/HarperFast/documentation/pull/652)** — `chore: add missing final newlines to config files`. Five files, all outside Prettier's reach (`prettier` cannot infer a parser for `.gitignore`/`.npmrc`/`.node-version`/`.git-blame-ignore-revs`/`CODEOWNERS`), so the formatting gate never covered them.

On the `secrets.md` space-before-tab: **there is nothing to fix, and running the formatter does not change it.** That indentation is Prettier's own required output here. Checked in-repo, so the real config applies:

```
$ npx prettier --check reference/security/secrets.md     # as committed, with the tab
All matched files use Prettier code style!

$ npx prettier --check tmp/spaces-variant.md             # same file, tab -> spaces
[warn] Code style issues found in the above file.
```

The fence sits inside a numbered list item, so its lines carry a 3-space list indent, and `@harperfast/code-guidelines` sets `useTabs: true`, so Prettier indents the embedded JSON with a tab. `3 spaces` + `tab` is precisely what git's heuristic flags — and converting it to spaces makes CI fail. Removing the pattern would mean flipping `useTabs` org-wide or not nesting fenced code in lists; neither seems worth it for a heuristic.

Same heuristic has another false positive in the repo: `caching-old.md:41`'s trailing double-space is a markdown hard line break that Prettier preserves deliberately. `git diff --check` and Prettier disagree on markdown, and Prettier is the gate — reasoning documented in #652 so this doesn't get re-litigated.

That leaves this PR with just the `static get` change from your review. @heskew — your `CHANGES_REQUESTED` from Aug 26 is the only thing blocking now; all three items are addressed in the threads above.

kriszyp (commented):
Looks good, maybe one thing to cleanup.

---
Proposed inline comments (anchors failed):
- `.github/workflows/deploy-fabric.yaml:13`: This production-deploy path filter omits `reference/**` (and the versioned Reference files/config) even though those are the site's primary documentation inputs; the preview workflow includes them at `.github/workflows/pr-preview.yaml:7-10`. A reference-only merge—such as the MCP change this branch started with—will therefore never update the Fabric deployment. Please mirror the Reference path entries here (`reference/**`, `reference_versioned_docs/**`, `reference_versioned_sidebars/**`, and `reference_versions.json`) before enabling this production path.
- `reference/security/secrets.md:234`: This fenced JSON block has spaces before tabs on lines 234-237; `git diff --check b1e7f82a..HEAD` reports each one. Please run the required formatter (or normalize the indentation) so the formatting gate is clean.
- `.gitignore:44`: This new final line has no terminating newline. Please run the required formatter so the whole-repository formatting check remains clean.

Ethan-Arrowood:
@kriszyp — the three "anchors failed" items in your latest review are the same ones from Aug 26, re-emitted by the bot against the new commit. They're main's content sitting in this branch's diff range because of the `main` merge, not changes from this PR. All three are now resolved outside this PR, and you approved both:

- **[#651](https://github.com/HarperFast/documentation/pull/651)** — the `reference/**` path filter, plus two things the original report didn't cover: `deploy.yaml` (Pages, the canonical deploy) omits them too, and `historic-redirects.ts` was missing from all three workflows.
- **[#652](https://github.com/HarperFast/documentation/pull/652)** — the missing final newlines, in five files rather than one.

The `secrets.md` space-before-tab is the one item with nothing to fix: that indentation is Prettier's own required output under this repo's config, and converting it to spaces makes `format:check` fail. Demonstration and reasoning are in #652's description so it doesn't get re-raised a third time.

So this PR is down to just the `static get` change you asked for. @heskew — your `CHANGES_REQUESTED` from Aug 26 is the only thing blocking; all three of your items are answered in the threads above.

PR #572, opened 2026-07-07
docs(analytics): document transaction-commit-time metric

kriszyp:
Documents the `transaction-commit-time` metric added by [Harper PR #1688, "feat(analytics): record write-transaction commit latency"](https://github.com/HarperFast/harper/pull/1688), including its RocksDB-only submit-to-settle scope, raw/aggregate fields, aggregation delay, and relationship to the versioned overload guard. It also corrects the duplicated analytics examples and links the metric from storage tuning and the 5.2 release notes.

## For the human reviewer

1. **Sibling sequencing:** This PR remains the canonical home for `transaction-commit-time`. [Documentation PR #573, "docs(analytics): document transaction queue depth metrics"](https://github.com/HarperFast/documentation/pull/573) should merge afterward, rebase, drop its duplicate commit-time row/prose, and fold its release note under this PR's Analytics heading.
2. **Schema correction:** The generic raw/aggregate examples now match Harper source: raw rows have `mean`, compressed `distribution`, and `count`; aggregates have all nine percentiles. This was already true in Harper v5.0.0, so the old v5 docs were stale rather than describing a 5.2 schema change. Source: [`resources/analytics/write.ts` at v5.0.0](https://github.com/HarperFast/harper/blob/v5.0.0/resources/analytics/write.ts#L159-L184).
3. **Verbatim operational strings:** The exact HTTP 503 and server-log text remains quoted so operators can grep for what Harper emits. If those strings should not be treated as a stable interface, paraphrasing is the alternative.
4. **Alerting route:** The page keeps direct `hdb_analytics` querying because it already teaches both system tables. `get_analytics` remains available through the Operations API; a reviewer may prefer linking that operation as the recommended alerting path.

## Verification

- End-to-end route: `npm run build` completed successfully and validated the MDX, anchors, and links. It reported only the pre-existing broken `#deployment-operations` anchor warning in the v5.1 release notes.
- `npm run format:write` followed by `npm run format:check` passed.
- `npm run typecheck` and `npm run lint` passed.
- Core claims were spot-checked against Harper v5.0.0 and v5.2.1: raw distributions and aggregate percentiles, empty `path`/`method`/`type`, both fulfilled and rejected async attempts, current oldest-attempt guard behavior, 503/log strings, and update/publish bypass rules.

## Review coverage

Authored by GPT-5.6 Codex. Full review: Claude Opus 5, Cursor Composer 2.5, and Claude Opus 5 Harper-domain adjudication. Final delta: Claude Opus 5. Gemini via agy returned no output in all rounds; Cursor Composer and domain adjudication were policy-pruned on the final narrow delta. Receipt @ `175121c5`.

<sub>Generated by GPT-5.6 Codex.</sub>

<sub>Review-Coverage: authored=unknown; ran=none; rounds=1 @ 3b6e5b27f5ec</sub>

<sub>Human-Review-Need: 4 @ 3b6e5b27f5ec</sub>


reference/analytics/overview.md
```diff
@@ -169,6 +169,25 @@ Harper automatically tracks the following metrics for all services. Applications
 | `bytes-received`      | node.database       | `replication` | `blob`    | bytes | Bytes received for blob replication                        |
 | `replication-latency` | node.database.table |               | `ingest`  | ms    | Time difference from source commit timestamp to local time |
 
+### Storage Metrics
+
+| `metric`                  | `path` | `method` | `type` | Unit | Description                                                       |
+| ------------------------- | ------ | -------- | ------ | ---- | ----------------------------------------------------------------- |
+| `transaction-commit-time` |        |          |        | ms   | Time a write transaction is outstanding (submit → durable commit) |
```
cb1kenobi: **Medium: description undersells what the metric measures — it is not "submit → durable commit"**

The table description says the metric measures "submit → durable commit," implying it only records successful, durably-committed writes. Per the harper implementation (`resources/DatabaseTransaction.ts`), the recorder is attached with `commitResolution.then(record, record)` — it fires on **both fulfillment and rejection**. The code comment there is explicit: "Recorded on both fulfilment and rejection since a slow-then-failed commit still consumed queue time," and elsewhere calls the interval "submit → settle," not "submit → durable commit."

This matters for the doc's own thesis: the metric is a good leading indicator for the 503 precisely *because* it captures the full outstanding window regardless of outcome (matching `checkOverloaded()`'s clock exactly, including transient-conflict retries that reject and re-issue `commit()`). Calling it "submit → durable commit" could lead a reader to assume failed/retried commits are excluded, undercounting exactly the pathological case (slow, contended, or failing commits) that the metric exists to surface.

Suggested fix:
```suggestion
| `transaction-commit-time` |        |          |        | ms   | Time a write transaction is outstanding (submit → settle, success or failure) |
```

—
_Generated by Barber AI_

reference/analytics/overview.md
```diff
@@ -169,6 +169,25 @@ Harper automatically tracks the following metrics for all services. Applications
...
+
+| `metric`                  | `path` | `method` | `type` | Unit | Description                                          |
+| ------------------------- | ------ | -------- | ------ | ---- | --------------------------------------------------- |
+| `transaction-commit-time` |        |          |        | ms   | Time a write transaction is outstanding (submit → durable commit) |
+
+`transaction-commit-time` is recorded as a distribution (`mean`, `median`, `p90`, `p95`, `p99`,
+`p999`, `count`) per write commit. It measures the same interval the storage engine's overload guard
+uses: when the oldest outstanding commit exceeds `storage.maxTransactionQueueTime` (default 45s),
+Harper rejects new write transactions with `Outstanding write transactions have too long of queue,
+please try again later` (HTTP 503).
```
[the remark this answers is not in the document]
kriszyp: Verified against `harper`'s `resources/DatabaseTransaction.ts:406` — the actual product error message is `'Outstanding write transactions have too long of queue, please try again later'`, typo included. I kept the doc's quoted message verbatim to match what operators will actually see in logs/responses, rather than silently "fixing" it into a message the product doesn't emit. The underlying typo is a real (pre-existing) bug in `harper`, tracked as a finding for a separate fix in that repo rather than this docs PR.

— Claude Opus (dev-agent)

cb1kenobi (approved):

reference/analytics/overview.md
```diff
@@ -169,6 +167,41 @@ Harper automatically tracks the following metrics for all services. Applications
...
+
+`transaction-commit-time` is recorded on the RocksDB asynchronous commit path only; it is not
+emitted for LMDB-backed databases, and not for the synchronous `commitSync()` path used during
+transaction-log replay. Each sample covers one commit attempt, not one logical write transaction — a
+transient-conflict retry re-issues the commit and records its own sample, so `count` can exceed the
+number of logical writes. A sample is only recorded once an attempt settles, so a commit that is
+still outstanding contributes nothing yet. Raw entries (`hdb_raw_analytics`) carry `mean`,
+`distribution`, and `count`; percentiles (`median`, `p90`, `p95`, `p99`, `p999`) are only available on
+the per-minute aggregate (`hdb_analytics`) once raw entries are rolled up — query the aggregate table
+for percentile-based alerting. It shares a timebase with the RocksDB storage engine's overload guard,
```
Ethan-Arrowood, line 186: The percentile list reads as exhaustive, but the aggregator emits nine, not five (`resources/analytics/write.ts:983-984`: `p1, p10, p25, median, p75, p90, p95, p99, p999`). Since this sentence is specifically teaching the raw-vs-aggregate split, an incomplete list undercuts the point it's making.

```suggestion
`distribution`, and `count`; percentiles (`p1`, `p10`, `p25`, `median`, `p75`, `p90`, `p95`, `p99`,
`p999`) are only available on the per-minute aggregate (`hdb_analytics`) once raw entries are rolled
up — query the aggregate table for percentile-based alerting. It shares a timebase with the RocksDB
storage engine's overload guard,
```

reference/analytics/overview.md
```diff
@@ -169,6 +167,41 @@ Harper automatically tracks the following metrics for all services. Applications
...
+which times only the commit attempt it arms: when a tracked outstanding commit on a thread exceeds
+`storage.maxTransactionQueueTime` (default 45s), Harper rejects new record updates and publishes on
```
Ethan-Arrowood, line 188: Two things here. `storage.maxTransactionQueueTime` appears three times in this section as bare inline code with no link; it's documented in `storage-tuning.md`, and #573 already links it in exactly this relative form — worth matching on first mention.

Separately, "times only the commit attempt it arms" carries the same single-slot assumption as the paragraph below (see my other comment) — on `main` the guard times the oldest of all tracked outstanding commits. Neutral phrasing here is true on both.

```suggestion
which times outstanding commit attempts: when a tracked outstanding commit on a thread exceeds
[`storage.maxTransactionQueueTime`](../database/storage-tuning.md#storagemaxtransactionqueuetime)
(default 45s), Harper rejects new record updates and publishes on
```

reference/analytics/overview.md
```diff
@@ -169,6 +167,41 @@ Harper automatically tracks the following metrics for all services. Applications
...
+that thread with `Outstanding write transactions have too long of queue, please try again later`
+(HTTP 503) — deletes and writes applied from a canonical source (e.g. replication or a caching
+source) bypass this check. The guard tracks at most one outstanding commit per thread: a retry
+issued while the prior attempt still holds that slot (a coordinated retry, or an early backoff
+retry) recommits before the slot clears and is never armed, so a wedge there won't trip the 503; a
+later, backoff-delayed retry recommits after the slot clears and is tracked like a fresh attempt.
```
Ethan-Arrowood, line 194: **This documents a bug that `harper` has already fixed.**

The single-slot description is an accurate read of `v5.2.0` (`let outstandingCommit, outstandingCommitStart;` plus `if (!outstandingCommit)` arming). But `319714cf1` ("Track every outstanding write commit, not just one per thread", harper#2009) replaced the slot with a linked list: `origin/main` now holds `oldestOutstandingCommit`/`newestOutstandingCommit`, and the source comment names the single-slot behaviour as the defect — a wedged retry was invisible to `checkOverloaded()` forever. I confirmed `319714cf1` is **not** an ancestor of `v5.2.0`, so it ships in the next minor and this paragraph becomes false then.

I'd delete it rather than version-qualify it. `<VersionBadge>` means *since*, not *only in*, so there's no way to scope it — and it isn't operator-actionable either way, since nobody can observe or act on per-thread arming slots. Everything actionable in this section survives the cut.

```suggestion
source) bypass this check.
```

reference/analytics/overview.md
```diff
@@ -169,6 +167,41 @@ Harper automatically tracks the following metrics for all services. Applications
...
+
+A rising `p99`/`p999` signals commits are taking longer to drain — from write volume, large
+transactions, or a saturated storage volume — and is a useful early warning to shed or throttle write
+load. But the metric shares only a timebase with the guard, not its population: a wedged commit
+contributes no sample until it settles, and an early retry that wedges can go untracked by the guard
+entirely (see above). Don't treat this distribution as a leading indicator on its own — watch the
+server log for the "Rejecting writes on this thread" error the guard emits when it does trip, and
+don't rely solely on percentiles trending toward `storage.maxTransactionQueueTime`. Tune the
+threshold against a baseline for your workload.
```
Ethan-Arrowood, line 203: Follow-on from the retry-slot paragraph above: this clause depends on it and is false on `main` for the same reason. Reflowed to drop it. The rest of the guidance is unchanged and holds on both `v5.2.0` and `main`, including the `Rejecting writes on this thread` reference, which I verified exists.

```suggestion
load. But the metric shares only a timebase with the guard, not its population: a wedged commit
contributes no sample until it settles. Don't treat this distribution as a leading indicator on its
own — watch the server log for the "Rejecting writes on this thread" error the guard emits when it
does trip, and don't rely solely on percentiles trending toward `storage.maxTransactionQueueTime`.
Tune the threshold against a baseline for your workload.
```

Ethan-Arrowood (changes requested):
Reviewed against the `harper` source at `v5.2.0` (`c11e0976`) and `origin/main`. The metric name, emission path, units, the raw/aggregate field split, the 503 error string, and the bypass rules all check out — I verified each rather than assuming. Four things to fix, plus one sequencing decision shared with #573.

**Sequencing with #573 — read this first.** #573 documents the same `transaction-commit-time` metric in the same file, in the *Resource Usage Metrics* table, with a contradictory attribute list (`mean`, `median`, `p90`, `p95`, `p99`, `p999` — i.e. exactly the raw/aggregate confusion this PR correctly fixes). I ran the trial merge: git merges the two **cleanly, with zero conflicts**, so nothing warns you. The merged page documents the metric in two different tables with contradictory guidance, and `5.2.md` ends up with two `## Analytics` H2s.

**Merge order: this PR (#572) first, then #573.** This PR should own `transaction-commit-time` — it has the purpose-built `### Storage Metrics` section, its table shape (`path`/`method`/`type`) matches how the metric is actually emitted via `recordAction()` with no `byThread` flag, and it fixes the stale raw-entry example. #573 then rebases onto this, drops its `transaction-commit-time` row and duplicated prose, cross-links `#storage-metrics`, and folds its release note under the `## Analytics` heading this PR creates. I've left the matching comments on #573.

**One thing that is _not_ a problem, so it doesn't get re-litigated:** the `Rejecting writes on this thread` log string at line 201 is real. I grepped it at both `v5.2.0` (`DatabaseTransaction.ts:394`) and `origin/main` (`:485`), and the source comment there confirms it fires once per stuck commit. Keep it.

**Aggregate example left stale while the raw one was fixed** (can't suggest inline — unchanged, so not in the diff). You correctly rewrote the raw example, but the aggregate example just below it (`reference/analytics/overview.md` lines ~107-120) still shows only `median`/`mean`/`p95`/`p90`, omitting the `p99`/`p999` this PR's new prose tells operators to alert on. A reader who follows the guidance and looks at the example won't find the fields they were just told to query. Same drive-by class as the fix already in here.

Two smaller ones:

- The alerting advice doesn't mention that the aggregate has a 60s floor (`analytics.aggregatePeriod`, default 60), which bounds how fast a percentile alert can fire. One clause would make the guidance actionable.
- `storage-tuning.md`'s `storage.maxTransactionQueueTime` section says "Harper rejects new writes with HTTP 503" and now has an observable signal to point at — worth adding the reverse link to `#storage-metrics`.

reference/analytics/overview.md
cb1kenobi: **High: the "one outstanding commit per thread" guard description is stale — harper tracks every attempt since v5.2.1**

This paragraph describes the pre-v5.2.1 guard: a single module-level `outstandingCommit` slot that only timed the attempt it happened to arm. That was replaced in harper by [`Track every outstanding write commit, not just one per thread` (319714cf1)](https://github.com/HarperFast/harper/commit/319714cf1), which is on `main` and shipped in `v5.2.1`/`v5.2.2`. In `resources/DatabaseTransaction.ts`, `trackOutstandingCommit()` now links **every** submitted attempt into a list and `checkOverloaded()` rejects on the **oldest** of them. The source comment is explicit that the behavior documented here was the bug, not the design:

> Every attempt is tracked unconditionally: a coordinated retry round and a chained second-store commit are both issued from inside the preceding commit's own resolve handler ... so anything conditional on "is something already outstanding" skips them and leaves a wedged retry or chain invisible to checkOverloaded() forever.

The operational advice therefore inverts: a coordinated or early-backoff retry that wedges *does* now trip the 503, on its own timer. (The `v5.2.0` badge is correct for the metric itself — `transaction-commit-time` landed in 4209c7539 and is present in the `v5.2.0` tag — but the guard semantics changed a release later.)

Two other spots carry the same stale claim and need the same correction: line 187 ("which times only the commit attempt it arms") and lines 199-200 ("an early retry that wedges can go untracked by the guard entirely").

Suggested fix:
```suggestion
source) bypass this check. The guard tracks every outstanding commit attempt on the thread and
rejects once the oldest of them exceeds the limit, so a conflict retry or a chained commit that
wedges is caught on its own timer rather than hiding behind an earlier attempt.
```

—
_Generated by Barber AI_

reference/analytics/overview.md
```diff
@@ -169,6 +167,41 @@ Harper automatically tracks the following metrics for all services. Applications
 | `bytes-received`      | node.database       | `replication` | `blob`    | bytes | Bytes received for blob replication                        |
 | `replication-latency` | node.database.table |               | `ingest`  | ms    | Time difference from source commit timestamp to local time |
 
+### Storage Metrics
```
cb1kenobi, line 184: **Medium: this section documents `transaction-commit-time` a second time, and contradicts its sibling PR**

[HarperFast/documentation#573 — docs(analytics): document transaction queue depth metrics](https://github.com/HarperFast/documentation/pull/573) also documents `transaction-commit-time` in this same file: it adds a row to the existing **Resource Usage Metrics** table plus a `#### Transaction Queue Depth Metrics` subsection describing the same submit-to-settle duration. The two hunks don't overlap, so git will merge them cleanly and the page ends up describing one metric in two sections, with two attribute tables and two anchors.

The two PRs also disagree on the headline guidance:

- #573: "a rising `p99`/`p999` ... **is a leading indicator** of _gradual_ slowdowns approaching that limit"
- here, line 200: "**Don't treat this distribution as a leading indicator** on its own"

Related: lines 200-201 point operators at the server log as the only other signal, but `write-transaction-queue-depth` (in `resources/analytics/metadata.ts`, present since the `v5.2.0` tag) is the metric whose `depth` stays elevated for as long as a commit is wedged — and #573 documents exactly that. Worth pointing here instead of log-only.

Pick one home for the metric across the two PRs, and one framing of the leading-indicator caveat, before either merges.

—
_Generated by Barber AI_

release-notes/v5-lincoln/5.2.md
```diff
@@ -43,3 +43,9 @@ Components can now declare `host` and `urlPath` in `config.yaml`, or pass them t
 ### Web Application Firewall
 
 Harper Pro now includes a Web Application Firewall that evaluates rule-based IP/CIDR, method, path, header, and query conditions before authentication and application routing. Rules support block, log, and score actions; cluster-wide monitor and off modes; per-rule shadowing; node activation gates; live replicated updates; and RE2-backed regular expressions. See [Web Application Firewall](/reference/v5/web-application-firewall/overview).
+
+## Analytics
```
cb1kenobi: **Low: duplicate `## Analytics` release-note section — the sibling PR adds one too**

[HarperFast/documentation#573 — docs(analytics): document transaction queue depth metrics](https://github.com/HarperFast/documentation/pull/573) inserts its own `## Analytics` H2 (with a `### Transaction queue depth metrics` entry that also announces `transaction-commit-time`) after the `set_configuration` entry, above `## CLI`. This entry is appended at the end of the file instead, so the two merge cleanly into a page carrying two `## Analytics` headings that both announce the same metric and link to different anchors.

Fold this entry into the same `## Analytics` section #573 adds.

—
_Generated by Barber AI_

reference/analytics/overview.md
kriszyp: Addressed across `2bd3f8d8` and `3ac1b4bf`: the prose and both aggregate examples now list all nine emitted percentiles (`p1`, `p10`, `p25`, `median`, `p75`, `p90`, `p95`, `p99`, and `p999`).

— GPT-5.6 Codex

reference/analytics/overview.md
kriszyp: Addressed across `2bd3f8d8` and `175121c5`: the first setting mention now links to Storage Tuning, the guard is described in terms of the oldest tracked attempt, and the v5.2.1 tracking semantics are explicitly badged.

— GPT-5.6 Codex

reference/analytics/overview.md
kriszyp: Addressed in `2bd3f8d8` and refined in `175121c5`: the stale single-slot behavior is removed, and the docs now describe the v5.2.1 oldest-wins tracking behavior with a changed-version badge.

— GPT-5.6 Codex

reference/analytics/overview.md
kriszyp: Addressed in `2bd3f8d8`: the false early-retry-untracked clause is removed. The guidance now distinguishes the distribution of settled samples from the authoritative guard log.

— GPT-5.6 Codex

reference/analytics/overview.md
kriszyp: Addressed in `175121c5`: both the analytics page and canonical storage-tuning page badge v5.2.1, state that every attempt is tracked, and clarify that the oldest tracked attempt triggers the guard. The v5.2.0 limitation is explicit.

— GPT-5.6 Codex

reference/analytics/overview.md
kriszyp: Resolved by the agreed sequencing: [PR #572, "docs(analytics): document transaction-commit-time metric"](https://github.com/HarperFast/documentation/pull/572) remains the canonical home. [PR #573, "docs(analytics): document transaction queue depth metrics"](https://github.com/HarperFast/documentation/pull/573) will rebase after this PR, drop its duplicate commit-time row and prose, and cross-link here. This PR no longer attempts to document queue-depth details.

— GPT-5.6 Codex

release-notes/v5-lincoln/5.2.md
kriszyp: Resolved by the same merge order: [PR #572, "docs(analytics): document transaction-commit-time metric"](https://github.com/HarperFast/documentation/pull/572) owns the Analytics heading and release note. [PR #573, "docs(analytics): document transaction queue depth metrics"](https://github.com/HarperFast/documentation/pull/573) will rebase and fold its queue-depth entry under that heading instead of adding another one.

— GPT-5.6 Codex

reference/database/storage-tuning.md
```diff
@@ -33,11 +33,17 @@ storage:
...
+<VersionBadge type="changed" version="v5.2.1" />
+
 Type: `string` (duration)
 
 Default: `45s`
 
-The maximum estimated time a write may wait in the commit queue before Harper rejects new writes with HTTP 503. Acts as backpressure when downstream disk I/O cannot keep up with incoming writes.
+The maximum time a tracked write commit may remain unsettled before Harper rejects new application-originated record updates and publishes on that thread with HTTP 503. This is a per-commit duration check, not a queue-length threshold. Deletes and writes applied from a canonical source (e.g. replication or a caching source) bypass this check.
+
+Beginning with v5.2.1, Harper tracks every outstanding commit attempt for this check. In v5.2.0, a retry or chained commit submitted while another attempt was already tracked could be omitted from overload detection.
```
cb1kenobi: **Low: the v5.2.0 gap was broader than retries and chained commits**

In v5.2.0 the guard armed a single module-level slot — `if (!outstandingCommit) { outstandingCommit = commitResolution; outstandingCommitStart = performance.now(); ... }` in `resources/DatabaseTransaction.ts` — claimed by whichever commit found it free. So *any* commit submitted while another was still outstanding went untracked, not just a retry or a chained commit. That is the case [`Track every outstanding write commit, not just one per thread` (319714cf1)](https://github.com/HarperFast/harper/commit/319714cf12c68066d8fdce724ae5af4821f65bef) names first: "every other commit in flight at that moment was invisible: if one of those wedged, no 503 was ever raised and the write queue grew unbounded with nothing logged."

As written, an operator still on 5.2.0 who uses neither conflict retries nor multi-database writes could reasonably conclude they are not exposed.

Suggested fix:
```suggestion
Beginning with v5.2.1, Harper tracks every outstanding commit attempt for this check. In v5.2.0 only one commit per thread was tracked at a time, so any commit — including a conflict retry or a chained commit — submitted while another was already outstanding could be omitted from overload detection.
```

—
_Generated by Barber AI_

kriszyp: Addressed in `94bf4e4c`: the v5.2.0 note now states that only one commit per thread was tracked and that any commit submitted while it was occupied could be omitted, with retries and chained commits retained as examples.\n\n— GPT-5.6 Codex

reference/analytics/overview.md
```diff
@@ -169,6 +173,47 @@ Harper automatically tracks the following metrics for all services. Applications
...
+conflict retries and chained commits. It rejects once the oldest tracked attempt exceeds the limit,
+so later attempts can no longer be omitted from overload detection.
+
+A rising `p99`/`p999` signals commits are taking longer to drain — from write volume, large
+transactions, or a saturated storage volume — and provides an early warning to shed or throttle
+write load. However, a wedged commit contributes no sample until it settles. Use this distribution
+with the server log: the "Rejecting writes on this thread" error is the authoritative signal when the
+guard trips. Tune the threshold against a baseline for your workload.
+
 ### Resource Usage Metrics
```
cb1kenobi, line 227: **Medium: `transaction-commit-time` is documented in two tables with inconsistent field lists**

This PR adds a correct, detailed "### Storage Metrics" table/section for `transaction-commit-time` right above this heading. But the table below (`### Resource Usage Metrics`, inherited from main) still carries its own `transaction-commit-time` row with `Key attributes: mean, median, p90, p95, p99, p999` (line 227). That list is stale against the actual code (`resources/analytics/write.ts:1036-1037`): the aggregate table (`hdb_analytics`) actually exposes nine percentiles — `p1, p10, p25, median, p75, p90, p95, p99, p999` — and it conflates those aggregate-only fields with the raw-only fields (`mean`, `distribution`, `count`, per `write.ts:194-200`), which the new Storage Metrics section correctly keeps separate.

A reader who only sees the Resource Usage Metrics table gets an incomplete/wrong field list and no indication that `transaction-commit-time` is now fully documented elsewhere in the same page.

Suggested fix: drop the `transaction-commit-time` row from the Resource Usage Metrics table (line 227) now that Storage Metrics is its canonical home, or correct its Key attributes to match code and cross-reference the Storage Metrics section.

—
_Generated by Barber AI_

kriszyp: Fixed in ccc9d178 (and c249733c): the duplicate row is dropped from Resource Usage Metrics, `### Storage Metrics` is the metric's single field reference, and the queue-depth prose now links to `#storage-metrics` instead of restating where percentiles live. The 5.2 release note points its commit-time sentence at the same anchor.

PR #573, opened 2026-07-07
docs(analytics): document transaction queue depth metrics

kriszyp:
## Summary

Documents the v5.2 `write-transaction-queue-depth` and `read-transaction-queue-depth` metrics, including their per-thread sampling and aggregation caveats, and corrects the `storage.maxTransactionQueueTime` description.

This PR must follow [#572, "docs(analytics): document transaction-commit-time metric"](https://github.com/HarperFast/documentation/pull/572). After #572 lands, the remaining review threads require removing this branch's duplicate `transaction-commit-time` row and prose, cross-linking the Storage Metrics section, and folding the queue-depth release note into #572's Analytics section.

## For the human reviewer

1. The current fix commits make no new product-semantics choice. The remaining work is deliberately blocked on #572's merge order; stacking its unmerged branch here would make this PR temporarily own both changes.
2. The full independent pass also surfaced existing judgment calls around reference-page alerting guidance, documenting LMDB's zero-valued gauges versus fixing that core behavior, whether the `storage.maxTransactionQueueTime` correction needs a changed-version badge, and whether that correction should remain bundled here. These remain for human review after the coordination blocker clears.

## Verification

- `npm run build` - passed; it reports the pre-existing broken `#deployment-operations` anchor in the v5.1 release notes as a warning.
- `npm run format:check` - passed.
- `npm run typecheck` - passed.
- `npm run lint` - passed (the repository script is a no-op).

## Review coverage

Original change authored by Claude; current review fixes authored by GPT-5 Codex. Full review: Claude Opus 5 and Harper-domain adjudication succeeded; Gemini produced no output. Final delta review: Claude Opus 5 succeeded, Gemini via agy returned no findings, and the domain pass was pruned as a narrow low-risk delta. Cursor legs were disabled.

Implements [HarperFast/harper#592, "Metrics on in-flight transactions"](https://github.com/HarperFast/harper/issues/592) through the shipped [HarperFast/harper#1689, "feat(analytics): expose write/read transaction queue depth (count) metrics"](https://github.com/HarperFast/harper/pull/1689).

Updated by GPT-5 Codex.

<sub>Human-Review-Need: 4 @ a7a40707af72</sub>

### DIFF
--- a/reference/analytics/overview.md
+++ b/reference/analytics/overview.md
@@ -175,10 +175,40 @@ Harper automatically tracks the following metrics for all services. Applications
 | ------------------------- | ------------------------------------------------------------------------------------------------ | ------------------- | ------- | --------------------------------------------------------------------------------- |
 | `database-size`           | `size`, `used`, `free`, `audit`                                                                  | `database`          | bytes   | Database file size breakdown                                                      |
 | `main-thread-utilization` | `idle`, `active`, `taskQueueLatency`, `rss`, `heapTotal`, `heapUsed`, `external`, `arrayBuffers` | `time`              | various | Main thread resource usage: idle/active time, queue latency, and memory breakdown |
+| `read-transaction-queue-depth`  | `depth`, `maxDepth`                                                                        |                     | count   | Open read (snapshot) transactions (see [transaction queue depth](#transaction-queue-depth-metrics)) |
 | `resource-usage`          | (see below)                                                                                      |                     | various | Node.js process resource usage (see [resource-usage](#resource-usage-metric))     |
 | `storage-volume`          | `available`, `free`, `size`                                                                      | `database`          | bytes   | Storage volume size breakdown                                                     |
 | `table-size`              | `size`                                                                                           | `database`, `table` | bytes   | Table file size                                                                   |
 | `utilization`             |                                                                                                  |                     | %       | Percentage of time the worker thread was processing requests                      |
+| `write-transaction-queue-depth` | `depth`, `maxDepth`                                                                        |                     | count   | In-flight write-transaction commits (see [transaction queue depth](#transaction-queue-depth-metrics)) |
+
+#### Transaction Queue Depth Metrics
+
+`write-transaction-queue-depth` and `read-transaction-queue-depth` expose how many transactions are
+in flight against the storage engine, giving operators a leading indicator before a write-heavy
+workload hits the `Outstanding write transactions have too long of queue, please try again later`
+(HTTP 503) rejection.
+
+| Field      | Unit  | Description                                                                                         |
+| ---------- | ----- | -------------------------------------------------------------------------------------------------- |
+| `depth`    | count | Instantaneous depth sampled at emit time                                                            |
+| `maxDepth` | count | High-water mark observed over the sampling period                                                   |
+
+- **`write-transaction-queue-depth`** counts write commits handed to the storage engine but not yet
+  durably committed — the backlog that produces the overload error when it drains too slowly.
+- **`read-transaction-queue-depth`** counts open read (snapshot) transactions. Persistently high read
+  depth indicates long-lived read snapshots, which can hold back compaction.
+
+Both are gauges sampled per worker thread and summed across threads in the aggregate table (the raw
+per-thread entries in `hdb_raw_analytics` retain each thread's own depth). Note that the aggregate
+`maxDepth` is the sum of each thread's peak, which can read higher than any true simultaneous global
+peak since per-thread spikes need not coincide — treat it as an upper bound, not an exact concurrent
+queue length. Because the write queue can fill and drain within a single sampling period, always
+alert on `maxDepth` (the per-period peak) rather than `depth` alone — an instantaneous sample will
+routinely read low even while short spikes are occurring. A healthy system keeps `write-transaction-queue-depth.maxDepth`
+near zero; a sustained non-zero peak that trends upward is the signal to shed or throttle write load
+before commits start timing out. Tune the concrete alert threshold against a baseline for your
+workload, since absolute depth scales with worker-thread count and per-transaction size.
 
 #### `resource-usage` Metric
 
--- a/reference/analytics/overview.md
+++ b/reference/analytics/overview.md
@@ -171,16 +171,16 @@ Harper automatically tracks the following metrics for all services. Applications
 
 ### Resource Usage Metrics
 
-| `metric`                  | Key attributes                                                                                   | Other               | Unit    | Description                                                                       |
-| ------------------------- | ------------------------------------------------------------------------------------------------ | ------------------- | ------- | --------------------------------------------------------------------------------- |
-| `database-size`           | `size`, `used`, `free`, `audit`                                                                  | `database`          | bytes   | Database file size breakdown                                                      |
-| `main-thread-utilization` | `idle`, `active`, `taskQueueLatency`, `rss`, `heapTotal`, `heapUsed`, `external`, `arrayBuffers` | `time`              | various | Main thread resource usage: idle/active time, queue latency, and memory breakdown |
-| `read-transaction-queue-depth`  | `depth`, `maxDepth`                                                                        |                     | count   | Open read (snapshot) transactions (see [transaction queue depth](#transaction-queue-depth-metrics)) |
-| `resource-usage`          | (see below)                                                                                      |                     | various | Node.js process resource usage (see [resource-usage](#resource-usage-metric))     |
-| `storage-volume`          | `available`, `free`, `size`                                                                      | `database`          | bytes   | Storage volume size breakdown                                                     |
-| `table-size`              | `size`                                                                                           | `database`, `table` | bytes   | Table file size                                                                   |
-| `utilization`             |                                                                                                  |                     | %       | Percentage of time the worker thread was processing requests                      |
-| `write-transaction-queue-depth` | `depth`, `maxDepth`                                                                        |                     | count   | In-flight write-transaction commits (see [transaction queue depth](#transaction-queue-depth-metrics)) |
+| `metric`                        | Key attributes                                                                                   | Other               | Unit    | Description                                                                                           |
+| ------------------------------- | ------------------------------------------------------------------------------------------------ | ------------------- | ------- | ----------------------------------------------------------------------------------------------------- |
+| `database-size`                 | `size`, `used`, `free`, `audit`                                                                  | `database`          | bytes   | Database file size breakdown                                                                          |
+| `main-thread-utilization`       | `idle`, `active`, `taskQueueLatency`, `rss`, `heapTotal`, `heapUsed`, `external`, `arrayBuffers` | `time`              | various | Main thread resource usage: idle/active time, queue latency, and memory breakdown                     |
+| `read-transaction-queue-depth`  | `depth`, `maxDepth`                                                                              |                     | count   | Open read (snapshot) transactions (see [transaction queue depth](#transaction-queue-depth-metrics))   |
+| `resource-usage`                | (see below)                                                                                      |                     | various | Node.js process resource usage (see [resource-usage](#resource-usage-metric))                         |
+| `storage-volume`                | `available`, `free`, `size`                                                                      | `database`          | bytes   | Storage volume size breakdown                                                                         |
+| `table-size`                    | `size`                                                                                           | `database`, `table` | bytes   | Table file size                                                                                       |
+| `utilization`                   |                                                                                                  |                     | %       | Percentage of time the worker thread was processing requests                                          |
+| `write-transaction-queue-depth` | `depth`, `maxDepth`                                                                              |                     | count   | In-flight write-transaction commits (see [transaction queue depth](#transaction-queue-depth-metrics)) |
 
 #### Transaction Queue Depth Metrics
 
@@ -189,10 +189,10 @@ in flight against the storage engine, giving operators a leading indicator befor
 workload hits the `Outstanding write transactions have too long of queue, please try again later`
 (HTTP 503) rejection.
 
-| Field      | Unit  | Description                                                                                         |
-| ---------- | ----- | -------------------------------------------------------------------------------------------------- |
-| `depth`    | count | Instantaneous depth sampled at emit time                                                            |
-| `maxDepth` | count | High-water mark observed over the sampling period                                                   |
+| Field      | Unit  | Description                                       |
+| ---------- | ----- | ------------------------------------------------- |
+| `depth`    | count | Instantaneous depth sampled at emit time          |
+| `maxDepth` | count | High-water mark observed over the sampling period |
 
 - **`write-transaction-queue-depth`** counts write commits handed to the storage engine but not yet
   durably committed — the backlog that produces the overload error when it drains too slowly.
--- a/reference/analytics/overview.md
+++ b/reference/analytics/overview.md
@@ -171,29 +171,33 @@ Harper automatically tracks the following metrics for all services. Applications
 
 ### Resource Usage Metrics
 
-| `metric`                        | Key attributes                                                                                   | Other               | Unit    | Description                                                                                           |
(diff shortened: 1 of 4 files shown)


reference/analytics/overview.md
```diff
@@ -171,14 +171,44 @@ Harper automatically tracks the following metrics for all services. Applications
...
+| `table-size`                    | `size`                                                                                           | `database`, `table` | bytes   | Table file size                                                                                       |
+| `utilization`                   |                                                                                                  |                     | %       | Percentage of time the worker thread was processing requests                                          |
+| `write-transaction-queue-depth` | `depth`, `maxDepth`                                                                              |                     | count   | In-flight write-transaction commits (see [transaction queue depth](#transaction-queue-depth-metrics)) |
+
+#### Transaction Queue Depth Metrics
+
+`write-transaction-queue-depth` and `read-transaction-queue-depth` expose how many transactions are
+in flight against the storage engine, giving operators a leading indicator before a write-heavy
```
cb1kenobi: **High: overclaims this metric's predictive value for the write-queue 503 rejection**

The companion implementation PR ([harper#1689](https://github.com/HarperFast/harper/pull/1689) "feat(analytics): expose write/read transaction queue depth (count) metrics") says outright that this per-thread *count* "doesn't capture transaction size/duration... so count can tell you which thread is submitting more, but not how close the system is to the 503," and names a different metric (`transaction-commit-time`, [harper#1688](https://github.com/HarperFast/harper/pull/1688) "feat(analytics): record write-transaction commit latency") as "the primary leading indicator" for that rejection.

That matches the actual code: the 503 (`checkOverloaded()` in `resources/DatabaseTransaction.ts`) fires when a *single* outstanding commit stays unresolved past `STORAGE_MAXTRANSACTIONQUEUETIME` (45s default) — a duration check on one commit, not a function of `write-transaction-queue-depth`'s concurrent-commit count. A thread can carry `maxDepth` of 50 with fast commits and never hit the 503, or sit at `depth`=1 for 45s and trip it while this metric barely registers as elevated.

The "leading indicator" framing here, and the alerting guidance further down ("a sustained non-zero peak that trends upward is the signal to shed or throttle write load before commits start timing out"), overstate what this metric can tell an operator about that specific failure mode.

Suggested fix: drop the 503-prediction framing for this metric; describe `write-transaction-queue-depth` as a concurrency/throughput signal (how many commits a thread is juggling), not an overload predictor, and point to `transaction-commit-time` as the leading indicator for the 503 once it's documented.

—
_Generated by Barber AI_

reference/analytics/overview.md
```diff
@@ -171,14 +171,44 @@ Harper automatically tracks the following metrics for all services. Applications
...
+
+| Field      | Unit  | Description                                       |
+| ---------- | ----- | ------------------------------------------------- |
+| `depth`    | count | Instantaneous depth sampled at emit time          |
+| `maxDepth` | count | High-water mark observed over the sampling period |
+
+- **`write-transaction-queue-depth`** counts write commits handed to the storage engine but not yet
+  durably committed — the backlog that produces the overload error when it drains too slowly.
+- **`read-transaction-queue-depth`** counts open read (snapshot) transactions. Persistently high read
+  depth indicates long-lived read snapshots, which can hold back compaction.
```
cb1kenobi: **High: missing caveat that these metrics are RocksDB-only — always read zero under `HARPER_STORAGE_ENGINE=lmdb`**

In `resources/DatabaseTransaction.ts`, the comment above `writeTxnQueueDepth` states the accounting backing both metrics is "RocksDB-write-path only": `resources/LMDBTransaction.ts` maintains its own separate, unrelated `trackedTxns` set and never calls `enterWriteQueue`/`leaveWriteQueue` or touches the `trackedTxns` set that `getTransactionQueueDepths()` reads. Since `HARPER_STORAGE_ENGINE=lmdb` is still a supported, selectable engine, an LMDB-backed deployment will see `depth`/`maxDepth` on both `write-transaction-queue-depth` and `read-transaction-queue-depth` pinned at 0 at all times, regardless of actual write/read load — indistinguishable from "healthy, no backlog."

Suggested fix: add a line to this section noting both metrics are RocksDB-only and will always read zero on LMDB-backed installations.

—
_Generated by Barber AI_

cb1kenobi (approved):

reference/analytics/overview.md
```diff
@@ -171,14 +171,71 @@ Harper automatically tracks the following metrics for all services. Applications
...
+| `resource-usage`                | (see below)                                                                                      |                     | various | Node.js process resource usage (see [resource-usage](#resource-usage-metric))                                         |
+| `storage-volume`                | `available`, `free`, `size`                                                                      | `database`          | bytes   | Storage volume size breakdown                                                                                         |
+| `table-size`                    | `size`                                                                                           | `database`, `table` | bytes   | Table file size                                                                                                       |
+| `transaction-commit-time`       | `mean`, `median`, `p90`, `p95`, `p99`, `p999`                                                    |                     | ms      | Duration from write commit submission to settlement (see [transaction queue depth](#transaction-queue-depth-metrics)) |
```
Ethan-Arrowood, line 182: `transaction-commit-time` belongs in #572's `### Storage Metrics` section, not in this table — it isn't a resource-usage gauge. It flows through `recordAction()` with no `byThread` flag, so it aggregates as a cross-thread distribution, unlike every other row here. Dropping the row also resolves the incomplete and self-contradictory attribute list.

```suggestion
```

reference/analytics/overview.md
```diff
@@ -171,14 +171,71 @@ Harper automatically tracks the following metrics for all services. Applications
...
+duration limit (default 45s) that actually trips the 503.
+
+`transaction-commit-time` records each commit's submit-to-settle duration on that same clock, and a
+rising `p99`/`p999` (in the `hdb_analytics` aggregate table, where percentiles are computed — they
+aren't present on `hdb_raw_analytics`) is a leading indicator of _gradual_ slowdowns approaching that
+limit. It doesn't help with a single commit that hangs indefinitely, though: the metric only records
+once a commit settles, so a genuinely wedged commit contributes no sample at all, while
+`write-transaction-queue-depth`'s `depth` stays elevated on that thread for as long as the commit
+remains outstanding. Harper also logs once per stuck commit when the 503 check itself fires, which is
+the authoritative signal for that specific failure.
```
Ethan-Arrowood, line 205: With #572 owning the metric, this paragraph should shrink to the part that is genuinely specific to queue depth: the contrast between a settled-only distribution and a gauge that stays elevated while a commit is wedged. That contrast is the most useful thing in this section and it is **not** in #572, so keep it — just drop the raw/aggregate percentile aside, which #572 covers more completely (and which correctly lists all nine percentiles).

The log sentence stays; I verified it against the source.

```suggestion
[`transaction-commit-time`](#storage-metrics) records each commit's submit-to-settle duration on that
same clock, and a rising `p99`/`p999` is a leading indicator of _gradual_ slowdowns approaching that
limit. It doesn't help with a single commit that hangs indefinitely, though: the metric only records
once a commit settles, so a genuinely wedged commit contributes no sample at all, while
`write-transaction-queue-depth`'s `depth` stays elevated on that thread for as long as the commit
remains outstanding. Harper also logs once per stuck commit when the 503 check itself fires, which is
the authoritative signal for that specific failure.
```

reference/analytics/overview.md
```diff
@@ -171,14 +171,71 @@ Harper automatically tracks the following metrics for all services. Applications
...
+  guarantee the write has been synced to disk.
+- **`read-transaction-queue-depth`** counts concurrently open tracked read transactions, including
+  ones opened with snapshot disabled. A high count can mean either many short-lived reads or a few
+  long-lived ones — the count alone can't distinguish them, so use it as a concurrency signal; a
+  duration-based metric would be needed to identify a single transaction held open long enough to
+  hold back compaction.
+
+Both metrics are gauges tracked only on the RocksDB write/read path, sampled per worker thread. On an
+LMDB-backed database (`storage.engine: lmdb`), `depth` and `maxDepth` for both metrics always read `0`
+— indistinguishable from a healthy, empty queue — regardless of actual read/write load. All per-thread
```
Ethan-Arrowood, line 224: "On an LMDB-backed database ... always read `0`" is imprecise, and the imprecision hides a worse failure mode. These counters are per-thread and process-global — no `database`/`table` attribute, which matches the empty "Other" column in the table. Mixed installs are normal, since `storage.engine` only selects the engine for *new* databases while existing ones keep theirs. On a mixed install the gauges don't read `0`; they report RocksDB traffic only and silently under-count, which is harder to notice than a flat zero.

```suggestion
Both metrics are gauges tracked only on the RocksDB write/read path, sampled per worker thread. They
carry no `database` dimension, so activity against LMDB-backed databases is not counted at all: on a
mixed install they report RocksDB traffic only and silently under-count, and on an install with no
RocksDB databases (`storage.engine: lmdb`) both always read `0` — indistinguishable from a healthy,
empty queue — regardless of actual read/write load. All per-thread
```

Ethan-Arrowood (changes requested):
Re-reviewed at `91d9fecd`. I verified the substantive claims against the merged implementation in `harper`, and the hard ones are right: the "sum of per-thread averages" aggregation semantics (`write.ts:942-957`), the "no row at all on a quiet thread" caveat (the analytics timer is only armed from inside `recordAction()`, so a thread with zero recorded actions never reaches the gauge push), and the `storage.maxTransactionQueueTime` rewrite (both `checkOverloaded()` call sites are gated on `!context?.source`; deletes never reach it). Those are the valuable parts of this PR and I'm not asking you to touch them.

The blocker is coordination with #572, not correctness.

**Blocker: semantic collision with #572 — and git merges the two cleanly, so nothing warns you.** Both PRs document `transaction-commit-time` in `reference/analytics/overview.md`. #572 puts it in a new `### Storage Metrics` section; this PR adds it to `### Resource Usage Metrics`. I ran the trial merge: **"Automatic merge went well", zero conflicts**. The merged page documents one metric in two tables with contradictory guidance, and `5.2.md` ends up with two `## Analytics` H2s.

**Merge order: #572 first, then this one.** #572 should own `transaction-commit-time` — it built a purpose-built section, its table shape (`path`/`method`/`type`) matches how the metric is actually emitted through `recordAction()` with no `byThread` flag, and it also fixes the stale raw-entry example. This PR should keep only the two queue-depth gauges and cross-link. The inline suggestions below do exactly that, and assume #572 has landed so `#storage-metrics` resolves. I've left the matching note on #572.

The Key attributes cell is a second reason to drop the row rather than relocate it: it lists `mean, median, p90, p95, p99, p999`, which both omits the raw fields (`distribution`, `count`) and silently drops four of the nine percentiles the aggregator actually emits — and it contradicts your own prose four lines down, which correctly says percentiles "aren't present on `hdb_raw_analytics`".

**Release note.** Once #572 lands, fold this entry under the `## Analytics` heading it creates (after `## Security`) instead of adding a second one between `## Configuration` and `## CLI`. Only one `## Analytics` should exist in `5.2.md`.

**One claim I checked that holds — keep it.** "Harper also logs once per stuck commit when the 503 check itself fires" is accurate. The log exists at `DatabaseTransaction.ts:394` (`v5.2.0`) and `:485` (`origin/main`), emitting `Rejecting writes on this thread: a commit has been outstanding for ...`, and the source comment literally reads "Log once per stuck commit (not once per rejected request, harper#2001)". Flagging this explicitly because an earlier grep of mine against an older ref missed it and I don't want it re-raised. One forward-looking note: on `main` that log has picked up an episode-level cooldown (`OVERLOAD_LOG_MIN_INTERVAL_MS`), so "once per stuck commit" weakens slightly next minor — nothing to change here, just worth knowing.

**PR body is stale.** It says "secondary/optional metric — see that PR for the open question on whether it ships." It shipped: `fe58546ec` is on `harper` `main`, along with `unitTests/resources/transactionQueueDepth.test.js`. Worth updating so the body doesn't read as speculative docs.

reference/analytics/overview.md
```diff
@@ -171,14 +171,71 @@ Harper automatically tracks the following metrics for all services. Applications
...
+
+| Field      | Unit  | Description                                             |
+| ---------- | ----- | ------------------------------------------------------- |
+| `depth`    | count | Instantaneous depth sampled at emit time                |
+| `maxDepth` | count | High-water mark since this thread's last emitted sample |
+
+- **`write-transaction-queue-depth`** counts write commits handed to the storage engine whose commit
+  promises have not yet settled — how many commits this thread is juggling concurrently. This is
+  in-flight, not durability: under `storage.writeAsync: true` a settled commit promise does not
+  guarantee the write has been synced to disk.
```
cb1kenobi: **Medium: the `storage.writeAsync` caveat can never apply to this metric**

This section (correctly) states two paragraphs down that both gauges are tracked only on the RocksDB path and read `0` on `storage.engine: lmdb`. But `storage.writeAsync` is consulted in exactly one place in `harper` — `utility/lmdb/OpenEnvironmentObject.ts:39-41`, where it sets LMDB's `noSync`. Nothing in the RocksDB path reads `STORAGE_WRITEASYNC` at all.

So the caveat describes a setting that only takes effect on the engine where this metric is pinned at zero: an operator who reads this and turns off `writeAsync` to "restore" durability for the commits this gauge is counting changes nothing about them.

The underlying point — a settled commit promise is not a durability guarantee — is still true on RocksDB, just for a different reason: `openRocksDatabase()` applies `options.disableWAL ??= true` (`resources/databases.ts:287`) to user data stores.

Suggested fix:
```suggestion
- **`write-transaction-queue-depth`** counts write commits handed to the storage engine whose commit
  promises have not yet settled — how many commits this thread is juggling concurrently. This is
  in-flight, not durability: a settled commit promise means the storage engine accepted the write,
  not that it has been synced to disk.
```

—
_Generated by Barber AI_

reference/analytics/overview.md
```diff
@@ -171,14 +171,71 @@ Harper automatically tracks the following metrics for all services. Applications
...
+- **`read-transaction-queue-depth`** counts concurrently open tracked read transactions, including
```
cb1kenobi: **Low: `read-transaction-queue-depth` also counts write-bearing transactions, not just reads**

The gauge is `trackedTxns.size` (`resources/DatabaseTransaction.ts:195`), and the single site that grows that set is `getReadTxn()` (`DatabaseTransaction.ts:414`) — which every `DatabaseTransaction` goes through, including ones that stage and commit writes (e.g. `Table.ts:2256`, `Table.ts:3040`, both on the write path). A transaction leaves the set only at commit/abort.

So under write load this metric rises with in-flight *write* transactions too, and "many short-lived reads or a few long-lived ones" isn't the complete set of explanations for a high value — which matters for the compaction inference this bullet draws, since a write transaction sitting in the set is a different diagnosis than a long-lived read snapshot.

Suggested fix: say it counts concurrently open tracked transactions holding a read handle — which includes write transactions, since those acquire one too — rather than "read transactions".

—
_Generated by Barber AI_

reference/analytics/overview.md
kriszyp: Addressed in b7a4594a: removed the inapplicable storage.writeAsync caveat and now states only the storage-engine acceptance-versus-disk-sync distinction.

— GPT-5 Codex

reference/analytics/overview.md
kriszyp: Addressed in b7a4594a: the table and detailed definition now describe open tracked transactions holding a read handle, explicitly including write-bearing and snapshot-disabled transactions.

— GPT-5 Codex

reference/analytics/overview.md
kriszyp: Addressed in b61c915e and a7a40707: the caveat now covers mixed installs reporting RocksDB activity only, explains the all-LMDB zero case, and separately states that the missing database dimension prevents attribution to a specific database.

— GPT-5 Codex

cb1kenobi:
Reviewed `a7a4070` — no issues found. This PR looks good, nice job!

—
_Generated by Barber AI_

PR #576, opened 2026-07-08
Gate the quota-hook example's counter table off the MCP surface

kylebernhardy:
Companion docs for **HarperFast/harper#1821**, which replaces the `mcp.<profile>.quota.*` config (a `resource:` reference to a public-by-default Resource) with a registration function, `server.setMcpQuotaHandler(fn)`.

## Change

Rewrites the durable-quota section of `reference/mcp/configuration.md` to the new API:
- **Removes** the `mcp.<profile>.quota.resource` / `.method` config keys (they no longer exist).
- **Documents** `server.setMcpQuotaHandler(handler)` — the policy is a plain function, never an exposed Resource, so it surfaces no `update_/delete_` MCP tools or REST/etc. a client could use to reset its own counter.
- The example now shows a registered function backed by an **internal** (`@table`, no `@export`) counter, plus a separate exported tool — the correct safe-by-default shape.

## History

This PR started as a fix to the old six-`exportTypes`-`false` workaround. @kriszyp's review correctly identified that the *implementation* (config references a public Resource) was the real problem, not the docs — so the fix moved into core (#1821), and this PR now documents that new API. The harper-mcp skill (HarperFast/skills#69, `generated` type) regenerates from this corrected reference.

Merges with #1821.

_Generated by an LLM (Claude Opus 4.8)._

### DIFF
--- a/reference/mcp/configuration.md
+++ b/reference/mcp/configuration.md
@@ -152,13 +152,7 @@ mcp:
 ```javascript
 // resources.js
 const DAILY_LIMIT = 100;
-export class McpQuota extends tables.QuotaCounter {
-	// Exporting makes the class config-addressable — but it would also surface
-	// update_/delete_McpQuota MCP tools and a REST endpoint, letting a
-	// permitted client reset its own counter. Keep the quota table off the MCP
-	// surface and restrict its REST permissions.
-	static exportTypes = { mcp: false };
-
+class McpQuota extends tables.QuotaCounter {
 	static async allowMcpCall({ identity, tool, user, profile, sessionId }) {
 		const id = identity ?? 'unknown';
 		const existing = await McpQuota.get(id);
@@ -170,6 +164,12 @@ export class McpQuota extends tables.QuotaCounter {
 		return true;
 	}
 }
+
+// Register the class so the quota hook can resolve it by name — WITHOUT
+// module-exporting it, which would surface update_/delete_McpQuota MCP tools
+// and a REST endpoint that let a permitted client reset its own counter.
+// exportTypes gates each transport independently (see the HTTP API reference).
+server.resources.set('McpQuota', McpQuota, { mcp: false, rest: false });
 ```
 
 ### `mcp.<profile>.quota.resource`
--- a/reference/mcp/configuration.md
+++ b/reference/mcp/configuration.md
@@ -153,6 +153,12 @@ mcp:
 // resources.js
 const DAILY_LIMIT = 100;
 export class McpQuota extends tables.QuotaCounter {
+	// Exporting makes the class config-addressable — but it would also surface
+	// update_/delete_McpQuota MCP tools and a REST endpoint, letting a
+	// permitted client reset its own counter. Keep the quota table off the MCP
+	// surface and restrict its REST permissions.
+	static exportTypes = { mcp: false };
+
 	static async allowMcpCall({ identity, tool, user, profile, sessionId }) {
 		const id = identity ?? 'unknown';
 		const existing = await McpQuota.get(id);
--- a/reference/mcp/configuration.md
+++ b/reference/mcp/configuration.md
@@ -118,7 +118,7 @@ Default: `0` (disabled)
 
 Sustained `tools/call` rate keyed on **client identity** rather than session (5.2.0+). Session-scoped buckets can be evaded by an anonymous client that cycles sessions (`initialize` → call → drop → repeat); the client bucket survives that loop. Identity is the client socket IP by default, or the value derived from [`identityHeader`](#mcpprofileratelimitidentityheader). Denials surface like other rate-limit hits: an `isError` tool result with `kind: 'rate_limited'`, `scope: 'per_client'`.
 
-Like the other buckets, state is in-memory per worker — it does not survive a restart and is not shared across workers. For durable quotas, see [`mcp.<profile>.quota.*`](#mcpprofilequota).
+Like the other buckets, state is in-memory per worker — it does not survive a restart and is not shared across workers. For durable quotas, see the [durable quota handler](#durable-quota-handler).
 
 ### `mcp.<profile>.rateLimit.perClientBurst`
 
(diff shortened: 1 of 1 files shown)


kriszyp (changes requested):
The `static exportTypes = { mcp: false }` line doesn't do what the comment claims — Harper's Resource/Table registration (`resources/jsResource.ts`, `resources/Resources.ts`) never reads an `exportTypes` static field off the exported class; that field only exists on the *registry entry*, populated via `@export(...)` (GraphQL schema tables) or a `server.resources.set(name, resource, exportTypes)` call. As written, `McpQuota` is registered with `exportTypes: undefined` and remains fully enumerated on both MCP and REST — exactly the hole this example is supposed to close.

Also worth fixing while here: even a working `{ mcp: false }` wouldn't restrict REST on its own (`Resources.getMatch`'s `!== false` defaulting means each transport is independently gated) — the comment's "restricts its REST permissions" claim doesn't follow from that key alone.

The mechanism that's actually wired today: `server.resources.set('McpQuota', McpQuota, { mcp: false, rest: false })`, documented in `reference/http/api.md`. Happy to help swap the example over to that if useful.

— KrAIs

reference/mcp/configuration.md
```diff
@@ -153,6 +153,12 @@ mcp:
 // resources.js
 const DAILY_LIMIT = 100;
 export class McpQuota extends tables.QuotaCounter {
+	// Exporting makes the class config-addressable — but it would also surface
+	// update_/delete_McpQuota MCP tools and a REST endpoint, letting a
+	// permitted client reset its own counter. Keep the quota table off the MCP
+	// surface and restrict its REST permissions.
+	static exportTypes = { mcp: false };
```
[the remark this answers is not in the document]
kylebernhardy: Superseded by bfcbd93: a static exportTypes field is never read off the class at all (see kriszyp's review), so neither { mcp: false } nor { mcp: false, http: false } as a static would have any effect. The example now uses the wired mechanism — server.resources.set('McpQuota', McpQuota, { mcp: false, rest: false }) — runtime-verified to hide both the MCP tools and the REST endpoint while the hook still resolves.

kriszyp (commented):
Nice catch tracing this down to actual `getMatch`/`exportTypes` call sites rather than just eyeballing the diff — this is a good find and worth closing all the way.

The fix un-exports `McpQuota` and sets `{ mcp: false, rest: false }` on the new `server.resources.set(...)` call (https://github.com/HarperFast/documentation/pull/576/files#diff-d541047bfcecdc219bc5e37f5e7ba72759822cb8dc1e849acab1c38953f7b11cR155), which closes the two transports the PR names. But `exportTypes` gates each transport independently, and the same generic `getMatch(path, exportType)` check is invoked with other keys that this PR leaves unset (unset defaults to exposed):

- `mqtt` — https://github.com/HarperFast/harper/blob/main/server/DurableSubscriptionsSession.ts#L188 and `:411`
- `ws` — https://github.com/HarperFast/harper/blob/main/server/REST.ts#L349
- `sse` — https://github.com/HarperFast/harper/blob/main/server/REST.ts#L38 (a distinct key from `rest`, not covered by `rest: false`)
- `graphql` — https://github.com/HarperFast/harper/blob/main/server/graphqlQuerying.ts#L309

Since `McpQuota extends tables.QuotaCounter` (inheriting standard Table CRUD), the same "permitted client resets its own quota counter" vulnerability this PR sets out to fix is still reachable through a GraphQL mutation, a WS write, or an MQTT publish to the resource's topic — just a different door than the one that got closed. This mirrors the doc's own precedent one screen up (https://github.com/HarperFast/harper/blob/main/reference/http/api.md#L281 — `{ rest: true, mqtt: false }`), which shows each key really is independent and unset ≠ off.

Two ways to close this all the way:
1. Extend the example at https://github.com/HarperFast/documentation/pull/576/files#diff-d541047bfcecdc219bc5e37f5e7ba72759822cb8dc1e849acab1c38953f7b11cR155 to `{ mcp: false, rest: false, ws: false, graphql: false, mqtt: false }`, or
2. If the intent is only to demonstrate MCP+REST gating, adjust the explanatory comment at https://github.com/HarperFast/documentation/pull/576/files#diff-d541047bfcecdc219bc5e37f5e7ba72759822cb8dc1e849acab1c38953f7b11cR172 so it doesn't read as "fully closed" when it's two of five surfaces — readers copy-pasting this example would otherwise inherit the same partial gate.

Either way keeps the example accurate for anyone who copies it verbatim. Thanks for pinning down the exact call sites — makes this easy to act on.

— Claude (Sonnet 5), reviewed via review-queue

kriszyp (commented):
Thanks for catching this — the fix itself is right. The pre-PR example used `export class McpQuota`, which auto-surfaced the inherited CRUD verbs (`update_/delete_McpQuota` as MCP tools, plus REST/SSE/WS/GraphQL/MQTT), so a client subject to the quota could reset its own counter. Registering via `server.resources.set(..., { …all false })` keeps it resolvable by the hook while removing that surface — strictly better. 👍

A few things to reconcile before merge:

**1. The PR description is stale and would mislead.** It says the fix adds `static exportTypes = { mcp: false }`, but the diff uses `server.resources.set(...)`. This matters: `static exportTypes` on an auto-exported class isn't honored (the auto-export path calls `set()` without reading a class static), so a reader who follows the description instead of the diff would leave the surface exposed. Please update the description to match the code.

**2. The prose right below still contradicts the example.** The `mcp.<profile>.quota.resource` section reads "Path of an **exported** Resource… an **exported subclass** (and its policy) wins on reload," while the example now deliberately does *not* export.
- Does a `server.resources.set()`-registered (non-exported) class actually resolve and "win on reload" the same way an exported subclass does? The prose specifically credits *export* for the reload behavior — if the non-exported path reloads differently, the example may not behave as documented.
- Either way, the prose and the example need to agree.

**3. The bigger concern — probably core, not this doc.** A quota counter is inherently *internal enforcement state*, yet it's exposed on every transport by default, and the only thing between "safe" and "any permitted client can reset its own quota" is the author remembering to set six flags to `false`. `unset means exposed` is a dangerous default for the one built-in whose entire purpose is server-side enforcement. Should `QuotaCounter` (or an internal-resource idiom) default to *unexported*, so the safe path is the default and we don't have to teach a six-false workaround? A docs example can only paper over this — the real fix looks like it belongs in core. Happy to file that if you agree.

None of this blocks the security fix. Mainly: stop the description + prose from pointing at the broken/contradictory approach, and decide whether the default should change upstream.

_🤖 Reviewed with KrAIs (Claude Opus 4.8)_

reference/mcp/configuration.md
kriszyp: I guess this is the fundamental problem; there is nothing wrong with the docs, but the problem is with the implementation. Referencing a resource from a config to define a hook is fundamentally wrong, when the default exposure of a resource is to the public. I think we need to fix how we configure a quote handler. The could be with a registration function that allows you specify a quota check function. Or it could be configured to reference a module with some export, but that seems clumsier.

kriszyp (changes requested):

reference/mcp/configuration.md
kylebernhardy: Agreed, and Kyle's called it — we'll take your direction. Pivoting the quota config away from referencing a Resource and to a **registration function** that takes the quota-check function directly, so the hook is never an exposed Resource and there's no six-`false` footgun to remember.

I'll draft the core API (a `server`-side `setMcpQuotaHandler`-style registration) and migrate `#1610`'s config-name resolution (`checkDurableQuota` → `getResources().get(path)`) over to it, then loop you in on the exact shape before it's final. Since this rewrites the whole `mcp.<profile>.quota.*` story, the doc change here will follow the new API rather than the six-`false` example — so I'll rework this PR (or land a fresh docs PR against the core change) once the implementation is in. Holding the current doc edits meanwhile.

Thanks for pushing to the root cause — this is the right fix.

Comment generated by kAIle (Claude Opus 4.8)

kylebernhardy: Done — this PR now documents the new API instead of the workaround. harper#1821 replaces `mcp.<profile>.quota.resource` config with `server.setMcpQuotaHandler(fn)` (the policy is a plain function, never an exposed Resource), and this rewrite of the durable-quota section reflects that: the config keys are gone and the example registers a function backed by an internal (non-`@export`) counter table. The two merge together.

Comment generated by kAIle (Claude Opus 4.8)

kriszyp (approved):
Approving — the shift from an exported Resource to a plain `setMcpQuotaHandler` function is a clean simplification and correctly keeps the internal quota table off the transport surface.

One perf nit worth taking if there's a follow-up pass: the example still increments/`put`s the counter before checking the limit, so a rate-limited client's rejected requests still cost a DB write each. Checking the limit first (read-only fast-path denial) would set a better precedent for anyone copying this example — not blocking, since it's a pre-existing pattern from the old example, not something this PR introduced.

PR #577, opened 2026-07-09
docs: fix incorrect @relation directive to @relationship

Ethan-Arrowood:
Fixes `@relation` to `@relationship` across the docs.

### DIFF
--- a/reference/resources/query-optimization.md
+++ b/reference/resources/query-optimization.md
@@ -95,12 +95,12 @@ Example of an indexed foreign key that enables efficient join queries:
 type Product @table {
 	id: Long @primaryKey
 	brandId: Long @indexed # foreign key — index this
-	brand: Related @relation(from: "brandId")
+	brand: Related @relationship(from: "brandId")
 }
 type Brand @table {
 	id: Long @primaryKey
 	name: String @indexed # indexed — enables efficient brand.name queries
-	products: Product @relation(to: "brandId")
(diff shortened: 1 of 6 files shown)


reference/resources/query-optimization.md
```diff
@@ -95,12 +95,12 @@ Example of an indexed foreign key that enables efficient join queries:
 type Product @table {
 	id: Long @primaryKey
 	brandId: Long @indexed # foreign key — index this
-	brand: Related @relation(from: "brandId")
+	brand: Related @relationship(from: "brandId")
```
cb1kenobi, line 98: **Low: two `@relation` occurrences missed — the rename is not complete "across the docs"**

The PR body says it fixes `@relation` to `@relationship` across the docs, but this same brand/product example still uses the invalid `@relation` in `release-notes/v4-tucker/4.3.0.md:22` and `:28`:

```graphql
brand: Related @relation(from: "brandId")
products: Product @relation(to: "brandId")
```

`@relation` was never a valid Harper directive — Harper's GraphQL schema handler only recognizes `relationship` (see `harper/resources/graphql.ts`, known-directive list and `directiveName === 'relationship'`), with no `relation` alias — so that release-note example would fail schema validation if copied. Update those two lines to `@relationship` for consistency with the rest of this change. (Commenting here because that file has no diff hunk to anchor to.)

—
_Generated by Barber AI_

cb1kenobi (approved):

PR #579, opened 2026-07-12
docs(env-vars): clarify loadEnv/.env cannot shape Harper's instance config

kriszyp:
## Summary

The Environment Variables page documents `loadEnv` / `.env` but doesn't state an important scope boundary that trips users up: **`.env` supplies *application* environment variables, not Harper's own configuration.**

Config-shaping variables (`HARPER_CONFIG`, `HARPER_SET_CONFIG`, `HARPER_DEFAULT_CONFIG`) delivered through a `.env` file are read **after** the instance config has already been composed, so they're ignored. Harper now logs a warning when it detects one, rather than silently no-op'ing. This adds a `:::warning` admonition to `reference/environment-variables/overview.md` documenting that behavior and pointing users to the config file / real process env as the correct mechanism.

Documents the warn-don't-honor behavior shipped in [harper#1513](https://github.com/HarperFast/harper/issues/1513) (fix PR #1580).

## Where to look
- One-paragraph warning admonition added after the "Load Order" section.
- `<VersionBadge type="changed" version="v5.2.0" />` — **please confirm the exact release** #1580 ships in (main is currently `5.2.0-alpha`, so v5.2.0 is my best estimate; adjust if it lands in a 5.1.x patch).

Surfaced by exploratory QA while regression-verifying #1513. 🤖 Generated by an LLM (KrAIs / Claude Opus 4.8).

### DIFF
--- a/reference/environment-variables/overview.md
+++ b/reference/environment-variables/overview.md
@@ -49,7 +49,9 @@ Because Harper is a single-process application, environment variables are loaded
 :::warning
 `loadEnv` supplies **application** environment variables — values your component code reads from `process.env` (secrets, API endpoints, app settings). It does **not** configure Harper itself.
 
-Harper's own instance-wide configuration is composed once at startup, **before** any component's `loadEnv` runs. As a result, config-shaping variables such as `HARPER_CONFIG`, `HARPER_SET_CONFIG`, and `HARPER_DEFAULT_CONFIG` delivered through a `.env` file are read too late to take effect and are ignored (Harper logs a warning when it detects one). <VersionBadge type="changed" version="v5.2.0" />
+Harper's own instance-wide configuration is composed once at startup, **before** any component's `loadEnv` runs. As a result, config-shaping variables such as `HARPER_CONFIG`, `HARPER_SET_CONFIG`, and `HARPER_DEFAULT_CONFIG` delivered through a `.env` file are read too late to take effect and are ignored (Harper logs a warning when it detects any of them).
+
+<VersionBadge type="changed" version="v5.2.0" />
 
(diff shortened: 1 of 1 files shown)


reference/environment-variables/overview.md
```diff
@@ -46,6 +46,14 @@ myApp:
 
 Because Harper is a single-process application, environment variables are loaded onto `process.env` and are shared across all components. As long as `loadEnv` is listed before dependent components, those components will have access to the loaded variables.
 
+:::warning
+`loadEnv` supplies **application** environment variables — values your component code reads from `process.env` (secrets, API endpoints, app settings). It does **not** configure Harper itself.
+
+Harper's own instance-wide configuration is composed once at startup, **before** any component's `loadEnv` runs. As a result, config-shaping variables such as `HARPER_CONFIG`, `HARPER_SET_CONFIG`, and `HARPER_DEFAULT_CONFIG` delivered through a `.env` file are read too late to take effect and are ignored (Harper logs a warning when it detects one). <VersionBadge type="changed" version="v5.2.0" />
```
cb1kenobi: **Nit: VersionBadge placement vs. repo convention**

CONTRIBUTING.md says to place `<VersionBadge>` "on its own line directly below the heading it describes" and to "reserve `<VersionBadge>` for standalone placement after headings" (mid-sentence use is called out as awkward). Here it trails the end of a paragraph inside the admonition. Consider moving it to its own line within the warning.

Separately, `type="changed"` is correct here and should stay — this documents a behavior change, which is exactly the `changed` case in CONTRIBUTING.md (contrary to the earlier suggestion on this line to drop it). The `v5.2.0` value is still the open question you flagged; confirm against where harper#1580 actually ships.

—
_Generated by Barber AI_

cb1kenobi (approved):

PR #583, opened 2026-07-13
docs: replicating the system database with a constrained topology (5.2)

kriszyp:
## Summary
Documents the 5.2 replication change that lets the `system` database replicate while keeping a constrained (non-mesh) topology.

- **`reference/replication/overview.md`** (Controlling Replication Flow): replaces the previous _"avoid replicating the `system` database"_ note with two-sided per-database `sendsTo`/`receivesFrom` examples, the exact receiver fallback behavior, and a `<VersionBadge type="changed" version="v5.2.0" />` subsection explaining constrained `system` replication, its opt-in nature, and its trust boundary.
- **`reference/replication/clustering.md`**: documents the v5.2 `add_node` / `set_node` database-scoping surface and its current peer-qualified-entry limitation.
- **`release-notes/v5-lincoln/5.2.md`**: adds a "Replicating the System Database with a Constrained Topology" section.

## Companion PR

Documents the merged [harper-pro PR #572, "Constrain replication mesh when the system database is replicated"](https://github.com/HarperFast/harper-pro/pull/572).

## Related issue

- [harper-pro#710, "add_node_back does not rewrite peer-qualified sendsTo/receivesFrom entries"](https://github.com/HarperFast/harper-pro/issues/710) tracks the Operations API limitation documented here.

## Where to look

The behavior-change framing, the two-sided route example and fallback rules, and the `add_node` caveat are the load-bearing parts of this change.

## Verification

- `npm run format:check`
- `npm run typecheck`
- `npm run lint`
- `npm run build` (passes; reports the known pre-existing broken `#deployment-operations` anchor in the v5.1 release notes)

_Drafted by Claude Opus 4.8; updated by GPT-5 Codex._

<sub>Human-Review-Need: 4 @ ef788bf49e9a</sub>


reference/replication/clustering.md
```diff
@@ -32,6 +32,9 @@ Adds a new Harper instance to the cluster. If `subscriptions` are provided, it c
   - `table` — table name
   - `subscribe` — if `true`, transactions on the remote table are replicated locally
   - `publish` — if `true`, transactions on the local table are replicated to the remote node
+- `sendsTo` / `receivesFrom` _(optional)_ — database-scoped controlled-flow entries for this node, mirroring the config-route `replicates.sendsTo` / `replicates.receivesFrom` syntax (see [Controlling Replication Flow](./overview.md#controlling-replication-flow)). Each entry is a database name or an object: `{ database, excludeTables? }`. When provided, replication with this node is restricted to the listed databases; to replicate all databases except specific tables, use a wildcard entry (`excludeTables` with no `database`). Mutually exclusive with `subscriptions`.
```
Ethan-Arrowood: **"Each entry is a database name or an object" is wrong**, and it fails silently. In the implementation a bare-string entry is matched as a **peer/node name**, never a database — see [`routeEntriesIncludePeer`](https://github.com/HarperFast/harper-pro/blob/63e0e1c469b58d69cefd5c57c68c4d27e890c1ec/replication/knownNodes.ts#L932) (`if (entry === peerName) return true`) and the gate fallback in [`shouldReplicateFromNode`](https://github.com/HarperFast/harper-pro/blob/63e0e1c469b58d69cefd5c57c68c4d27e890c1ec/replication/knownNodes.ts#L675) (`sendsTo === getThisNodeName()`). So a user writing `"sendsTo": ["cardata"]` authorizes nothing at all and gets no error — replication simply doesn't happen.

Fix: `"Each entry is an object `{ database, excludeTables? }`"` — or document strings as node names, matching `computeSelfReplicates`'s `{ target: e }` normalization.

---

**Separately, the `sendsTo` direction is ambiguous — and the implementation's perspective is the opposite of config routes.** In config routes, `replicates.sendsTo` means "the *local* node sends these to the route's peer" (`computeSelfReplicates` puts them in the self record). In `add_node`, `req.sendsTo` is written verbatim into the **added peer's** record ([`setNode.ts:201`](https://github.com/HarperFast/harper-pro/blob/63e0e1c469b58d69cefd5c57c68c4d27e890c1ec/replication/setNode.ts#L201)), where the receive gate reads a record's `sendsTo` as "this peer sends to me" — i.e. `add_node`'s `sendsTo` means "databases the *added node* sends to the local node".

"Mirroring the config-route syntax" invites the local-perspective reading, which is inverted. Symmetric configs (same list in both keys) mask it; one-way flows silently go the wrong direction. Is the added-node perspective intended, or should `setNode` match the config-route perspective? No test pins it today.

---

**Minor:** "mutually exclusive with `subscriptions`" is guidance, not enforcement — the Joi schema accepts both and `setNode` silently prefers `subscriptions` (`if (req.subscriptions) … else if (req.sendsTo || req.receivesFrom)`). Either say "if both are provided, `subscriptions` takes precedence", or file a harper-pro validation follow-up.

reference/replication/clustering.md
```diff
@@ -32,6 +32,9 @@ Adds a new Harper instance to the cluster. If `subscriptions` are provided, it c
...
+
+  > **Note**: `sendsTo` / `receivesFrom` here scope only the connection to _this_ node — they don't change how the local node advertises itself to the rest of the cluster. A node's own directional (non-mesh) `hdb_nodes` self-record is derived solely from its `harper-config.yaml` routes, so replicating `system` without collapsing to a full mesh requires directional routes in config, not just `add_node`/`set_node` scoping (see [Replicating the `system` database with controlled flow](./overview.md#replicating-the-system-database-with-controlled-flow)).
```
Ethan-Arrowood: "Scope only the connection to *this* node" overstates the containment, and it breaks in exactly the scenario this PR documents (replicated `system`). Entries written by `add_node` carry no `target`/`source`, and the peer record replicates cluster-wide; `routeEntriesIncludePeer` treats an absent `target`/`source` as "any peer", so a third node holding the same database can subscribe to the added peer under those entries. Note that `computeSelfReplicates` deliberately stamps `target`/`source` ("fully qualified… for the discovered-peer gates") while `setNode` does not.

The note's main point — that it doesn't make *your* node directional — is correct; it's the "restricted to this connection only" guarantee that doesn't hold once the record propagates. (Same phrase appears in the 5.2 release notes.) Arguably the real fix is upstream: should `add_node`-written entries stamp `target`/`source` like `computeSelfReplicates` does, making the claim true?

Ethan-Arrowood (changes requested):
The core 5.2 constrained-topology content is accurate — I traced every major claim to merged harper-pro#572 and its source (`computeSelfReplicates`, `shouldReplicateFromNode`, `setNode.ts`) and it holds up. The `host:`→`hostname:` fix the Gemini bot asked for is in at this head too.

The problem is concentrated in the new `add_node` parameter bullet, which has a factual error serious enough to break a user's setup silently: a bare-string `sendsTo` entry is matched as a **peer name**, not a database name, so the documented `"sendsTo": ["cardata"]` authorizes nothing and replication just doesn't happen. Details inline, along with a direction-perspective problem and an overstated containment claim.

One of these may actually be an implementation-intent question rather than a docs bug — see the `sendsTo` direction comment. Since no test pins `add_node`'s direction, it'd be worth confirming with you directly which behavior is intended; the doc needs to state a direction explicitly either way, and a pinning integration test would settle it permanently.

cb1kenobi (approved):

reference/replication/overview.md
```diff
@@ -198,7 +200,41 @@ replication:
...
+    - hostname: node-two
+      replicates:
+        sendsTo:
+          - database: config # push central config downstream
+          - database: system # push central config (users, roles, schemas) downstream
+        receivesFrom:
+          - database: cardata # aggregate telemetry upstream
+```
+
+`sendsTo` / `receivesFrom` are declared from the perspective of the node whose `harper-config.yaml` they're in, for its route to that one peer, and — because a directional route also gates what it's willing to send — both sides normally need a matching entry. To aggregate a database upstream instead of pushing it downstream — for example, so a role created on a roadside node reaches a middle-tier node — the **roadside** node's route to middle needs `sendsTo: [{ database: system }]`, and the **middle-tier** node's route to roadside needs the matching `receivesFrom: [{ database: system }]`. If the middle tier is missing its `receivesFrom` half, it never attempts the subscription; if roadside is missing its `sendsTo` half, middle's subscription attempt is rejected as unauthorized.
```
Ethan-Arrowood, line 221: **A matching `receivesFrom` is not required when the peer has no directional route for that neighbor** — and this contradicts the YAML example directly above, which shows `sendsTo` on one side only and presents it as working (it is).

[`shouldReplicateFromNode`](https://github.com/HarperFast/harper-pro/blob/9cd884dca/replication/knownNodes.ts#L697) gates on the config route only when it's present and object-shaped:

```js
const configRoute = (node as any).configRouteReplicates; // undefined when no route matches this peer
if (configRoute && typeof configRoute === 'object') {
  peerFeedsUs = configRoute.receives || routeEntriesIncludePeer(configRoute.receivesFrom, node.name, databaseName);
} else {
  peerFeedsUs = node.replicates?.sends || node.replicates?.sendsTo?.some?.(…); // the peer's ADVERTISED record
}
```

`configRouteReplicates` is `matchingRoute ? matchingRoute.replicates : undefined`, and a plain `- hostname: roadside` route normalizes to boolean `true` in `iterateRoutes` — also not the object form. Either way middle falls through to roadside's advertised `hdb_nodes` record, which `computeSelfReplicates` has already fully qualified with `target`, so the fallback matches and middle subscribes with no `receivesFrom` of its own. That's the canonical 5.2 shape.

The send-authority half is correct — I confirmed roadside closes with `1008 Unauthorized database subscription`.

```suggestion
`sendsTo` / `receivesFrom` are declared from the perspective of the node whose `harper-config.yaml` they're in, for its route to that one peer. Because a directional route also gates what a node is willing to send, the **sending** side always needs the matching `sendsTo` entry. To aggregate a database upstream instead of pushing it downstream — for example, so a role created on a roadside node reaches a middle-tier node — the **roadside** node's route to middle needs `sendsTo: [{ database: system }]`; without it, middle's subscription attempt is rejected as unauthorized. The **receiving** side needs a matching `receivesFrom` only when it has its own directional route for that peer: a middle-tier node with a directional route to roadside is gated by that route, so omitting `receivesFrom` there means it never attempts the subscription. If middle has no route to roadside at all — or only a plain, non-directional one — it falls back to roadside's advertised registry record and subscribes on the strength of roadside's `sendsTo` alone. Omitting `receivesFrom` is therefore not a way to block inbound replication.
```

reference/replication/overview.md
```diff
@@ -198,7 +200,41 @@ replication:
...
+
+Notes and current limitations:
+
+- This applies only when a node has **directional** routes (`replicates` with `sends`/`receives` or `sendsTo`/`receivesFrom`). A node with no directional routes keeps the legacy full-mesh advertisement.
+- This constrains replication subscriptions only. On-demand residency/retrieval connections (for example, sharded or invalidated-cache reads) use a separate mechanism governed by data residency, not by this registry record, and can still open a direct socket to a non-neighbor node.
+- Central visibility of every node is not guaranteed: an aggregation node may not list every distant leaf in its `hdb_nodes` registry (the registry relay differs from data relay). This does not open a connection either way.
+- Route changes to a node's own directionality take effect on restart.
+- Replicating `system` upstream (edge → core) propagates `hdb_user`/`hdb_role` along with everything else in the database: a role or user created — or a compromised edge node's route table altered — anywhere on the upstream path reaches every node it flows to. Weigh this against your trust boundary for edge nodes before routing `system` upstream from them.
```
Ethan-Arrowood, line 237: Minor, on an otherwise good addition: "route table" conflates two things. Routes come from `harper-config.yaml` and aren't replicated — what replicates is the `hdb_nodes` registry. A node's *own* row gets rewritten from its config on restart, so the poisonable surface is peers' copies of *other* nodes' rows.

```suggestion
- Replicating `system` upstream (edge → core) propagates `hdb_user`/`hdb_role` along with everything else in the database: a role or user created — or a compromised edge node's `hdb_nodes` registry rows altered — anywhere on the upstream path reaches every node it flows to. Weigh this against your trust boundary for edge nodes before routing `system` upstream from them.
```

reference/replication/overview.md
```diff
@@ -282,7 +318,7 @@ The following data operations are replicated across the cluster:
 
 **Destructive schema operations are not replicated**: `drop_database`, `drop_table`, and `drop_attribute` must be run on each node independently.
 
-Users and roles are not replicated across the cluster.
+Users and roles are not replicated across the cluster by default. As of v5.2, they do propagate when the `system` database (where `hdb_user` and `hdb_role` live) is included in replication — see [Replicating the `system` database with controlled flow](#replicating-the-system-database-with-controlled-flow).
```
Ethan-Arrowood, line 321: This version-gates the wrong half. Replicating `system` propagated `hdb_user`/`hdb_role` before 5.2 too — the old guidance was to disable `system` replication for a tight topology, *which also lost global users/roles/schema propagation*. What 5.2 changed is that you can now do it without collapsing to a full mesh.

```suggestion
Users and roles are not replicated across the cluster by default. They do propagate when the `system` database (where `hdb_user` and `hdb_role` live) is included in replication; as of v5.2 this no longer forces a full mesh — see [Replicating the `system` database with controlled flow](#replicating-the-system-database-with-controlled-flow).
```

Ethan-Arrowood (changes requested):
My earlier CHANGES_REQUESTED at `3ac6fee` is stale — **all four of those findings are genuinely fixed**, and I re-traced each against merged [harper-pro#572](https://github.com/HarperFast/harper-pro/pull/572) rather than taking the wording at face value:

- Bare-string entries are now correctly documented as peer names, not database names — matches `routeEntriesIncludePeer`.
- The `add_node` direction is now stated explicitly as the **added node's** perspective. I re-derived it from `setNode` writing `replicates.sendsTo` onto the peer's row and the receive gate reading a row's `sendsTo` as "this peer sends to me." As documented.
- The "scopes only the connection to this node" containment claim is gone, replaced with the accurate opposite.
- `subscriptions` precedence is now described as behavior rather than enforcement.

The unprompted additions — the `VersionBadge` tags, the Gossip Discovery full-mesh caveat, and especially the users/roles trust-boundary bullet — are all good and correct.

Re-requesting changes for one thing that is **new in these commits**, not a leftover: the two-sided-config paragraph added in `9c07b2b9`/`8b8328e4` states a hard requirement the implementation doesn't have, and it contradicts the one-sided YAML example three lines above it. The send-authority half is right; the receive half isn't. `shouldReplicateFromNode` only consults a config route when `configRouteReplicates` exists *and* is the directional object form — a missing route, or a plain `- hostname: roadside` route (which `iterateRoutes` normalizes to boolean `true`), both fall through to the peer's advertised `hdb_nodes` record. Since `computeSelfReplicates` stamps `if (!entry.target) entry.target = peer`, roadside advertises `sendsTo: [{ database: 'system', target: 'middle' }]` and the fallback matches it exactly. That fallback *is* the 5.2 feature — roadside dials middle, middle knows roadside only by gossip, and no `receivesFrom` on middle is needed. Suggestion inline.

The sharper risk is the second-order read: this paragraph invites treating a missing `receivesFrom` as a *constraint*. It isn't. A reader who omits it believing that blocks inbound `system` replication gets it anyway — immediately after a new trust-boundary warning about exactly that data.

Two smaller inline suggestions on other newly added lines, both non-blocking.

Worth noting this hinges on one `undefined` fallback branch, so if you read `shouldReplicateFromNode` differently I'd rather resolve it here than land either version. And the real question underneath: is the two-sided requirement the *intended* contract — should the discovered-peer fallback be tightened so an explicit `receivesFrom` really is required? If so that's a harper-pro change, not prose.

Two unrelated asks before merge, neither blocking the inline fix:

1. The `add_node` bullet in `clustering.md` is now the sole record of an unfixed harper-pro defect (`add_node_back` carries the entry array over without rewriting per-entry peer references). I traced it and the claim is real, but I searched harper-pro and there's no issue tracking it and no test covering `add_node` with `sendsTo`/`receivesFrom` at all. A prose warning with no owner and no expiry silently becomes wrong the day someone fixes the code. Please file the harper-pro issue and link it from the bullet so the caveat has something to retire it. Related: the bullet leans on `target`/`source`, which aren't documented anywhere in this repo — either define them there or just say peer-scoping doesn't work through `add_node`, use config routes.
2. The PR body still says the harper-pro change is on branch `kris/systemdb-routing-repro` with a "feature PR to be opened." That's harper-pro#572, merged. Please update the link.

reference/replication/overview.md
kriszyp: Addressed in e729648b: the docs now show both sides of the directional route, state that the sender always needs sendsTo, and explain that a receiver without an object-shaped directional route falls back to the sender self-record. That self-record qualifies entries from the enclosing route hostname, so middle matches while a non-neighbor such as core does not. The text also scopes the warning about omitted receivesFrom to receivers without a directional route. — GPT-5 Codex

reference/replication/overview.md
kriszyp: Addressed in 2e7319e6 and refined in a7d49d86: route table is now hdb_nodes registry rows. The warning also states the concrete privilege and advertised-directionality risk and gives downstream-only system routing as the mitigation for untrusted edges. — GPT-5 Codex

reference/replication/overview.md
kriszyp: Addressed in 2e7319e6: the text now says users and roles propagate whenever system is replicated; only the ability to keep that replication from forcing a full mesh is version-gated to v5.2. — GPT-5 Codex

reference/replication/overview.md
```diff
@@ -198,7 +200,60 @@ replication:
...
...tional route also gates what a node is willing to send, the **sending** side always needs the matching `sendsTo` entry. To aggregate a database upstream instead of pushing it downstream — for example, so a role created on a roadside node reaches a middle-tier node — the **roadside** node's route to middle needs `sendsTo: [{ database: system }]`; without it, middle's subscription attempt is rejected as unauthorized. The **receiving** side needs a matching `receivesFrom` only when it has its own directional route for that peer: a middle-tier node with a directional route to roadside is gated by that route, so omitting `receivesFrom` there means it never attempts the subscription. If middle has no route to roadside at all — or only a plain, non-directional one — it falls back to roadside's advertised `hdb_nodes` self-record. That self-record qualifies each `sendsTo` entry with the hostname of the route that produced it: middle subscribes because roadside's route to middle carries `database: system`, while core appears in no such route and matches no entry. On a receiver with no directional route for the peer, omitting `receivesFrom` is therefore not a way to block inbound replication.
```
cb1kenobi: **Low: the send side isn't unconditional either — "always needs the matching `sendsTo` entry" has the same route-presence caveat this paragraph spells out for the receive side**

The receive half is now correctly qualified ("only when it has its own directional route for that peer"), but the send half is stated as unconditional. The send-authority gate is gated on the same precondition: `getConfigRouteReplicates(options, remoteNodeName)` is only consulted when it returns the directional object form, so `configSendDecision` is `undefined` for a peer reached by a plain (non-directional) route, and the handler falls through to the dynamic `hdb_nodes` path — which authorizes the send on the *subscriber's* advertised `receivesFrom`/`receives`, with no `sendsTo` of ours required. A node with directional routes to some peers and a plain route to one peer will still serve that peer.

This matters for the trust-boundary bullet below (line 256), which offers "omit it from the edge node's `sendsTo`" as the way to keep `system` from flowing upstream. That control binds only while the edge's route to its parent is directional — exactly the symmetric hazard the last sentence of this paragraph warns about for `receivesFrom`.

Suggested fix: mirror the qualification, e.g. "the **sending** side needs the matching `sendsTo` entry whenever it has a directional route for that peer; a node whose route to the peer is plain or absent falls back to the peer's advertised `receivesFrom` and will serve it regardless."

—
_Generated by Barber AI_

kriszyp: Addressed in c9802c57 and ef788bf4: the sender rule is now qualified like the receiver rule. A sender with no directional route falls back to the subscriber’s advertised `receivesFrom`, and the downstream-only `system` guidance now requires directional routes on both nodes. I also replaced the ambiguous “object-shaped” wording with the exact directional fields and version-gated per-database controlled flow to v5.2. — GPT-5 Codex

reference/replication/overview.md
```diff
@@ -51,6 +51,8 @@ You can also manage nodes dynamically through the [Operations API](./clustering.
 
 Harper automatically replicates node information to other nodes in the cluster using [gossip-style discovery](https://highscalability.com/gossip-protocol-explained/). This means you only need to connect to one existing node in a cluster, and Harper will automatically detect and connect to all other nodes bidirectionally.
 
+As of v5.2, this full-mesh, bidirectional auto-connect behavior applies to nodes with no directional routes. A node configured with [directional routes](#controlling-replication-flow) advertises a constrained registry record instead, so discovered non-neighbor nodes do not receive a replication connection — see [Controlling Replication Flow](#controlling-replication-flow).
```
cb1kenobi: **Medium: this overstates the guarantee — a directional node still opens a subscription to a discovered non-neighbour that advertises the legacy full-mesh record**

"A node configured with directional routes advertises a constrained registry record instead, so discovered non-neighbor nodes do not receive a replication connection" reads as: making *your* node directional is sufficient. It isn't — the containment depends on the **other** node's advertised record, not on this one's.

Traced through `harper-pro` at `origin/main`:

- `shouldReplicateFromNode` (`replication/knownNodes.ts:676`) resolves `configRouteReplicates` from *this* node's route to the peer. A discovered non-neighbour has no route, so it is `undefined` and the `else` branch runs.
- That branch reads the **peer's** advertised `hdb_nodes` record: `typeof node.replicates === 'object' ? … : node.replicates`. A legacy full-mesh peer advertises `replicates: true`, so `peerFeedsUs` is `true`.
- The remaining conjuncts pass: `selfNodeReplicates(…)` reads this node's own record, which is the directional `{ sendsTo, receivesFrom }` object — truthy.

So a directional node subscribes to a legacy non-neighbour. The detailed section later in this same PR states this correctly ("A node that advertises the legacy full-mesh record can still be reached through the advertised-record fallback when a receiver has no directional route for it", line 262), so the summary here contradicts it — and this is the sentence most readers meet first.

Suggested fix: carry the precondition up here, e.g. "…so discovered non-neighbor nodes that also advertise a directional record do not receive a replication connection".

—
_Generated by Barber AI_

reference/replication/overview.md
```diff
@@ -198,7 +200,70 @@ replication:
 
 In this example, the local node only receives from `node-two` (one-way inbound) and only sends to `node-three` (one-way outbound).
 
-> **Note**: When using controlled flow replication, avoid replicating the `system` database. The `system` database contains node configurations, so replicating it would cause all nodes to have identical (and incorrect) route configurations.
+### Per-database controlled flow
+
+<VersionBadge version="v5.2.0" />
```
cb1kenobi: **Medium: this section is version-gated to v5.2.0, but per-database `sendsTo` / `receivesFrom` config routes shipped in the 5.1 line**

Verified with `git tag --contains` on the actual merge commits in `HarperFast/harper-pro` (tags fetched fresh and cross-checked against `git ls-remote --tags`; the local `v5.1.0` object `1a58b2b9` matches the remote exactly, so this is not a stale-tag artefact):

- The `sendsTo` / `receivesFrom` route surface, including per-entry `database` matching and `iterateRoutes` normalising route-level `sendsTo`/`receivesFrom`, came in with `ed9cca06` — "feat(replication): per-route excludeTables for bridge migrations (#240)", 2026-06-02. Earliest stable tag containing it: **v5.1.0**. It is in no `v5.0.x` tag.
- Enforcing the local config route's own direction (`configRouteReplicates` / `routeEntriesIncludePeer`, which is what makes the two-sided example on this page behave as described) came in with `3af29c08` — "(#506)". Earliest stable tag: **v5.1.15**.
- `v5.2.0` is correct for the *directional self-record* (`computeSelfReplicates`, harper-pro#572, merge `0b49587f`) — i.e. the badge on "Replicating the `system` database with controlled flow" below is right. It is this section's badge that is wrong.

As written, a reader on 5.1 concludes the feature does not exist for them.

Suggested fix: change this badge to `<VersionBadge version="v5.1.0" />` (optionally noting that the local-route enforcement described further down landed in v5.1.15), and leave the v5.2.0 badge on the `system`-database section.

—
_Generated by Barber AI_

reference/replication/clustering.md
```diff
@@ -32,6 +32,9 @@ Adds a new Harper instance to the cluster. If `subscriptions` are provided, it c
...
...to the added peer carries the array over without rewriting each entry's peer reference for the new direction, so a bare string or a `target`/`source` value that's correct for your side of the connection ends up wrong on the peer's copy, and the peer's send-authority check silently rejects the subscription — replication doesn't happen ([harper-pro#710 - `add_node_back` does not rewrite peer-qualified `sendsTo`/`receivesFrom` entries](https://github.com/HarperFast/harper-pro/issues/710)). Use [config routes](./overview.md#controlling-replication-flow) for anything that needs to be scoped to one peer. Unlike a config route's `replicates.sendsTo` / `replicates.receivesFrom`, which describes the _local_ node's own direction, these describe the **added node's** perspective: `sendsTo` lists what the added node sends (what this node receives from it), and `receivesFrom` lists what the added node receives (what this node sends to it). If both `subscriptions` and `sendsTo`/`receivesFrom` are provided, `subscriptions` takes precedence and `sendsTo`/`receivesFrom` are ignored. To replicate all databases while excluding specific tables, use a wildcard entry (`excludeTables` with no `database`).
```
cb1kenobi: **Medium: the v5.2.0 badge is wrong (this parameter shipped in v5.1.0), and the change 5.2 *did* make to it is a silent behaviour change that goes unmentioned**

Two separate problems, both verified against `HarperFast/harper-pro`:

**1. The badge.** `add_node` / `set_node` have accepted `sendsTo` / `receivesFrom` since `ed9cca06` ("feat(replication): per-route excludeTables for bridge migrations (#240)", 2026-06-02), which added `sendsTo: Joi.array()` / `receivesFrom: Joi.array()` to `replication/setNode.ts`'s validation schema. `git tag --contains` on that commit gives **v5.1.0** as the earliest stable tag (tags freshly fetched; local `v5.1.0` `1a58b2b9` matches `ls-remote`; contained in no `v5.0.x`).

**2. What 5.2 actually changed, and the upgrade hazard it creates.** In 5.1 `setNode` wrote

```js
nodeRecord.replicates = { sends: true, sendsTo: req.sendsTo, receivesFrom: req.receivesFrom };
```

and `shouldReplicateFromNode` short-circuits on `node.replicates?.sends` before it ever looks at `sendsTo`. So on 5.1 an `add_node` carrying `sendsTo` authorized **every** database — the database scoping was inert. harper-pro#572 (`0b49587f`, v5.2.0) dropped the blanket `sends: true`, so the same call now restricts to the databases named.

That means an operator who already uses `add_node` with `sendsTo` gets a real replication-scope change on upgrade with no config change on their side — databases that were replicating stop. Neither this bullet nor the 5.2 release note mentions it; the release note (`release-notes/v5-lincoln/5.2.md:113`) instead presents the parameter itself as new in 5.2.

Suggested fix: change the badge to `<VersionBadge version="v5.1.0" />` (or `<VersionBadge type="changed" version="v5.2.0" />` if the intent is to flag the behaviour change), and add a line to the 5.2 release note along the lines of: "Before 5.2 an `add_node` carrying `sendsTo` / `receivesFrom` also set `sends: true` on the peer record, which authorized every database and made the entries inert; as of 5.2 the entries restrict replication to the databases they name. Re-check any existing `add_node` scoping before upgrading."

—
_Generated by Barber AI_

cb1kenobi:
Reviewed `c3e63752` — re-review of the three open Mediums from the last round. All resolved; no new findings.

- **`overview.md:54`** (containment overclaim) — fixed. The Gossip Discovery summary now says a directional node still opens a subscription to a discovered peer that advertises the legacy full-mesh record, matching the limitations bullet ~200 lines later. Confirmed against `shouldReplicateFromNode`'s advertised-record fallback in `harper-pro` (`replication/knownNodes.ts`).
- **`overview.md:205`** (badge) — fixed. Badge is now `v5.1.0`, with a new paragraph explaining the `v5.1.15` enforcement gap. Confirmed via `git tag --contains` on `ed9cca06` (earliest tag `v5.1.0`) and `3af29c08` (earliest tag `v5.1.15`).
- **`clustering.md:35`** (badge) — fixed, and goes past the badge: the new "Changed in v5.2" callout explains the `sends: true` short-circuit removal (narrowing) and the `receivesFrom` wildcard widening. Traced both halves in source: `setNode.ts` drops `sends: true` exactly at `0b49587f` (harper-pro#572, tagged `v5.2.0`), and the send-authority gate in `replicationConnection.ts` switches from a strict `source`+`database` match to the wildcard-capable `routeEntriesIncludePeer` at the same commit — with an inline comment crediting the exact same reasoning.

All three tag objects (`v5.1.0`, `v5.1.15`, `v5.2.0`) were fetched fresh and cross-checked against `git ls-remote --tags` before use — none stale.

Also re-verified per the "must stay true" list: both replication gates, the two-sided YAML example (traces correctly through both sides), the `add_node_back` / [harper-pro#710, "add_node_back does not rewrite peer-qualified sendsTo/receivesFrom entries"](https://github.com/HarperFast/harper-pro/issues/710) caveat (issue still open), `replication.databases` as an allow-list, and the `hdb_secret` confidentiality bullet (Harper Pro's custody-key architecture backs this). No convergence/ordering/durability overclaim in the new text.

This PR looks good, nice job!

—
_Generated by Barber AI_

cb1kenobi:
Reviewed `3650e85812c8661b3a5c9c30a588bdb82eaadce6` — no issues found. This PR looks good, nice job!

Confirmation pass over the one new commit (`docs: add an add_node controlled-flow example and the durability boundary`). Merge base `6bf676d4` computed separately for both heads and identical, so the delta is exactly this commit; `git range-diff` shows the other 18 unchanged.

The new durability claim is the substantive addition, and it checks out against `harper-pro` at `v5.2.5`:

- `ensureThisNode` (`replication/subscriptionManager.ts:574`) recomputes the self-record with `computeSelfReplicates(iterateRoutes(options))` and structurally diffs it, rewriting the row when it differs — matching "rewrites its own `hdb_nodes` row from its `harper-config.yaml` routes".
- Both trigger paths are real: the in-code comment records that `startOnMainThread` runs once per process so a route change lands "on the next restart — where this rewrite fires", and the watcher is keyed so "a `deploy_component` reload (which re-fires this callback)" also rewrites.
- `computeSelfReplicates` ends with `if (!sawDirectional) return true;` — "Opt-in: no directional route → preserve the legacy full-mesh self-record" — which is exactly the "goes back to advertising the legacy full-mesh record" half.

The new request example matches the handler: `setNode.ts:21` accepts top-level `sendsTo` in the Joi schema, and `setNode.ts:224` carries `{ sendsTo: req.sendsTo, receivesFrom: req.receivesFrom }` onto the added node's peer record, so the "read from the added node's perspective" gloss is the right direction.

Version claims were re-verified with `git tag --contains` on the real merge commits, against tags force-refetched and cross-checked against `git ls-remote --tags` (all 95 harper-pro tags matched the remote exactly — no stale tag object):

| claim | merge commit | earliest stable tag |
| --- | --- | --- |
| `sendsTo`/`receivesFrom` accepted since v5.1.0 | `ed9cca0` (#240) | **v5.1.0** |
| stored but not enforced until v5.1.15 | `3af29c0` (#506) | **v5.1.15** (no v5.1.0–v5.1.14 tag contains it) |
| gating behavior as of v5.2 | `0b49587f` (#572) | **v5.2.0** |

Re-checked the standing items: the `overview.md:54` containment statement is untouched and still agrees with the limitations bullet at line 276 — both condition the fallback on the discovering node having no directional route for that peer. The `clustering.md:35` explanation still carries both halves, and both are in `0b49587f`: the `sends: true` short-circuit removal (`setNode.ts:219` — "NO blanket `sends: true`: that short-circuits shouldReplicateFromNode's else-branch to 'feeds every database'") and the `receivesFrom` wildcard widening (`routeEntriesIncludePeer`, `knownNodes.ts:1102` — absent peer or absent `database` matches any). No convergence, ordering, or durability overclaim in the added text; the new material is a stated limitation, not a guarantee.

Relative-link anchors in all three changed files resolve. The `/reference/v5/...` links in the release notes are Docusaurus versioned routes, not filesystem paths. Pre-existing broken anchors in untouched files were left alone. CI is green (6 passing, 1 skipped, none pending).

—
_Generated by Barber AI_

PR #585, opened 2026-07-14
Document the @expiresAt schema directive

kriszyp:
Adds a reference section for the `@expiresAt` field directive, which was previously undocumented (only `@createdTime`/`@updatedTime` were).

Covers:
- `@expiresAt` marks an absolute per-record expiry timestamp (epoch ms)
- authority over the table-level `@table(expiration:)` default, in both directions
- put-clears / patch-preserves semantics (session-renewal must re-include the field)
- caching-table boundary: use `context.expiresAt` on sourced tables

Companion to harper#1810 (the code change making the field authoritative). Cross-linked there.

_Generated by KrAIs (Claude Opus 4.8)._

### DIFF
--- a/reference/database/schema.md
+++ b/reference/database/schema.md
@@ -324,6 +324,24 @@ type Event @table {
 }
 ```
 
+### `@expiresAt`
+
+Marks a field as the record's absolute expiration time (Unix epoch milliseconds). Once the timestamp is in the past, the record is treated as expired: it is hidden from reads and physically removed by the eviction sweep. Useful for session, token, or per-record cache-entry tables where each record carries its own expiry.
+
+```graphql
+type Session @table {
+	id: ID @primaryKey
+	token: String
+	expiresAt: Long @expiresAt
+}
+```
+
+The `@expiresAt` field is authoritative over the table-level [`expiration`](#table) default, in both directions — a per-record value may extend a record past, or expire it before, the table default. When a record omits the field, the table default (if any) applies.
+
+- The field must be an absolute timestamp (Unix epoch milliseconds), not a duration. Negative values are ignored and fall back to the table default.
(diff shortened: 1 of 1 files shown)


reference/database/schema.md
```diff
@@ -324,6 +324,24 @@ type Event @table {
 }
 ```
 
+### `@expiresAt`
```
cb1kenobi, line 327: **Nit: add a `<VersionBadge>` for consistency**

The `@expiresAt` authoritative behavior is new (companion code in harper#1812, not yet released). Every other recently-introduced field directive in this file carries a version tag — e.g. `@embed` has `<VersionBadge version="v5.1.0" />` and `@indexed` has `<VersionBadge type="changed" version="v4.4.0" />`. Adding a badge here (matching #1812's release version) keeps the section consistent and tells readers which version to require.

(Purely a consistency nicety — the documented behavior itself checks out against the #1812 code and tests. `@createdTime`/`@updatedTime` also lack badges, so this is optional.)

—
_Generated by Barber AI_

cb1kenobi: Do we even know which version `@expiresAt` was introduced?

cb1kenobi (approved):

reference/database/schema.md
kriszyp: I would guess around 4.1.

Ethan-Arrowood (commented):
One thing to flag before this lands: the both-directional authoritative behavior for `@expiresAt` (a record's value vs the table default) ships only in v5.2.0 — harper#1812 is contained in 5.2.0-beta.2, with no 5.1.x tag — but the section has no `<VersionBadge>`. A reader on 5.1.x would be misled about what's available. Worth adding the 5.2.0 badge to that section.

Ethan-Arrowood (approved):

PR #586, opened 2026-07-15
docs: add Multiple Applications on One Cluster guide

BboyAkers:
## Summary

Adds a new developer guide, **Multiple Applications on One Cluster**, covering how to run multiple applications on a single Harper cluster:

- Applications as components on a shared cluster
- Structuring each application with `config.yaml`
- Keeping applications separate along three boundaries: **routes, data, and roles**
- Registering applications declaratively (`harperdb-config.yaml`) and imperatively (CLI)
- Deploying across the cluster with `replicated=true`
- Guidance on when to co-locate vs. use separate clusters

Written in the platform's technical-documentation tone, consistent with the sibling `harper-applications-in-depth.mdx` guide.

### DIFF
--- a/learn/developers/multiple-applications.mdx
+++ b/learn/developers/multiple-applications.mdx
@@ -1,0 +1,118 @@
+---
+title: Multiple Applications on One Cluster
+---
+
+# Multiple Applications on One Cluster
+
+A single Harper cluster can host any number of applications side by side. Because the database, cache, application logic, messaging, and search run within one distributed runtime, an application is not a group of containers wired together—it is a [component](./), a self-contained module that Harper loads and serves. You add an application to a cluster by registering another component, not by provisioning a new cluster.
+
+This guide covers how to run multiple applications on one cluster: how they coexist, how to isolate them where required, and how to deploy them across every node.
+
+## Applications are components
+
+The unit you deploy in Harper is a component. In the [Operations API](../operations-api/components) and CLI, "component" refers to an application—a collection of schemas, resources, routes, and static assets that Harper loads and serves.
+
+Two consequences follow:
+
+- **A cluster is a shared resource.** To run a new application, register another component on an existing cluster rather than creating a new cluster.
+- **Co-located applications share the runtime.** Because they run within the same converged process, one application's resources can access another application's tables through an in-process call—no cross-service HTTP and no separate connection pool.
+
+Co-located applications must be kept separate along three boundaries: **routes, data, and roles.** The sections below describe how to define each boundary.
+
+## Structuring each application
+
+Every Harper application is configured with a `config.yaml` file in the root of its directory. This file specifies which built-in extensions the application uses—`rest`, the `graphqlSchema` loader, custom JavaScript resources, Fastify routes, static file serving, and others:
+
+```yaml
+# listings-api/config.yaml
+rest: true
+graphqlSchema:
+  files: '*.graphql'
+jsResource:
+  files: 'resources.js'
+roles:
+  files: 'roles.yaml'
+fastifyRoutes:
+  files: 'routes/*.js'
+  urlPath: '/listings'
+static:
+  files: 'web/**'
+  urlPath: '/listings'
+```
+
+A `config.yaml` **completely replaces** the default configuration; it is not merged with Harper's defaults. Each application therefore defines its own surface area explicitly. For the full list of options, see the [Component Configuration](../../reference/components/configuration) and [Built-In Extensions](../../reference/components/built-in-extensions) reference pages.
+
+### Namespacing routes
+
+Applications on the same instance respond on the same HTTP port (`9926` by default), so route collisions must be designed around.
+
+For [Fastify routes](./define-routes) and static assets, set a unique `urlPath` prefix per application—for example `/listings`, `/inventory`, or `/admin`. Custom routes and web content are then kept separate.
+
+REST endpoint paths are derived from the resource and table names an application exports, not from `urlPath`. Give each application uniquely named resources or, preferably, its own database. Two applications that export a `User` resource into the same database will conflict; two applications with tables in separate databases will not.
+
+## Isolating data
+
+Isolate co-located applications through the database. Assign each application its own database so its tables belong to it. Those tables are queryable by that application's resources and remain hidden from other applications unless a query deliberately reaches across.
+
+When one application needs data from another—for example, an `admin-dashboard` application reading from the `listings-api` application—it queries the table directly, in-process, rather than opening a network connection to another service. Data boundaries are defined per application.
+
+## Configuring authentication per application
+
+Each application can include its own [`roles.yaml`](./defining-roles) file, defining the roles and permissions that control access to its resources. Co-located applications do not share a permission model: a public-facing listings API and an internal admin tool can enforce different access rules on the same cluster. Define each application's roles around its own data; the runtime enforces them independently.
+
+## Registering multiple applications
+
+There are two ways to add applications to an instance.
+
+### Declaratively, through the instance config
+
+Harper reads applications from the `harperdb-config.yaml` file in its root path (usually `~/hdb`). Each entry points to a package—a GitHub repository, an npm package, a tarball, a local path, or a URL:
+
+```yaml
+# ~/hdb/harperdb-config.yaml
+listings-api:
+  package: my-org/listings-api#v1.4.0
+inventory-service:
+  package: my-org/inventory-service#v2.1.0
+admin-dashboard:
+  package: my-org/admin-dashboard#v0.9.0
+```
+
+Reference a Git repository with a semver tag to lock each application to a specific, reproducible version. Harper translates these entries into a `package.json` file, runs an install to resolve them, and loads each as a component. The entry name is arbitrary and does not need to match a package dependency name.
+
+### Imperatively, through the CLI
+
+To deploy from an application directory, run `harperdb deploy` inside the project. This packages the current directory and sends it to the active instance:
+
+```bash
+cd listings-api
+harperdb deploy
+```
+
+For local iteration, `harperdb dev .` runs the application and watches for file changes, restarting worker threads on edit. See the [Applications reference](../../reference/components/applications) for the complete set of deployment options.
+
+## Deploying across the cluster
+
+The preceding sections deploy an application to a single instance. To deploy it across every node in the cluster, include the `replicated=true` parameter.
+
+When deploying in a clustered environment, set `replicated=true` to spread the deployment to all nodes:
+
+```bash
+harperdb deploy_component \
+  project=listings-api \
+  package=https://github.com/my-org/listings-api#v1.4.0 \
+  target=https://cluster-node-1.example.com:9925 \
+  replicated=true
+```
+
+`target` points to a node's operations endpoint. `replicated=true` sends the deployment to the rest of the cluster, extending the "deploy everywhere" behavior of Harper's [replication](../replication/) to the application itself.
+
+> Restart afterward to apply the changes—for example, `harperdb restart target=https://cluster-node-1.example.com:9925 replicated=true`.
+
+On Harper Fabric, a single `harperdb deploy` from the project directory deploys the application across regions, with replication, routing, and failover managed by the platform.
+
+## When to co-locate
+
+Co-locate applications on one cluster when they share a data domain, benefit from in-process access to each other's tables, or are small enough that separate clusters would sit mostly idle. This covers common cases such as an API, its admin interface, a background worker, and a public site.
+
+Use separate clusters when applications require independent scaling, have tenancy or compliance requirements that mandate separation, or have significantly different availability needs. Choose based on the workload rather than defaulting to one cluster per service.
--- a/learn/developers/multiple-applications.mdx
+++ b/learn/developers/multiple-applications.mdx
@@ -4,13 +4,13 @@ title: Multiple Applications on One Cluster
 
 # Multiple Applications on One Cluster
 
-A single Harper cluster can host any number of applications side by side. Because the database, cache, application logic, messaging, and search run within one distributed runtime, an application is not a group of containers wired together—it is a [component](./), a self-contained module that Harper loads and serves. You add an application to a cluster by registering another component, not by provisioning a new cluster.
+A single Harper cluster can host any number of applications side by side. Because the database, cache, application logic, messaging, and search run within one distributed runtime, an application is not a group of containers wired together—it is a [component](/reference/v5/components/overview), a self-contained module that Harper loads and serves. You add an application to a cluster by registering another component, not by provisioning a new cluster.
 
 This guide covers how to run multiple applications on one cluster: how they coexist, how to isolate them where required, and how to deploy them across every node.
 
 ## Applications are components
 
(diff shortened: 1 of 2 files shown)


kriszyp (changes requested):
Thanks for putting this together — the mechanics that are here (config.yaml precedence, ports, `deploy_component replicated=true`) are accurate, and the broken cross-references from the first pass are fixed.

Requesting changes on **scope/completeness**, though — for a "multiple applications on one cluster" guide, the routing story is the crux and it's currently thin:

**Significant — routing is incomplete.** The guide only covers static assets / Fastify. It doesn't explain **middleware routing by URL path**, which is the primary mechanism for fronting multiple apps on one instance (how a request path maps to a given application). That should be the backbone of this guide, not an aside.

**Significant — virtual hosts are missing.** Hostname-based routing (virtual hosts) is arguably *the* feature for hosting multiple apps on one cluster — routing by host, not just path. A multi-app guide that omits it leaves out the main pattern users will reach for.

**Significant — module loader context isolation.** When several apps run in one instance, how their module loader contexts are isolated (or shared) is a real correctness/operational concern and belongs in this guide — at minimum a section explaining the isolation model so users understand what is and isn't shared between co-located apps.

Happy to point at the canonical references for each of these if that helps. The foundation is good; it just needs the routing + isolation story filled in before it's the guide people should follow for this topic.

— review via KrAIs (Claude) on Kris's behalf

kriszyp (changes requested):
Requesting changes, but narrower than the automated review suggested — checked the two API-shape claims against Harper's actual `Request` implementation before posting, and they don't hold up:

- `request.host` and `request.pathname` **are** real, documented Harper `Request` properties (`server/serverHelpers/Request.ts`), not something that needs `request.headers.get(...)` instead — no change needed there.
- Marking the handlers `async` isn't required for crash-safety either: `requestHandler` in `server/http.ts` already wraps the whole chain call in a `try`/`catch` (`await httpChain[port](request)`), which catches a synchronous throw from a non-async middleware just as reliably as an async one.

The one thing genuinely worth fixing: the first example's prose says "dispatch on a custom header" but the code shows a path-prefix check (`request.pathname.startsWith('/listings')`) — text and code don't match each other. Worth making the example match the text (a header check, since that's the case path-prefix routing doesn't already cover) so the doc isn't demonstrating the wrong tool for the job it describes. Not a correctness or performance bug, just an internal consistency fix.

kriszyp (changes requested):
Thanks for adding the hostname-routing section. One significant correction is still needed: the current example manually branches on `request.host`, but Harper 5.2 has first-class virtual-host routing.

Components can declare `host` in `config.yaml`; the scoped `server` API automatically carries that value into the component’s HTTP, WebSocket, and upgrade registrations. Custom code can register the same routed chain directly with `server.http(handler, { host: "admin.example.com" })`. Harper then matches the request `Host` header and selects the corresponding isolated middleware chain.

Please make that declarative `host` configuration the primary application-routing pattern, show the direct `{ host }` option as the programmatic equivalent, and reserve branching on `request.host` for genuinely custom dispatch. The SNI material should remain separate: SNI selects the certificate, while `host` selects the application middleware chain.

I opened [Document middleware host and path routing (#595)](https://github.com/HarperFast/documentation/pull/595) to add the missing canonical references for both `host` and `urlPath`; this guide can link to those sections once it lands.

— review via KrAIs (GPT-5) on Kris’s behalf

learn/developers/multiple-applications.mdx
```diff
@@ -0,0 +1,173 @@
...
+
+`runFirst: true` inserts the handler at the front of the chain. Reach for the middleware chain when path prefixes and resource naming are not enough—for example, to dispatch on a custom header, as above, or to rewrite a path before it reaches an application. See the [HTTP API](/reference/v5/http/api) reference for the full `Request` object.
+
+### Host-based routing
+
+To route by hostname—serving `listings.example.com` and `admin.example.com` from the same instance—declare `host` in the application's `config.yaml`, the same way you declare `urlPath`:
+
+```yaml
+# admin-dashboard/config.yaml
+host: admin.example.com
```
kriszyp, line 89: Could we move `host` under `fastifyRoutes` here? Harper reads `host` from each plugin's config and the scoped `server` proxy applies it to registrations made by that plugin; a bare top-level `host` entry does not scope the whole application. As written, these Fastify routes remain reachable on every host. If `rest` or `static` should share the hostname, configure those plugin entries too. This is new in v5.2.0, so please also add a `<VersionBadge version="v5.2.0" />` to this section and ensure the companion HTTP reference update lands first.

learn/developers/multiple-applications.mdx
```diff
@@ -0,0 +1,173 @@
...
+
+## Isolating data
+
+Isolate co-located applications through the database. Assign each application its own database so its tables belong to it. Those tables are queryable by that application's resources and remain hidden from other applications unless a query deliberately reaches across.
+
+When one application needs data from another—for example, an `admin-dashboard` application reading from the `listings-api` application—it queries the table directly, in-process, rather than opening a network connection to another service. Data boundaries are defined per application.
+
+## Configuring authentication per application
+
+Each application can include its own [`roles.yaml`](/reference/v5/users-and-roles/configuration) file, defining the roles and permissions that control access to its resources. Co-located applications do not share a permission model: a public-facing listings API and an internal admin tool can enforce different access rules on the same cluster. Define each application's roles around its own data; the runtime enforces them independently.
```
kriszyp, line 114: Roles are not enforced independently per application. Every `roles.yaml` reconciles role names into the instance-wide `system.hdb_role` table, and users/sessions share that RBAC system. Two apps defining the same role name can update the same role. Please describe unique, application-prefixed role names with permissions scoped to each database, while making the shared permission model explicit.

learn/developers/multiple-applications.mdx
```diff
@@ -0,0 +1,173 @@
...
+server.http(handleAdmin, { host: 'admin.example.com' });
+```
+
+Reach for that direct form, or for branching on `request.host` inside a handler, only when the routing decision needs logic beyond a fixed host—for example, resolving the target host at runtime instead of at deploy time.
+
+When serving multiple domains over HTTPS, configure a certificate per domain with [SNI](/reference/v5/http/tls#multi-domain-certificates-sni): define `tls` as an array with a `host` entry for each domain. SNI selects the certificate; `host` selects the application's middleware chain—the two are independent.
+
+## Isolating data
+
+Isolate co-located applications through the database. Assign each application its own database so its tables belong to it. Those tables are queryable by that application's resources and remain hidden from other applications unless a query deliberately reaches across.
```
kriszyp, line 108: A separate database provides namespacing, not enforced isolation: every co-located component can access it through the shared `databases` API. Could we rename this section/boundary accordingly and explicitly say that buggy or untrusted component code is not prevented from crossing it? Security or compliance isolation requires a separate instance/cluster.

learn/developers/multiple-applications.mdx
```diff
@@ -0,0 +1,173 @@
...
+
+### Path-based routing
+
+Path prefixes are the primary way to keep applications separate. Each built-in extension that serves HTTP—`fastifyRoutes`, `static`, and others—accepts a `urlPath` that mounts its files under a URL prefix. Give each application a unique prefix such as `/listings`, `/inventory`, or `/admin`, as in the `config.yaml` above.
+
+When `urlPath` is `.` or begins with `./`, Harper prepends the component name automatically, so an application named `listings-api` mounts under `/listings-api` by default. Set an explicit prefix to control the path yourself. You can also override the mount path at deploy time with the `urlPath` parameter to `deploy_component` (for example, `urlPath=/api/v2`).
+
+REST endpoint paths behave differently: they are derived from the resource and table names an application exports, not from `urlPath`. Give each application uniquely named resources or, preferably, its own database. Two applications that export a `User` resource into the same database will conflict; two applications with tables in separate databases will not.
```
kriszyp, line 65: This is stale for v5.2: `rest` can carry `urlPath` (and `host`) in its plugin config, and Harper mounts the REST middleware on that routed chain. Resource registration remains shared, so duplicate resource keys may still conflict, but the external REST path can be namespaced. Please distinguish those two concerns.

kriszyp (changes requested):
Can you update this for https://github.com/HarperFast/documentation/pull/612 ?

The overall guide is useful, but the host-routing example is configured at the wrong level and would leave the advertised application routes reachable on every host.
The data and authentication sections also describe per-application isolation that Harper does not enforce: databases are shared namespaces, and users/roles are instance-wide.


learn/developers/multiple-applications.mdx
BboyAkers: Fixed, and reframed around #612 since that landed after this comment.

Path routing is now led by the application mount on the root-config entry (`host`/`urlPath` on the `harper-config.yaml` application entry), with a plugin's own `urlPath` positioning it *within* the app and composing with the mount. The two concerns you flagged are now split explicitly in a new **What a mount does not do** section: a mount namespaces the external URL an application answers on, but it is not a resource namespace — exports land in one instance-wide registry, so two apps that both export `User` still conflict wherever each is mounted.

learn/developers/multiple-applications.mdx
BboyAkers: Fixed — though via #612 rather than by moving `host` under `fastifyRoutes`, since that merged after this comment and moved the mount to the root config.

- `host`/`urlPath` are now declared on the application's entry in the root `harper-config.yaml` (deployment concern, remappable per environment), with `deploy_component host=... urlPath=...` as the deploy-time equivalent and `server.http(handler, { host, urlPath })` as the programmatic one. Branching on `request.host` is now explicitly reserved for genuinely custom dispatch.
- The `fastifyRoutes` trap you were pointing at is now documented rather than worked around: `fastifyRoutes` registers as a global fallback outside the routed chain, so a `urlPath` mount applies as the route prefix but a `host` mount does not — and Harper refuses to load a host-mounted app that declares it. I also dropped `fastifyRoutes` from the guide's primary `config.yaml` example (it's discouraged in the reference anyway) and point at custom resources / `server.http()` instead.
- Added `<VersionBadge version="v5.2.0" />` to the mounting section, and the guide now links the merged #595/#612 reference sections (`http/overview#middleware-routing`, `HttpOptions`, `deploy_component`). Branch is merged up with main so those anchors resolve; `npm run build` is clean.

learn/developers/multiple-applications.mdx
BboyAkers: Fixed. Section is renamed **Namespacing data by database**, and the boundary is now described as what it is: a namespace, not enforced isolation. It states plainly that `databases` is a process-wide object, that every co-located component can reach every database through it, and that nothing in the runtime prevents buggy or untrusted component code from crossing the boundary — co-located applications share one trust domain. Security or compliance isolation now explicitly requires a separate instance or cluster, and **When to co-locate** carries the same framing (untrusted/third-party components moved to the separate-cluster column).

learn/developers/multiple-applications.mdx
BboyAkers: Fixed. The section is now **Access control is instance-wide** and drops the per-application enforcement claim entirely. It says that every `roles.yaml` reconciles into the instance's single role registry (a missing role is created, an existing one has its permissions overwritten), and that users and sessions authenticate against the instance as a whole. Guidance is the two practices you described: prefix role names per application (two apps declaring `reader` are one role, last load wins, silently changing what the other expects) and scope each role's permissions to that application's database. Closes with the explicit statement that 'different access rules per application' means different roles in one shared RBAC system, and that a truly separate user directory or role namespace needs a separate instance.

BboyAkers:
Rebased-forward (merged `main`) and pushed a revision addressing all four open threads, updated for #612 as requested — that merged after the last review round and moved application mounting from the component's `config.yaml` to the application's root-config entry, so the host-routing section was rewritten around that rather than by moving `host` under `fastifyRoutes`.

**Routing (was configured at the wrong level)**
- The root `harper-config.yaml` application entry is now the primary mount: `host` + `urlPath` per application, with `deploy_component host=… urlPath=…` as the deploy-time equivalent and `server.http(handler, { host, urlPath })` as the programmatic one. Branching on `request.host` is reserved for genuinely custom dispatch (resolving a target host at runtime).
- New **What a mount does not do** section: a mount is not a resource namespace (exports land in one instance-wide registry, so duplicate exported names still conflict), and it does not host-constrain `fastifyRoutes` — those register as a global fallback outside the routed chain, a `urlPath` mount applies but a `host` mount does not, and Harper refuses to load a host-mounted app that declares them. `fastifyRoutes` is out of the primary `config.yaml` example; a plugin's own `urlPath` is now framed as positioning it *within* the app and composing with the mount.
- `<VersionBadge version="v5.2.0" />` on the mounting section, and SNI stays separate (certificate selection vs. middleware-chain selection).

**Data (described isolation Harper does not enforce)**
- Section renamed **Namespacing data by database**: a database is a namespace, `databases` is process-wide, every co-located component can reach every database through it, and buggy or untrusted component code is not prevented from crossing. Co-located apps share one trust domain; security/compliance isolation needs a separate instance or cluster. **When to co-locate** now reflects that too.

**Roles (described per-application enforcement)**
- Section renamed **Access control is instance-wide**: every `roles.yaml` reconciles into the instance's single role registry (create-or-overwrite), users and sessions are instance-wide. Guidance is application-prefixed role names (two apps declaring `reader` are one role, last load wins) with permissions scoped to each application's database, and an explicit note that per-app access rules mean different roles inside one shared RBAC system.

Also swapped the deprecated `runFirst` in the middleware example for the `name`/`before`/`after` ordering guidance. `prettier --check` and `npm run build` are clean (no broken links or anchors — the #595/#612 reference anchors resolve now that main is merged in).

learn/developers/multiple-applications.mdx
```diff
@@ -0,0 +1,212 @@
...
+
+Because the permission model is shared, "different access rules per application" means different roles within one RBAC system, not separate systems. If two applications must not share a user directory or a role namespace at all, run them on separate instances.
+
+## Registering multiple applications
+
+There are two ways to add applications to an instance.
+
+### Declaratively, through the instance config
+
+Harper reads applications from the `harper-config.yaml` file in its root path (usually `~/hdb`). Each entry points to a package—a GitHub repository, an npm package, a tarball, a local path, or a URL—and, as shown in [Mounting an application](#mounting-an-application), carries that application's `host` and `urlPath`:
```
kriszyp: Could this say that an entry *may* point to a package? The guide correctly says a root-config entry does not need `package` for an already deployed component at line 78, and line 182 relies on that to mount a payload-deployed application. Saying every entry points to a package makes that valid route-only configuration sound invalid.

kriszyp (approved):
Looks good

learn/developers/multiple-applications.mdx
BboyAkers: Fixed. The sentence now says an entry *may* point to a package and *may* carry `host`/`urlPath`, and spells out that either half stands on its own: a `package` with no routing keys installs an application served on the default chain, and routing keys with no `package` mount a component that was deployed some other way (e.g. `harper deploy`). That keeps it consistent with the route-only configuration described at the mounting section and used for the payload-deploy case below.

Ethan-Arrowood (approved):

PR #588, opened 2026-07-15
docs(sql): 5.2 SQL engine performance guidance

kriszyp:
Adds guidance to the SQL operations-api reference for the new Resource-API SQL engine landing in 5.2 ([harper #1285 — New SQL engine on the Resource API](https://github.com/HarperFast/harper/pull/1285)), per Kris's request to document which SQL is well supported / runs efficiently and what to avoid.

## Changes to `reference/operations-api/sql.md`

- **Replaced the blanket "SQL is not recommended for production" warning** with an accurate 5.2 note: the new engine serves index-plannable queries as streaming, bounded-memory operations and transparently falls back to the legacy path for anything it can't plan. REST is still steered as the choice for the hottest production read paths.
- **New `## Query Performance` section:**
  - **Index-served (efficient):** PK lookups, equality/`IN` on indexed attributes, range/`BETWEEN`, prefix `LIKE 'x%'`, index-order `ORDER BY` with `LIMIT`/`OFFSET` pushdown, `GROUP BY` on an indexed attribute, `INNER`/`LEFT JOIN` on an indexed inner key, projection pushdown.
  - **Full-scan fallback (avoid on large tables):** un-indexed predicates, cross-attribute `OR`, suffix/substring `LIKE`, `!=`/`NOT` negations, un-driven `ORDER BY`, `SEARCH_JSON`; plus a note that `RIGHT`/`FULL JOIN`, no-`WHERE` UPDATE/DELETE, and `COUNT(DISTINCT)` run on the legacy engine.
  - **The fallback contract:** `auto` default, results never change (only the performance profile), and the `sql-engine v2 fallback:` log signal.
  - **Practical guidance:** index what you filter/sort/join on, prefer prefix `LIKE`, constrain before sort/group/join/`SEARCH_JSON`, use REST for hot paths.

## Accuracy notes

- Guidance is grounded in the engine's optimizer rules (predicate normalization for `LIKE`/`BETWEEN`, sort-from-index, validate-scannable) and the phase-5 cutover parity work in #1285.
- Reconciled against the existing Features Matrix: `UNION`, sub-`SELECT`, and `INSERT … SELECT` are unsupported by *both* engines, so they are called out as "not supported" rather than "slow fallback."
- Verified the `RIGHT`/`FULL JOIN` fallback claim against source: `normalizer.ts` throws `EngineUnsupportedError` for both, and the router catches exactly that under `auto` to fall back to legacy (which supports them).

Note: #1285 is still open and targeted at 5.2; the "As of Harper 5.2" framing lines up with that release. If the merge/version slips, the version reference should move with it.

### DIFF
--- a/reference/operations-api/sql.md
+++ b/reference/operations-api/sql.md
@@ -118,8 +118,8 @@ These map directly to indexed storage operations:
 - **Range predicates on an indexed attribute** — `<`, `<=`, `>`, `>=`, and `BETWEEN` (compiled to a bounded index range).
 - **`OR` of indexed predicates** — `WHERE a = 1 OR b = 2` is served as a union of index scans, provided **every** branch is an index-driving predicate on an indexed attribute.
 - **Prefix `LIKE`** — `WHERE name LIKE 'Har%'` compiles to an indexed `starts_with` scan.
-- **`ORDER BY` on an indexed attribute** — served directly from index order, with no separate in-memory sort; `LIMIT`/`OFFSET` push into the scan so only the rows you return are read.
-- **`GROUP BY` on an indexed attribute** — aggregates stream group-by-group with O(1) memory per group.
+- **`ORDER BY` on an indexed attribute, combined with an indexed `WHERE`** — served directly from index order, with no separate in-memory sort; `LIMIT`/`OFFSET` push into the scan so only the rows you return are read. Without an index-driving `WHERE` condition, an `ORDER BY` currently runs on the legacy path.
+- **`GROUP BY` on an indexed attribute, combined with an indexed `WHERE`** — aggregates stream group-by-group with O(1) memory per group (memory is bounded by the group count; every matching row is still read). Without an index-driving `WHERE` condition, a `GROUP BY` currently runs on the legacy path.
 - **`INNER`/`LEFT JOIN` on an indexed join key** — executed as an index-nested-loop join (stream the outer table, probe the inner by index). Keep the join key indexed on the inner (probed) table.
 - **Column projections** — selecting a subset of columns pushes down so only those attributes are read.
 
@@ -139,7 +139,7 @@ Note that `UNION`, sub-`SELECT`, and `INSERT … SELECT` are not supported by ei
 
 ### The fallback contract
 
-The engine runs in `auto` mode by default: it executes every query it can plan on the new engine and transparently falls back to the legacy engine for anything it can't. A query that falls back returns exactly what the legacy engine would — **its results don't change, only its performance profile.** Fallbacks are logged (`sql-engine v2 fallback: ...`) with the reason and the offending column, which is a useful signal for finding queries worth an added index or a reshape. (For queries the new engine _does_ plan, it corrects a few long-standing legacy quirks — see [Behavior changes from the legacy engine](#behavior-changes-from-the-legacy-engine).)
+The engine runs in `auto` mode by default: it executes every query it can plan on the new engine and transparently falls back to the legacy engine for anything it can't. A query that falls back returns exactly what the legacy engine would — **its results don't change, only its performance profile.** Fallbacks are logged (`SQL engine v2 fallback: ...`) with the reason, which is a useful signal for finding queries worth an added index or a reshape. (For queries the new engine _does_ plan, it corrects a few long-standing legacy quirks — see [Behavior changes from the legacy engine](#behavior-changes-from-the-legacy-engine).)
 
 ### Behavior changes from the legacy engine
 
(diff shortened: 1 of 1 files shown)


reference/operations-api/sql.md
```diff
@@ -11,8 +11,10 @@ title: SQL
 <!-- Source: versioned_docs/version-4.7/reference/sql-guide/reserved-word.md -->
 <!-- Source: versioned_docs/version-4.7/developers/operations-api/sql-operations.md -->
 
-:::warning
-SQL querying is not recommended for production use or on large tables. SQL queries often do not utilize indexes and are not optimized for performance. Use the [REST interface](../rest/overview.md) for production data access — it provides a more stable, secure, and performant interface. SQL is intended for ad-hoc data investigation and administrative queries.
+:::info Harper 5.2 introduces a new SQL engine
+As of Harper 5.2, SQL runs on a new engine built directly on Harper's indexed storage and Resource API. Queries that can be planned against an index now execute as streaming, index-driven operations with memory bounded by the size of the result — not the size of the table. Queries the engine can't plan against an index automatically fall back to the legacy execution path, returning the same results but with the older full-scan cost. See [Query Performance](#query-performance) for how to keep your queries on the fast path.
```
kylebernhardy: This is cool, how do users get observability into the ability to plan?

reference/operations-api/sql.md
```diff
@@ -97,6 +99,57 @@ ORDER BY d.dog_name
 
 ---
 
+## Query Performance
+
+Harper 5.2's SQL engine maps SQL clauses onto Harper's native indexed storage. When a query can be planned against an index, it runs as a streaming, index-driven operation with memory bounded by the size of the result — not the size of the table. When it can't, the engine falls back to the legacy path, which fetches candidate rows and evaluates the query in memory, with cost and memory growing with the number of rows scanned.
```
kylebernhardy, line 104: There should be some form of larger callout here that the fallback incurs full table scans which is problematic for large tables.

reference/operations-api/sql.md
```diff
@@ -97,6 +99,57 @@ ORDER BY d.dog_name
...
+- **`ORDER BY` on an indexed attribute** — served directly from index order, with no separate in-memory sort; `LIMIT`/`OFFSET` push into the scan so only the rows you return are read.
+- **`GROUP BY` on an indexed attribute** — aggregates stream group-by-group with O(1) memory per group.
+- **`INNER`/`LEFT JOIN` on an indexed join key** — executed as an index-nested-loop join (stream the outer table, probe the inner by index). Keep the join key indexed on the inner (probed) table.
+- **Column projections** — selecting a subset of columns pushes down so only those attributes are read.
+
+### Queries that fall back to a full scan (avoid on large tables)
+
+These can't be served from an index. They still return correct results, but the engine falls back to the legacy full-scan path, so cost and memory grow with table size:
+
+- **Filters on un-indexed attributes** — any equality, range, or `IN` on an attribute with no index.
```
kylebernhardy: What if there is another indexed filter in an AND?

reference/operations-api/sql.md
```diff
@@ -97,6 +99,57 @@ ORDER BY d.dog_name
...
+
+Note that `UNION`, sub-`SELECT`, and `INSERT … SELECT` are not supported by either engine — see the [Features Matrix](#features-matrix).
+
+### The fallback contract
+
+The engine runs in `auto` mode by default: it executes every query it can plan on the new engine and transparently falls back to the legacy engine for anything it can't, so **no query changes its results** — only its performance profile. Fallbacks are logged (`sql-engine v2 fallback: ...`) with the reason and the offending column, which is a useful signal for finding queries worth an added index or a reshape.
+
+### Practical guidance
+
+- Index every attribute you filter, sort, or join on.
```
kylebernhardy: Is it worth calling out performance and storage implications of over indexing?

reference/operations-api/sql.md
```diff
@@ -97,6 +99,69 @@ ORDER BY d.dog_name
...
+### Queries that run efficiently (index-served)
+
+These map directly to indexed storage operations:
+
+- **Primary-key lookups** — `WHERE id = ...`, `WHERE id IN (...)`.
+- **Equality and `IN` on an indexed attribute** — `WHERE status = 'active'`, `WHERE breed_id IN (1, 2, 3)`.
+- **Range predicates on an indexed attribute** — `<`, `<=`, `>`, `>=`, and `BETWEEN` (compiled to a bounded index range).
+- **`OR` of indexed predicates** — `WHERE a = 1 OR b = 2` is served as a union of index scans, provided **every** branch is an index-driving predicate on an indexed attribute.
+- **Prefix `LIKE`** — `WHERE name LIKE 'Har%'` compiles to an indexed `starts_with` scan.
+- **`ORDER BY` on an indexed attribute, combined with an indexed `WHERE`** — served directly from index order, with no separate in-memory sort; `LIMIT`/`OFFSET` push into the scan so only the rows you return are read. Without an index-driving `WHERE` condition, an `ORDER BY` currently runs on the legacy path.
```
cb1kenobi: **Medium: `ORDER BY` WHERE-requirement is overstated — misses the primary-key exception**

`validateScannable` (`sqlEngine/optimizer/rules/validateScannable.ts`) accepts a scan with **no** index-driving `WHERE` when the pushed sort drives the scan via `sortDrivesIndex` in `sqlEngine/optimizer/whereToConditions.ts` — but that check is restricted to the table's **primary key**:

```ts
return attr?.indexed === true && attr.isPrimaryKey === true;
```

Concretely, `SELECT ... FROM t ORDER BY id LIMIT 10` (no `WHERE`, `id` = PK) is served on the new engine's index-order scan — it does not fall back to legacy. This is a deliberate, tested exception (comment references design decision "D-219"), confirmed by `unitTests/sqlEngine/pipeline.test.js`:
- `'serves a no-WHERE ORDER BY on the PRIMARY KEY via index order (D-219)'` — passes on the new engine.
- `'rejects a no-WHERE ORDER BY on a NON-PK indexed attribute (index order != legacy sort)'` — this one does fall back.

The current blanket statement ("Without an index-driving `WHERE` condition, an `ORDER BY` currently runs on the legacy path") will mislead readers doing PK-based pagination — a common pattern — into adding an unneeded `WHERE` or assuming the fast path isn't already available.

(The parallel `GROUP BY` sentence on the next line is accurate as written — there's no equivalent drives-without-`WHERE` exception for `GROUP BY` in the optimizer.)

Suggested fix:
```suggestion
- **`ORDER BY` on an indexed attribute, combined with an indexed `WHERE`** — served directly from index order, with no separate in-memory sort; `LIMIT`/`OFFSET` push into the scan so only the rows you return are read. Without an index-driving `WHERE` condition, an `ORDER BY` on a *secondary* indexed attribute runs on the legacy path — except when sorting on the table's **primary key**, which drives the scan from the primary key's natural index order even with no `WHERE` clause.
```

—
_Generated by Barber AI_

cb1kenobi (approved):

Ethan-Arrowood (approved):
LGTM. I checked the performance claims against the engine source rather than just the prose — the index-served vs full-scan taxonomy, the OR-taint rule, the PK-sort exception, and the `auto` default with legacy fallback all match the optimizer/router behavior, and there's no contradiction with the #1844 PK-sort or #1845 A/B benchmark work. Good to land in lockstep with 5.2.

PR #590, opened 2026-07-16
Document RocksDB backup & restore operations

cb1kenobi:
## Summary

Documents the RocksDB backup & restore feature shipping in 5.2 (companion to HarperFast/harper#1831), in a dedicated **Backups** reference section:

- **Backups / Overview** (`reference/backups/overview.md`) — how the backup system works (incremental managed backups, snapshot download), its limitations (engine support, granularity, single storage root, online-vs-offline restore matrix), and two worked examples: incremental backup/restore with `harper create_backup` / `harper restore_backup`, and `harper get_backup` streaming a `tar.gz` off-host plus the manual restore procedure (stop Harper, extract into the database directory under `storage.path`).
- **Backups / Operations** (`reference/backups/operations.md`) — all seven operations (`create_backup`, `list_backups`, `verify_backup`, `delete_backup`, `purge_backups`, `restore_backup`, `get_backup`) with Operations API and CLI examples.
- **EngineBadge component** (`src/components/EngineBadge.tsx`, globally registered like `VersionBadge`) tags storage-engine support per operation: the managed operations are RocksDB-only; `get_backup` supports both RocksDB and LMDB.
- **Operations API** (`reference/operations-api/operations.md`) — Backup & Restore trimmed to the brief index pattern used by Logs/Certificate Management, linking to the new section.
- **CLI** (`reference/cli/commands.md`) — the Backup Commands section is removed (content lives in Backups now); the backup operations are listed in the CLI operations table (`reference/cli/operations-api-commands.md`).
- **Configuration** (`reference/configuration/options.md`) — the `backupPath` storage option (default `<rootPath>/backup`).
- **Release notes** (`release-notes/v5-lincoln/5.2.md`) — Backup & Restore entry, linking to the new section.

## Where to look

- The manual-restore example in the Overview instructs extracting the archive into `<storage.path>/<database>` (default `~/hdb/database/data`) — path verified against `DATABASES_DIR_NAME` in core and rocksdb-js's tar layout (entries unpack directly into a directory that opens as a RocksDB database, including `transaction_logs/`). Worth a second pair of eyes since a wrong path here damages user data.
- Cross-model review (Codex) surfaced four accuracy issues, all fixed in this branch: the whole-database claim now carves out LMDB `get_backup` partial exports, the single-root-store restriction is documented as a limitation, the `super_user` requirement is scoped to server-invoked operations (offline CLI is filesystem-governed), and LMDB manual restore points at `storage.path` rather than the default-only path. The Gemini review leg failed to run (headless permission denial), so outside-model coverage is Codex-only.
- EngineBadge prop shape is a comma-separated string (`engines="RocksDB, LMDB"`) rendered in the same style as VersionBadge — open to a different design.

Reviewers not assigned — leaving that to you.


reference/operations-api/operations.md
```diff
@@ -133,9 +132,114 @@ Drops an attribute and all its values from the specified table.
...
+
+Every managed backup and every RocksDB `get_backup` includes the audit/transaction log, so a restored database keeps its `read_audit_log` history as of the backup point. A restore is a point-in-time rollback; in a replicated cluster, coordinate a restore with replication before bringing the node back.
+
+| Operation        | Description                                                         | Role Required |
+| ---------------- | ------------------------------------------------------------------- | ------------- |
+| `create_backup`  | Creates a managed, incremental directory backup of a database (job) | super_user    |
+| `list_backups`   | Lists the managed backups for a database                            | super_user    |
+| `verify_backup`  | Verifies a managed backup's integrity (job)                         | super_user    |
+| `delete_backup`  | Deletes a single managed backup                                     | super_user    |
+| `purge_backups`  | Deletes all but the newest N managed backups                        | super_user    |
```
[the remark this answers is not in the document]
cb1kenobi: Fixed in f7211a60: the operations table now reads "newest `keep_count` managed backups".

reference/cli/commands.md
```diff
@@ -294,6 +294,73 @@ Database files are stored in the `hdb/database` directory. As long as the snapsh
...
+harper get_backup database=data target=https://node-2.example.com:9925 out=./data.tar.gz
+```
+
+### `harper restore_backup`
+
+Restore a database from a managed backup. Pass `backup_id=<id>` to choose a backup (defaults to the latest), or `target_database=<name>` to restore into a new database rather than in place. Run it with the server **stopped** to restore the `system` database, any database a loaded component keeps open, or into a `target_database`; while Harper is running, those cases are rejected with a pointer to stop the server.
+
+```bash
+harper restore_backup database=data backup_id=1
+```
```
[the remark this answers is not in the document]
cb1kenobi: Added in f7211a60: a `restore_backup … target_database=data_restored` CLI example under `harper restore_backup`.

reference/operations-api/operations.md
```diff
@@ -133,9 +132,114 @@ Drops an attribute and all its values from the specified table.
...
+| `restore_backup` | Restores a database from a managed backup (job)                     | super_user    |
+| `get_backup`     | Streams a full snapshot of a database in the response for download  | super_user    |
+
+### `create_backup`
+
+<VersionBadge version="v5.2.0" />
+
+Creates an incremental directory backup of the database (RocksDB only) under the configured backup root. Runs as a background [job](#jobs): the operation returns a `job_id` immediately, and [`get_job`](#get_job) reports the outcome including the new `backupId`, `size`, and `timestamp`. `database` defaults to `data`.
```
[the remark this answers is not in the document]
cb1kenobi: Fixed at the source: harper#1831 now returns snake_case, so the response field is `backup_id`. Docs updated to match.

reference/operations-api/operations.md
```diff
@@ -133,9 +132,114 @@ Drops an attribute and all its values from the specified table.
...
+
+```json
+{ "operation": "create_backup", "database": "dev" }
+```
+
+### `list_backups`
+
+<VersionBadge version="v5.2.0" />
+
+Returns the managed backups for a database (RocksDB only), each with its `backupId`, `timestamp`, `size`, and file count. Returns an empty array if no backups have been created yet.
```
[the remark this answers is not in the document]
cb1kenobi: Fixed at the source: the list entries now expose `backup_id`, `timestamp`, `size`, and `file_count` (mapped from the binding's camelCase; internal `appMetadata` dropped). Docs updated.

reference/backups/overview.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Limitations
+
+- **Managed backups require the RocksDB storage engine.** For LMDB databases, use [`get_backup`](./operations.md#get_backup) or [volume snapshots](../cli/commands.md#backing-up-with-volume-snapshots).
+- **Whole-database granularity.** There is no per-table backup or restore. (The one exception: `get_backup` on an LMDB database can stream individual tables, and includes the audit store only with `include_audit`.)
+- **One storage root per database.** A database whose tables use per-table `path` storage configs spans multiple root stores and cannot be backed up with these operations.
+- **Backups live on the node that created them.** The backup repository is a local directory. For disaster recovery, copy backup directories off-host, or use `get_backup` to pull snapshots from a running server.
+- **`get_backup` always streams the current state.** It cannot download a historical managed backup; to move a retained backup off-host, copy its backup directory.
+- **A restore is a point-in-time rollback.** In a replicated cluster, coordinate a restore with replication before bringing the node back.
```
cb1kenobi, line 40: **Medium: no documented recovery step for a crashed/interrupted restore**

`restore_backup` is destructive (purge-then-copy). If it's interrupted partway — process crash, power loss, disk error — the implementation leaves a `.restoring` marker next to the database directory and refuses to load that database on the next `harper start` (logged as an error: "Incomplete restore ... not loading it — rerun the restore to recover"). None of this is mentioned here or in `operations.md`: a reader who hits this after an interrupted restore has no way to know from the docs that the fix is simply to rerun `restore_backup` (same `backup_id`), not that data is lost or the install is broken.

Suggested fix: add a bullet to Limitations (or a short note under "When can a database be restored?"), e.g. "If `restore_backup` is interrupted before completing (crash, power loss), the affected database is left unloadable — Harper logs an incomplete-restore error and skips it on startup. Rerun `restore_backup` for the same database (and `backup_id`, if you passed one) to recover; do not attempt to load or manually repair the directory."

—
_Generated by Barber AI_

reference/backups/operations.md
```diff
@@ -0,0 +1,140 @@
...
+
+## `restore_backup`
+
+<VersionBadge version="v5.2.0" /> <EngineBadge engines="RocksDB" />
+
+Restores a database in place from a managed backup. `backup_id` defaults to the latest backup. The audit/transaction log is restored alongside the data.
+
+Through a running server this runs as a background [job](../operations-api/operations.md#jobs): Harper closes the database across all worker threads, restores it, and reloads it. This works only when no loaded component is holding the database open — restoring the `system` database, or a database a component keeps open, requires running the command from the CLI with the server stopped. See [when can a database be restored?](./overview.md#when-can-a-database-be-restored)
+
+From the CLI with the server stopped, `target_database=<name>` restores into a new database instead of overwriting the source. The target must not already exist; Harper picks the new database up on the next start.
```
cb1kenobi: **Low: `target_database` constraint is stated more strictly than the implementation**

"The target must not already exist" isn't quite what `restoreBackupOffline` enforces — it accepts an existing **empty** directory too (`isMissingOrEmptyDir`), rejecting only a non-empty one. Not harmful (following the doc's advice always works), but a reader with an empty leftover directory under that name would incorrectly conclude they need to pick a different name.

Suggested fix: "The target must not already exist (or must be empty); Harper picks the new database up on the next start."

—
_Generated by Barber AI_

reference/backups/overview.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Limitations
+
+- **Managed backups require the RocksDB storage engine.** For LMDB databases, use [`get_backup`](./operations.md#get_backup) or [volume snapshots](../cli/commands.md#backing-up-with-volume-snapshots).
+- **Whole-database granularity.** There is no per-table backup or restore. (The one exception: `get_backup` on an LMDB database can stream individual tables, and includes the audit store only with `include_audit`.)
+- **One storage root per database.** A database whose tables use per-table `path` storage configs spans multiple root stores and cannot be backed up with these operations.
+- **Backups live on the node that created them.** The backup repository is a local directory. For disaster recovery, copy backup directories off-host, or use `get_backup` to pull snapshots from a running server.
+- **`get_backup` always streams the current state.** It cannot download a historical managed backup; to move a retained backup off-host, copy its backup directory.
```
kriszyp: RocksDB's BackupEngine does not create one self-contained directory per backup ID. The files for retained IDs are split across shared metadata/data areas in `<backupPath>/<database>` (and Harper keeps logs in `transaction_logs/<id>`). Copying what looks like one backup directory can therefore omit shared files and produce an unrestorable disaster-recovery copy; a normal recursive copy can also race `create_backup`, `delete_backup`, or `purge_backups`. Please say to copy the entire per-database backup repository while backup writers are quiesced (or from an atomic filesystem snapshot), and update the same wording in `operations.md`.

reference/backups/operations.md
```diff
@@ -0,0 +1,140 @@
...
+
+```bash
+harper list_backups database=data
+```
+
+## `verify_backup`
+
+<VersionBadge version="v5.2.0" /> <EngineBadge engines="RocksDB" />
+
+Verifies a managed backup's file sizes, and optionally their checksums when `verify_checksum` is `true` (slower). Through a running server this runs as a background [job](../operations-api/operations.md#jobs). `backup_id` is required.
```
kriszyp: The implementation delegates to `backups.verify`, which verifies only BackupEngine-managed RocksDB files. Harper's separately copied `<repository>/transaction_logs/<backup_id>` subtree is not checked, even with `verify_checksum: true`. As written, "managed backup integrity" and "checksum-verified backups" imply the audit log is covered. Please either extend core verification to validate that snapshot or explicitly scope this operation to RocksDB database files and qualify the broad checksum-verification claims in the overview and release notes.

reference/backups/overview.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Example: incremental backups and restore
+
+Create a managed backup of the `data` database (the default):
+
+```bash
+harper create_backup database=data
+```
+
+The first backup copies the entire database; each one after that only copies files that changed, so frequent backups are cheap. Schedule `create_backup` as often as your recovery point requires, and manage retention with `purge_backups`:
```
kriszyp: Only the RocksDB engine files are incremental/shared. Harper calls `db.backup(..., { transactionLogs: true })`, and rocksdb-js copies a full transaction-log snapshot into `transaction_logs/<backup_id>` for every backup; that portion is explicitly non-incremental. With a large audit-retention window, frequent backups can repeatedly copy substantial data and consume much more space than this claims. Please qualify every "only copies what changed" / "frequent backups are cheap" statement and explain the transaction-log cost.

reference/backups/operations.md
```diff
@@ -0,0 +1,140 @@
...
+
+```bash
+harper create_backup database=data
+```
+
+## `list_backups`
+
+<VersionBadge version="v5.2.0" /> <EngineBadge engines="RocksDB" />
+
+Returns the managed backups for a database, each with its `backup_id`, `timestamp`, `size`, and `file_count`. Returns an empty array if no backups have been created yet.
```
kriszyp: Could we define these fields precisely? `timestamp` is epoch seconds, `size` is bytes from RocksDB `BackupInfo`, and `file_count` is the RocksDB file count. The separately stored transaction-log snapshot is excluded from both `size` and `file_count`, so callers cannot treat them as the backup's total on-disk footprint.

reference/backups/overview.md
```diff
@@ -0,0 +1,126 @@
...
+- **A restore is a point-in-time rollback.** In a replicated cluster, coordinate a restore with replication before bringing the node back.
+
+### When can a database be restored?
+
+An in-place restore purges and rewrites the database's files, which requires the database to be fully closed first. Whether a restore can run online depends on what is holding the database open:
+
+| Database                                           | Online `restore_backup` (server running) | Offline `harper restore_backup` (server stopped) |
+| -------------------------------------------------- | ---------------------------------------- | ------------------------------------------------ |
+| A user database not opened by any loaded component | Yes — restored in place                  | Yes                                              |
+| A user database that a loaded component keeps open | No — the job fails with a `409`          | Yes                                              |
```
kriszyp: Because `restore_backup` is created as a background job, the request has already returned a `job_id` before the job checks whether a component still holds the database open. That internal 409 becomes job status `ERROR` plus a message from `get_job`; the caller does not receive an HTTP 409 from the original restore request. Please describe this as the job entering `ERROR` with instructions to restore offline.

kriszyp (changes requested):
Ugh, sorry, I messed up the first review, I got the PR for docs and code confused. This looks good. I guess I will leave in the inline comments in place.

reference/backups/overview.md
```diff
@@ -0,0 +1,126 @@
...
+harper get_backup database=data target=https://node-2.example.com:9925 out=./data.tar.gz
+```
+
+The archive contains the database's current state — all tables plus the transaction log — and extracts into a directory that opens directly as a RocksDB database.
+
+To restore it, stop Harper, replace the database's directory under the storage path with the extracted archive, and start Harper again:
+
+```bash
+harper stop
+mv ~/hdb/database/data ~/hdb/database/data.old # optional: keep a copy of the previous database
```
Ethan-Arrowood, line 104: **Blocker.** This snapshot-restore recipe is incomplete against #1831's head and silently loses data. `get_backup`'s tar now includes `blobs/<rootIndex>/<relpath>` entries; extracting the whole archive with `-C ~/hdb/database/data` puts those inside the RocksDB directory instead of restoring them to their real blob roots. Any database with file-backed blobs restores "successfully" and is missing its blobs.

The blob surface is undocumented more broadly: `exclude_blobs`, blob snapshots in managed backups, and blob purge-and-rewrite on restore all need coverage, and the scope line at the top of this page ("all tables plus the audit/transaction log") plus the 5.2 release-notes entry need the blob carve-in.

reference/backups/overview.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Limitations
+
+- **Managed backups require the RocksDB storage engine.** For LMDB databases, use [`get_backup`](./operations.md#get_backup) or [volume snapshots](../cli/commands.md#backing-up-with-volume-snapshots).
+- **Whole-database granularity.** There is no per-table backup or restore. (The one exception: `get_backup` on an LMDB database can stream individual tables, and includes the audit store only with `include_audit`.)
+- **One storage root per database.** A database whose tables use per-table `path` storage configs spans multiple root stores and cannot be backed up with these operations.
+- **Backups live on the node that created them.** The backup repository is a local directory. For disaster recovery, copy backup directories off-host, or use `get_backup` to pull snapshots from a running server.
```
Ethan-Arrowood: Re-verified @kriszyp's point at #1831's current head — it stands. RocksDB BackupEngine backups aren't self-contained per id: there are shared data/metadata areas plus `transaction_logs/<id>` (and now `blobs/<id>`), so a per-id copy omits shared files, and a live copy races `create_backup`/`purge_backups`. This should say to copy the entire per-database repository while writers are quiesced.

reference/backups/overview.md
Ethan-Arrowood: Still accurate at #1831's head (@kriszyp raised this): `restore_backup` returns a `job_id` first, and the `verifyDatabaseClosed` 409 is thrown *inside* the job — so callers never see an HTTP 409, they see it via `get_job`. This matrix row should describe job status `ERROR` with the restore-offline message rather than "the job fails with a `409`".

reference/backups/overview.md
Ethan-Arrowood: The incremental-cost framing (@kriszyp's point, still standing) ignores that each backup copies the full transaction log — and now the blob snapshot — so "cheap incremental backups" overstates it. Worth stating what is and isn't incremental.

reference/backups/operations.md
Ethan-Arrowood: "Checksum-verified" overstates `verify_backup`'s scope (@kriszyp's point, verified still true): the implementation delegates to `backups.verify`, which only checks BackupEngine-managed files. The `transaction_logs/<id>` snapshot — and now the blob snapshot — is never verified, even with `verify_checksum: true`. Needs qualifying here, in the overview's "checksum-verified" bullet, and in the release notes.

reference/backups/operations.md
Ethan-Arrowood: "Must not already exist" is stricter than the code (@cb1kenobi's own note, verified at #1831 head): `restoreBackupOffline` accepts an existing **empty** directory via `isMissingOrEmptyDir`, rejecting only non-empty ones.

reference/backups/operations.md
Ethan-Arrowood: Worth defining these fields precisely (@kriszyp's point): `timestamp` is epoch seconds, and `size`/`file_count` come from RocksDB `BackupInfo`, which excludes the transaction-log and blob snapshots — so they aren't the backup's total footprint.

reference/backups/overview.md
Ethan-Arrowood, line 40: Also from @cb1kenobi's self-review and still open: a crashed restore leaves a `.restoring` marker, Harper then refuses to load the database, and the fix is simply rerunning `restore_backup`. Nothing in the docs says so — that's a bad state to hit undocumented.

Ethan-Arrowood (dismissed):
I re-verified @kriszyp's five inline comments against harper#1831's *current* head — all five still stand, so this needs another round regardless. On top of those, one finding of my own that I'd treat as the blocker: **blobs are undocumented end to end**, and the snapshot-restore example silently loses blob data for any database with file-backed blobs.

`get_backup`'s tar now carries `blobs/<rootIndex>/<relpath>` entries. The documented `tar -xzf … -C ~/hdb/database/data` extracts those *inside* the RocksDB directory, restoring no blobs to their real roots — a user follows the docs, gets a clean-looking restore, and has lost blob data. `exclude_blobs`, blob snapshots in managed backups, and blob purge-and-rewrite on restore are all missing, and the "all tables plus the audit/transaction log" scope line plus the 5.2 release-notes entry need the blob carve-in too.

One process note: your CHANGES_REQUESTED summary on this PR says "This looks good" — I think that landed on the wrong PR, since the inline comments are substantive. Either way a re-request will be needed to clear the state.

And the ordering question: #1831's blob work is itself under CHANGES_REQUESTED from kriszyp. These docs should re-sync once that settles — is a final drift pass planned before merge?

reference/backups/overview.md
cb1kenobi: Good catch — fixed in b2d60531. Limitations now says a DR copy must take the entire per-database repository (`<backupPath>/<database>`), not a single backup ID, and must run while no backup operation is in progress (or from an atomic filesystem snapshot); it also warns that a live recursive copy can race `create_backup`/`delete_backup`/`purge_backups`. The same guidance is carried into `operations.md` under `get_backup`.

— Generated by Barber AI

reference/backups/operations.md
cb1kenobi: Addressed in b2d60531 — `verify_backup` is now scoped, and the overview/release-notes claims are qualified. One clarification from reading #1831's current head: the transaction-log snapshot **is** verified (a framing check, always; RocksDB file checksums gated by `verify_checksum`) — it's the **blob** snapshot that isn't verified. So the docs now say exactly that rather than excluding the tx-log. Please sanity-check that against your read of `backups.verify`.

— Generated by Barber AI

reference/backups/overview.md
cb1kenobi: Fixed in b2d60531. The docs now state that only the RocksDB data files are shared/incremental, and that the transaction-log snapshot (and the blob snapshot) is copied in full on every backup — called out in "How managed backups work", the worked example, `create_backup`, and the 5.2 notes, with the large audit-retention cost noted.

— Generated by Barber AI

reference/backups/operations.md
cb1kenobi: Done in b2d60531. `list_backups` now defines each field: `backup_id` (monotonic int), `timestamp` (epoch seconds), `size` (bytes of RocksDB file payloads), `file_count` (RocksDB file count, some shared across backups) — and notes that `size`/`file_count` exclude the transaction-log and blob snapshots, so they undercount the true on-disk footprint.

— Generated by Barber AI

reference/backups/overview.md
cb1kenobi: Fixed in b2d60531. The matrix row now reads "the job ends in `ERROR`", and the prose explains that `restore_backup` returns a `job_id` first, the open-database check fails **inside** the job, and the failure surfaces via `get_job` as status `ERROR` (not an HTTP 409). The `system` database — and `target_database` while the server is running — are noted as synchronous up-front rejections before any job is created.

— Generated by Barber AI

reference/backups/overview.md
cb1kenobi: Thanks — this was the right blocker to raise. Fixed in b2d60531. The manual RocksDB restore is now split into two parts: RocksDB files into the database directory (`tar … --exclude='blobs*'`), and each `blobs/<rootIndex>/` tree into its blob root (default `<rootPath>/blobs/<database>`, or `storage.blobPaths`; the archive's `blobs/README.md` carries the index→path map), with an explicit warning that the naive single-extract silently loses blob data. Also documented `exclude_blobs`, blob snapshots in managed backups, and restore's blob purge-and-rewrite, and carved blobs into the top-of-page scope line and the 5.2 release notes.

— Generated by Barber AI

reference/backups/overview.md
cb1kenobi: Addressed in b2d60531 — see the reply on @kriszyp's original thread. DR copies now take the whole per-database repository (`<backupPath>/<database>`) while quiesced, with the race and the `blobs/<id>` / `transaction_logs/<id>` subtrees called out.

— Generated by Barber AI

reference/backups/overview.md
cb1kenobi: Addressed in b2d60531 — the matrix row and prose now describe job status `ERROR` via `get_job` rather than an HTTP 409, and note that `system` / `target_database`-while-running are synchronous up-front rejections.

— Generated by Barber AI

reference/backups/overview.md
cb1kenobi: Addressed in b2d60531 — the docs now separate the incremental RocksDB data files from the full-copy transaction-log and blob snapshots, and note the cost for large audit-retention windows.

— Generated by Barber AI

reference/backups/operations.md
cb1kenobi: Addressed in b2d60531. Note per #1831's current head: the tx-log snapshot **is** verified (framing, always), so only the blob snapshot is unverified — the docs now scope it that way rather than excluding the tx-log. Flagging in case that changed since your read.

— Generated by Barber AI

reference/backups/operations.md
cb1kenobi: Fixed in b2d60531 — `restore_backup` now reads "The target must not already exist, or must be an empty directory; Harper picks the new database up on the next start." Matches `isMissingOrEmptyDir`.

— Generated by Barber AI

reference/backups/operations.md
cb1kenobi: Addressed in b2d60531 — `list_backups` field definitions added, with the note that `size`/`file_count` come from the RocksDB backup engine and exclude the tx-log and blob snapshots.

— Generated by Barber AI

reference/backups/overview.md
cb1kenobi: Added in b2d60531 — a Limitations bullet and a `restore_backup` note now describe the interrupted-restore state (database marked incompletely restored, skipped on start with an incomplete-restore error) and the recovery: rerun `restore_backup` for the same database and `backup_id`; don't hand-repair the directory.

— Generated by Barber AI

reference/backups/overview.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Example: download a snapshot and restore it manually
+
+`get_backup` streams a full snapshot of a database from a running server — local or remote — which makes it the simplest way to keep backups off-host. Because it can stream from a remote `target`, it also doubles as a way to **clone** a database from another node onto this one: pull the snapshot with `target=<url>`, then restore it locally as shown below. What it streams depends on the storage engine, so the manual restore differs too. The examples below use the `data` database and the default storage path ([`storage.path`](../configuration/options.md#storage), default `<rootPath>/database`).
+
+### RocksDB <EngineBadge engines="RocksDB" />
+
+For a RocksDB database the stream is a `tar` archive, gzipped by default. It contains the database's current state — all tables, the transaction log, and, for a database with file-backed blobs, a `blobs/<rootIndex>/…` tree for each blob root (pass `exclude_blobs=true` to omit them):
```
kriszyp: Could we keep the new path notation and shell comments in plain ASCII per the repository content style? Please change `blobs/<rootIndex>/…` here and on line 110 to `blobs/<rootIndex>/...`, and replace the `→` characters in the comments on lines 114 and 118 with `->` (or plain prose).

kriszyp (approved):
Great work!

reference/backups/overview.md
cb1kenobi: Fixed in 7ddbe0e8 — the `blobs/<rootIndex>/...` notation (lines 92 and 110) and the `->` arrows in the manual-restore shell comments (lines 114 and 118) are now plain ASCII, matching the repo content style. Thanks for the approval and the catch!

— Generated by Barber AI

PR #591, opened 2026-07-16
docs: cross-database read consistency for sourcedFrom resolvers

kriszyp:
## Summary
Documents a confirmed behavior: `sourcedFrom` cache resolver reads are snapshot-atomic within a single Harper database, but not across databases — a resolver that reads database A, awaits, then reads database B can observe B's post-await state. Adds guidance to keep an assembled view's read sources within one database, or design for skew if it must span multiple.

Originated from QA finding D-222 (verified on harper main 3dbcf7b9e).

---
Signed, Opus 4.8

### DIFF
--- a/reference/resources/resource-api.md
+++ b/reference/resources/resource-api.md
@@ -446,6 +446,14 @@ class BlogSource extends Resource {
 Post.sourcedFrom(BlogSource);
 ```
 
+#### Read consistency across databases
+
+A resolver's reads are snapshot-consistent as long as every table it reads lives in the same [database](../database/overview.md#databases) — Harper pins a single transaction snapshot per database, so multiple `get()` calls into tables in that database always see the same point-in-time view, no matter how the resolver interleaves them.
+
+That snapshot is scoped per database. If a resolver reads from one database, `await`s something (an upstream fetch, another async call), and then reads from a second database, the second read is not guaranteed to reflect the same point in time as the first — it can observe writes that landed on the second database during the `await`.
+
+> **Note:** If a `sourcedFrom` resolver must assemble a self-consistent view from multiple tables, keep all of those tables in a single database. If it must span multiple databases, do not assume the combined result is atomic — design for skew between the sources instead.
+
 ---
 
 ### `getUpdatedTime(): number`


reference/resources/resource-api.md
```diff
@@ -446,6 +446,14 @@ class BlogSource extends Resource {
 Post.sourcedFrom(BlogSource);
 ```
 
+#### Read consistency across databases
+
+A resolver's reads are snapshot-consistent as long as every table it reads lives in the same [database](../database/overview.md#databases) — Harper pins a single transaction snapshot per database, so multiple `get()` calls into tables in that database always see the same point-in-time view, no matter how the resolver interleaves them.
```
cb1kenobi, line 451: **Medium: same-database snapshot guarantee has a known exception in shipped code**

This paragraph documents the *intended* design (a single `DatabaseTransaction` keyed by database path pins one snapshot spanning all column families in the env), and Claim 2 below (no cross-database atomicity) is accurate. But as of the currently shipped storage layer this same-database cross-table guarantee is not reliably true: [harper#1881](https://github.com/HarperFast/harper/issues/1881) reproduces (on 5.1.19 / 5.1.22 with `@harperfast/rocksdb-js` 2.4.0 — the version pinned in harper's `package-lock`) an async/cache-miss native read resolving against the transaction's own column family instead of the caller's, silently dropping rows on a second table's read within one request. The fix exists only as unmerged commits on rocksdb-js branch `kris/fix-cross-cf-async-reads` ([rocksdb-js#717](https://github.com/HarperFast/rocksdb-js/pull/717)), not yet an ancestor of `main`.

Suggested fix: either add a temporary caveat here noting the known exception (silent partial/empty results on the second table's async read, tracked in harper#1881), or gate this doc's merge on the rocksdb-js cross-CF fix (#717) landing so the documented guarantee matches behavior. Claim 2 can stay as-is.

—
_Generated by Barber AI_

cb1kenobi (approved):

Ethan-Arrowood (approved):
LGTM — both cross-database consistency claims check out against source (one pinned read snapshot per database, a fresh snapshot chained per additional database), no over-promise.

PR #593, opened 2026-07-17
Document operation authorization and explicit row filters

kriszyp:
Companion documentation for [HarperFast/harper#1915](https://github.com/HarperFast/harper/pull/1915) and [HarperFast/harper#1842](https://github.com/HarperFast/harper/pull/1842).

## Summary

- replace the automatic record-scoped `allowRead` documentation from #1786 with explicit `rowFilter` and subscription `eventFilter` guidance
- document operation overrides as the preferred place to make application authorization decisions with the complete target and request context
- clarify that legacy `allow*` hooks are deprecated one-time operation gates
- distinguish default instance behavior, built-in `loadAsInstance = false` behavior, and custom false-mode handlers that retain responsibility for authorization
- document HNSW predicate traversal, source-revalidated rows, subscription event composition, synchronous/fail-closed requirements, and non-authoritative tombstone/message handling
- correct the 5.2 release notes and describe false-mode search, subscribe, array PUT, query DELETE, and publish hook selection
- clarify custom MCP tool authorization, trusted `checkPermission = true` use, and collision-safe event-owner ID prefixes

## Validation

- Prettier formatting check
- full Docusaurus production build
- only the pre-existing broken `5.1#deployment-operations` anchor warning remains
- final factual review found no remaining blockers after correcting false-mode ownership, event-filter composition, MCP direct-call behavior, trusted authorization controls, generated metadata wording, and single-record loading nuance

Generated with GPT-5.6 Codex.

### DIFF
--- a/reference/database/schema.md
+++ b/reference/database/schema.md
@@ -535,51 +535,72 @@ let results = Document.search(
 Use a `rowFilter` function on a search target or subscription request when access depends on each record. Attach the filter in an operation override after deciding that the request itself is allowed to proceed. This keeps authorization ahead of query planning and table scans while still filtering the returned or delivered rows:
 
 ```javascript
+function canReadReport(record, context) {
+	const user = context.user;
+	if (user?.role?.permission?.super_user) return true;
+	return user?.username != null && record.ownerId != null && record.ownerId === user.username;
+}
+
 export class Reports extends tables.Reports {
 	get(target) {
 		const context = this.getContext();
 		if (target.isCollection) {
-			target.rowFilter = (record, liveContext) => record.ownerId === liveContext.user.id;
-		} else if (this.ownerId !== context.user.id) {
+			target.rowFilter = canReadReport;
+		} else if (!canReadReport(this, context)) {
 			return new Response(null, { status: 404 });
 		}
 		return super.get(target);
 	}
 
 	search(target) {
-		target.rowFilter = (record, context) => record.ownerId === context.user.id;
+		target.rowFilter = canReadReport;
 		return super.search(target);
 	}
 
 	subscribe(request) {
-		request.rowFilter = (record, context) => record.ownerId === context.user.id;
+		request.rowFilter = canReadReport;
 		return super.subscribe(request);
 	}
 }
 ```
 
-`rowFilter` receives the candidate record and the live request or subscription context. It must be synchronous, side-effect free, and fast. Ordinary object and array records are passed as shallow read-only views. Throwing an error or returning a promise aborts a query; on a subscription it terminates the stream. The filter is enforced across normal index and range searches, OR queries, HNSW traversal, source-revalidated records, subscription snapshots, and live row events.
+`rowFilter` is available only from the JavaScript API; clients cannot set it through REST or QUERY request data. It receives the candidate record and the live request or subscription context. It must be synchronous, side-effect free, and fast. Ordinary object and array records are passed as shallow read-only views. Throwing an error or returning a promise aborts a query; on a subscription it terminates the stream. The filter is enforced across normal index and range searches, OR queries, HNSW traversal, source-revalidated records, subscription snapshots, and live row events.
+
+`rowFilter` does not apply to a direct primary-key `get`. The example performs that check in `get()` after the default instance-loading flow has loaded the record. It therefore assumes the default `loadAsInstance` behavior. A false-mode handler must explicitly load any record data it needs before making the same decision.
 
 For vector queries, `rowFilter` participates in HNSW traversal. A caller therefore receives the k nearest _matching_ records rather than "nearest k, minus filtered records." `rowFilter` can apply to every candidate visited; prefer indexed query conditions for predicates that can be expressed declaratively.
 
-`rowFilter` applies only to authoritative row events. Delete tombstones and published messages may not contain the complete current record. A subscription that needs those events can also provide an `eventFilter`:
+`rowFilter` applies only when a non-raw `put` or `invalidate` event carries an authoritative row value. Delete tombstones, value-less invalidations, raw events, and published messages may not contain the complete current record. A subscription that needs such events can also provide an `eventFilter`:
 
 ```javascript
 subscribe(request) {
-	request.rowFilter = (record, context) => record.ownerId === context.user.id;
-	request.eventFilter = (event, context) =>
-		event.type === 'put' ||
-		event.type === 'invalidate' ||
-		(event.type === 'delete' && String(event.id).startsWith(`${context.user.tenantId}:`));
+	request.rowFilter = canReadReport;
+	request.eventFilter = (event, context) => {
+		const rowEvent = event.type === 'put' || event.type === 'invalidate';
+		if (!request.rawEvents && rowEvent && event.value != null) return true;
+		const username = context.user?.username;
+		const ownerPrefix = username == null ? null : `${encodeURIComponent(username)}:`;
+		return (
+			ownerPrefix != null &&
+			(rowEvent || event.type === 'delete') &&
+			String(event.id).startsWith(ownerPrefix)
+		);
+	};
 	return super.subscribe(request);
 }
 ```
 
+This example assumes report IDs are created with the same `encodeURIComponent(username) + ':'` owner prefix, allowing value-less events to be authorized without loading the deleted or invalidated record. Encoding the owner segment prevents one username from being a raw string prefix of another owner's IDs.
+
 `eventFilter` receives a shallow read-only view of every non-control event and the live context. It has the same synchronous and fail-closed requirements as `rowFilter` and is composed with it for authoritative row events. When a subscription has a `rowFilter`, non-row events are withheld unless an `eventFilter` explicitly accepts them. Transaction and reload control events continue to pass so the subscription remains coherent.
+
+Live subscriptions are periodically re-authorized with a freshly loaded user. This recheck reruns the operation-level `allowRead` grant; it does not call a custom `subscribe()` override again. The `rowFilter` and `eventFilter` callbacks receive the refreshed context for subsequent events. If an application-specific connection grant must terminate an existing stream when revoked, keep that revocable grant in an `allowRead` override composed with `super.allowRead`; admission logic that runs only in `subscribe()` cannot provide that teardown behavior.
 
 The legacy `allowRead`, `allowUpdate`, `allowCreate`, and `allowDelete` hooks are deprecated operation-level gates. When permission checking is active, Harper's standard instance flow evaluates the relevant hook once for the operation rather than once per row. Built-in table handlers with `loadAsInstance = false` likewise use one request/collection-scoped verdict; custom false-mode handlers remain responsible for authorization unless they delegate to those built-in handlers.
 
 Put application authorization and access-control decisions that need the complete target and context in operation overrides such as `get`, `put`, and `delete`. Collection operation overrides run before query planning or table scans and can attach a `rowFilter` or indexed conditions when an admitted request needs row-level narrowing. In the default single-record `get` flow, the record instance is loaded before the operation override runs, so its fields are available through `this`.
+
+See the [`RequestTarget`](../resources/resource-api.md#requesttarget) and [`SubscriptionRequest`](../resources/resource-api.md#subscriptionrequest-options) references for the filter properties.
 
 ### Tuning Filtered Traversal
 
--- a/reference/database/schema.md
+++ b/reference/database/schema.md
@@ -528,35 +528,58 @@ let results = Document.search(
 
 `vectorFilter` is available from the JavaScript API only (it cannot be expressed in a REST query string). The function receives the candidate record and must return a boolean — `true` to include the record in results, `false` to exclude it (it still routes traversal either way). It must be synchronous, side-effect free, and fast — it can run once per candidate record visited during traversal (verdicts are memoized per query). Records passed to it are frozen.
 
-### Record-Level Access Control (Record-Scoped `allowRead`)
+### Row-Level Access Control with Explicit Filters
 
 <VersionBadge version="v5.2.0" />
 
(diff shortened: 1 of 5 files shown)


reference/database/schema.md
```diff
@@ -558,6 +558,41 @@ How the one definition applies at each scope (when permission checking is active
...
+- **Single-record writes** (`PUT`/`PATCH`/`DELETE` on an id) — unchanged: evaluated at request entry with the record loaded, so the same override works there too.
+
+The default (non-overridden) hooks are table-level RBAC checks and keep their single request-entry evaluation with no per-record cost — including the collection-scope insert permission governing array `PUT`s. Operations API and SQL writes are governed by role permissions only; the `allow*` hooks apply to the Resource APIs (REST and JavaScript).
+
+One caveat: a class that overrides `delete()` itself replaces the framework's record-scoped delete path — its query-shaped deletes keep the request-entry check (a warning is logged when this combination is detected), and the custom `delete()` is responsible for any row-level enforcement.
```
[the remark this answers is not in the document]
kriszyp: This caveat is intentionally about the *instance* `delete()`: the framework's record-scoped delete path is the instance method carrying the `enforcesCheckPermission` marker, and it's an instance-level override that replaces it (the engine checks the resolved instance method). A `static delete` override is a different surface (it replaces the transactional entry point entirely), so rewording to `static delete()` would make the caveat incorrect. Keeping as-is. — KrAIs (Claude)

reference/database/schema.md
```diff
@@ -558,6 +558,117 @@ How the one definition applies at each scope (when permission checking is active
...
+		return this.ownerId === user.id; // per matching record
+	}
+	allowUpdate(user, updates, context) {
+		// `this` = the EXISTING record's resource; `updates` = the incoming data.
+		return this.ownerId === user.id;
+	}
+	allowCreate(user, record, context) {
+		// No existing record on a create — the incoming record is the parameter.
+		return record.ownerId === user.id;
+	}
```
dawsontoth: What happens with unauthenticated requests? Is `user` falsy? Or does it carry a falsy `id`?

dawsontoth: (We use `user?.` down below in the changes set forth here, that's why I ask.)

reference/database/schema.md
```diff
@@ -558,6 +558,117 @@ How the one definition applies at each scope (when permission checking is active
...
+	allowDelete(user, target, context) {
+		// Compose the table/RBAC grant first, so losing the role's delete denies.
+		if (!super.allowDelete(user, target, context)) return false;
+		if (user.role.permission.super_user) return true;
+		return this.ownerId === user.id; // per matching record
+	}
+	allowUpdate(user, updates, context) {
+		// `this` = the EXISTING record's resource; `updates` = the incoming data.
+		return this.ownerId === user.id;
+	}
```
cb1kenobi, line 558: **Medium: `allowUpdate`/`allowCreate` examples drop the RBAC composition shown for `allowDelete`**

The prose above the example states `super.allow*(...)` composes the role/RBAC baseline, and the `allowDelete` override demonstrates it (`if (!super.allowDelete(user, target, context)) return false;`). The `allowUpdate` and `allowCreate` overrides right below it skip that call entirely and gate solely on `this.ownerId === user.id` / `record.ownerId === user.id`. Copied verbatim, this authorizes any authenticated user who happens to own the row — including one whose role has no table `update`/`insert` permission at all — since nothing in the override consults RBAC. That's a real privilege gap if a reader lifts the pattern as-is, and it's inconsistent with the composition guidance stated one paragraph earlier.

Suggested fix: add `if (!super.allowUpdate(user, updates, context)) return false;` and `if (!super.allowCreate(user, record, context)) return false;` as the first line of each, matching the `allowDelete` example.

—
_Generated by Barber AI_

reference/database/schema.md
```diff
@@ -558,6 +558,117 @@ How the one definition applies at each scope (when permission checking is active
...
+}
+```
+
+#### Checklist
+
+1. List the classes that extend a table _and_ override an `allow*` hook. Anything extending a plain `Resource` can be skipped.
+2. For each, decide whether the override is meant to be a **row-level** decision (consults `this`) or a **whole-request** one. Make `allowRead` synchronous for the former; leave it `async` for the latter.
+3. Check whether any caller does a collection read, an array `PUT`, or a conditional `DELETE` against that table — including the automatic REST routes an `@export`ed table exposes, not just your own code paths.
+4. Update anything asserting on a `403` from those paths, which may now be a filtered `200`.
```
cb1kenobi: **Medium: checklist item conflates array `PUT` with the filter-semantics change, contradicting the doc's own array-PUT section**

Item 4 says a `403` from "a collection read, an array `PUT`, or a conditional `DELETE`" (item 3's list) "may now be a filtered `200`." That's correct for collection reads and conditional `DELETE`, but wrong for array `PUT`: the "Array `PUT` consults hooks that previously never ran" section 27 lines above explicitly states array `PUT` uses FAIL semantics — "any single denial fails the whole request with a `403` and commits no part of the batch" — never a filtered success. This is also what the implementation does (`AccessViolation` thrown per element aborts the whole transaction; there is no per-element filtering path for `allowUpdate`/`allowCreate`). A reader following this checklist item could wrongly conclude their array-PUT 403 assertions are now stale and safe to relax, when nothing about array PUT's status-code behavior changed — only that the hook is now invoked at all (a *new* 403 may appear where none did before, not the reverse).

Suggested fix: scope item 4 to the two hooks that actually filter — e.g. "Update anything asserting on a `403` from a collection read or conditional `DELETE`, which may now be a filtered `200`. Array `PUT` denials still return `403` — the change there is that the hook may now run for the first time (item 3)."

—
_Generated by Barber AI_

reference/database/schema.md
kriszyp: `user` is `undefined` for an unauthenticated caller — not an object with a falsy `id`. Confirmed against harper core (`security/auth.ts`): `request.user` is only assigned inside a matched auth branch (mTLS, `Authorization` header, session cookie, or local-bypass superuser); if none match, it's simply never set. The default hook implementations on `Resource` are written with `user?.role...` for exactly this reason, and it holds identically for `allowRead` and the write hooks — no read/write divergence.

Added a note to that effect in the doc: the `allowDelete` example's leading `super.allowDelete(...)` call denies before any `user.*` access runs (the default check is `user?.`-guarded), while `allowUpdate`/`allowCreate` skip that call and dereference `user.id` directly — which throws for an unauthenticated caller, but that still fails closed (denies), just via an exception rather than an explicit `false`.

— KrAIs (Claude Sonnet 5)

reference/database/schema.md
kriszyp: Kris's call on this one: "allow" is a grant, not a compositional restriction — RBAC could be consulted, but only as a fallback if the programmatic grant denies, not AND-ed on top as a second gate. Adding `if (!super.allowUpdate(...)) return false;` as a leading line (mirroring the `allowDelete` example) would impose exactly that double-restriction semantics, which isn't the intended composition model. Leaving `allowUpdate`/`allowCreate` as-is — the omission is deliberate, not dropped.

— KrAIs (Claude Sonnet 5)

reference/database/schema.md
kriszyp: Good catch, fixed — scoped item 4 to collection reads and conditional DELETE, and noted array PUT denials still return 403 (the change there is only that the hook may now run at all, per item 3).

— KrAIs (Claude Sonnet 5)

reference/database/schema.md
```diff
@@ -558,6 +558,119 @@ How the one definition applies at each scope (when permission checking is active
...
+
+The default (non-overridden) hooks are table-level RBAC checks and keep their single request-entry evaluation with no per-record cost — including the collection-scope insert permission governing array `PUT`s. Operations API and SQL writes are governed by role permissions only; the `allow*` hooks apply to the Resource APIs (REST and JavaScript).
+
+One caveat: a class that overrides `delete()` itself replaces the framework's record-scoped delete path — its query-shaped deletes keep the request-entry check (a warning is logged when this combination is detected), and the custom `delete()` is responsible for any row-level enforcement.
+
+For an unauthenticated caller, `user` itself is `undefined` — not an object with a falsy `id`. Leading with `super.allow*(user, ...)`, as the `allowDelete` example does, denies before any `user.*` access runs, since the default checks are written with `user?.`. An override that skips that call and dereferences `user.id` directly, as the `allowUpdate`/`allowCreate` examples do, will throw for an unauthenticated caller instead — which still fails closed (denies), just via an exception rather than an explicit `false`.
```
cb1kenobi: **Medium: this clarification documents a side effect of the RBAC-bypass gap without closing it**

This paragraph explains that skipping `super.allow*(...)` makes an unauthenticated caller throw instead of returning `false` — accurate, but it only addresses that narrow edge case. It doesn't fix why the `allowUpdate`/`allowCreate` examples above (lines 575-582) omit `super.allowUpdate(user, ...)` / `super.allowCreate(user, ...)` in the first place: any *authenticated* user who satisfies the ownership check (`this.ownerId === user.id`) is granted the write regardless of their role/RBAC permission on that table, because the RBAC baseline that `super.allow*(...)` composes (per the prose two paragraphs above) is never consulted for those two hooks. That's the privilege-bypass risk flagged in the prior review round — this addition explains a symptom of the gap rather than removing it.

Suggested fix: add `if (!super.allowUpdate(user, updates, context)) return false;` and `if (!super.allowCreate(user, record, context)) return false;` to the two examples, matching the `allowDelete` example immediately above. This paragraph can then be trimmed to the still-true note about `user` being `undefined` for unauthenticated callers.

—
_Generated by Barber AI_

cb1kenobi (approved):

reference/database/schema.md
kriszyp: The allow* methods are intended to programmatically provide "grants" to users otherwise not allowed (hence "allow"), not restricting users who already have permission (and could otherwise access the content through other means).

cb1kenobi (approved):

reference/database/schema.md
```diff
@@ -558,6 +558,119 @@ How the one definition applies at each scope (when permission checking is active
...
+- **Array `PUT` into a collection** — with an overridden `allowUpdate`/`allowCreate`, each element is authorized individually: an element whose record already exists is checked with `allowUpdate` (`this` = the existing record's resource), a new element with `allowCreate`. Any denial fails the **whole request** with a 403 and aborts the transaction — no partial batch is committed (each element was an explicit write target, unlike a query's incidental matches).
+- **Single-record writes** (`PUT`/`PATCH`/`DELETE` on an id) — unchanged: evaluated at request entry with the record loaded, so the same override works there too.
+
+The default (non-overridden) hooks are table-level RBAC checks and keep their single request-entry evaluation with no per-record cost — including the collection-scope insert permission governing array `PUT`s. Operations API and SQL writes are governed by role permissions only; the `allow*` hooks apply to the Resource APIs (REST and JavaScript).
```
Ethan-Arrowood: Defaults aren't fully unchanged: the create-within-record permission moved from `update` to `insert` in #1842. Folding that into this sentence, which is the natural home since it already describes default write-hook RBAC:

```suggestion
The default (non-overridden) hooks are table-level RBAC checks and keep their single request-entry evaluation with no per-record cost — including the collection-scope insert permission governing array `PUT`s. One default did change in 5.2, independent of any override: a record-targeted `publish`/`post` (creating _within_ an existing record) is now authorized against the table's **`insert`** permission, where it previously delegated to **`update`** (a create inside a record was treated as an update to that record). A caller with `update` but not `insert` that relied on `publish`/`post` to a record id will now receive a `403`. Operations API and SQL writes are governed by role permissions only; the `allow*` hooks apply to the Resource APIs (REST and JavaScript).
```

release-notes/v5-lincoln/5.2.md
```diff
@@ -12,7 +12,33 @@ All patch release notes for 5.2.x are available on the [releases page](https://g
...
+
+## Record-Scoped Authorization
+
+The `allow*` authorization hooks on tables are now **record-scoped when overridden**: instead of a single collection-scope verdict per request, an application-overridden hook is evaluated once per record, with access to that record's fields. Framework defaults are unchanged — a table that does not override a hook keeps its single request-entry RBAC check with no per-record cost — and the change applies only to classes that extend a table; a plain `Resource` subclass defines its own semantics and keeps the single entry check.
```
Ethan-Arrowood: "Framework defaults are unchanged" isn't quite right — the create-within-record RBAC operand changed (see #1842). Suggested rewording that keeps the record-scoping point but adds the operand change:

```suggestion
The `allow*` authorization hooks on tables are now **record-scoped when overridden**: instead of a single collection-scope verdict per request, an application-overridden hook is evaluated once per record, with access to that record's fields. The record-scoping change applies only to classes that extend a table — framework defaults keep their single request-entry RBAC check with no per-record cost, and a plain `Resource` subclass defines its own semantics and keeps the single entry check. One default RBAC operand did change: a record-targeted `publish`/`post` (creating _within_ an existing record) now requires the table's **`insert`** permission instead of **`update`**, so a caller with `update` but not `insert` on that table will now get a `403`.
```

Ethan-Arrowood (changes requested):
Requesting changes on one substantive gap. #1842 doesn't only add the record-scoped overrides documented here — it also tightens a *default* RBAC rule: it removed the `allowCreate` branch where creating *within* a record delegated to `allowUpdate`, so a record-targeted `publish`/`post` now requires the table's `insert` permission instead of `update`. That's a silent, user-visible change for apps that override nothing — `recordScopedWriteAuth.test.js` ("record-targeted publish uses insert RBAC because it creates a message entry") pins it: `insert:false, update:true` → 403, `insert:true, update:false` → 200.

Both this reference page and the 5.2 release note currently say framework defaults are "unchanged," which is misleading on exactly this point. A caller that had `update` but not `insert` and relied on `publish`/`post` to a record id will start getting 403s after upgrading, so it needs to be called out explicitly. Inline suggestions on both files below.

cb1kenobi (approved):

Ethan-Arrowood (approved):
Approving — the rewrite resolves my earlier change request, and the content traces accurately to the current state of both Harper PRs. Please hold merging until HarperFast/harper#1842 and HarperFast/harper#1915 actually land, since these docs remove behavior that still ships in 5.2.0-beta.3.

PR #595, opened 2026-07-17
docs: document middleware host and path routing

kriszyp:
## Summary

- document built-in virtual-host and URL-prefix middleware routing
- add `host`, `urlPath`, and middleware ordering fields to the HTTP API reference
- clarify that scoped plugin configuration automatically applies routing to HTTP, WebSocket, and upgrade handlers
- add the feature to the 5.2 release notes

Documents [Middleware ordering and routing (HarperFast/harper#397)](https://github.com/HarperFast/harper/pull/397).

## Verification

- `npm run format:write`
- `npm run format:check`
- `npm run typecheck`
- `npm run build` (passes with one pre-existing broken-anchor warning in the 5.1 release notes)

### DIFF
--- a/reference/http/api.md
+++ b/reference/http/api.md
@@ -182,7 +182,7 @@ type WsListener = (ws: WebSocket, request: Request, chainCompletion: Promise<voi
 
 ### `WsOptions`
 
-`WsOptions` supports the same routing and ordering options as [`HttpOptions`](#httpoptions), plus:
+WsOptions supports the same options as [HttpOptions](#httpoptions), plus:
 
 | Property     | Type   | Default | Description                                     |
 | ------------ | ------ | ------- | ----------------------------------------------- |
@@ -225,7 +225,7 @@ type UpgradeListener = (request: IncomingMessage, socket: Socket, head: Buffer, 
 
 ### `UpgradeOptions`
 
-`UpgradeOptions` supports the same routing and ordering options as [`HttpOptions`](#httpoptions).
+UpgradeOptions supports the same options as [HttpOptions](#httpoptions).
 
 ---
 
--- a/reference/http/api.md
+++ b/reference/http/api.md
@@ -40,11 +40,31 @@ To continue the middleware chain, call `next(request)`. To short-circuit, return
 
 ### `HttpOptions`
 
-| Property     | Type    | Default           | Description                                   |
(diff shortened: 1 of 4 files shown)


reference/http/api.md
```diff
@@ -40,11 +40,31 @@ To continue the middleware chain, call `next(request)`. To short-circuit, return
...
+| ------------ | ------- | ---------------------------- | ----------------------------------------------------------------------------------------------------------------------------- |
+| `name`       | string  | Registering component's name | Name this middleware entry so other entries can position themselves relative to it                                            |
+| `before`     | string  | -                            | Run this entry before the named middleware entry                                                                              |
+| `after`      | string  | -                            | Run this entry after the named middleware entry                                                                               |
+| `urlPath`    | string  | -                            | Only handle requests whose pathname matches this prefix. Harper removes the prefix before passing the request to the handler. |
+| `host`       | string  | -                            | Only handle requests whose `Host` header matches this virtual hostname                                                        |
```
cb1kenobi: **Low: host matching is case-sensitive but the doc doesn't say so**

`matchesRoute()` in `server/middlewareChain.ts` compares the `Host` header to `route.host` with a case-sensitive check (`requestHost !== route.host`), but HTTP `Host` headers are conventionally case-insensitive. A client sending `Host: API.example.com` against a configured `host: api.example.com` silently misses the routed chain and falls through to the default (or a 404) instead of matching. Worth calling out here since it fails silently in production.

Suggested fix:
```suggestion
| `host`       | string  | -                            | Only handle requests whose `Host` header matches this virtual hostname (case-sensitive)                                       |
```

—
_Generated by Barber AI_

cb1kenobi (approved):

Ethan-Arrowood (approved):
LGTM — I traced the routing semantics against the merged `middlewareChain.ts` and they're accurate: host+path > host > path precedence with a longer-path tiebreak, both-must-match, prefix stripping, and the `name` default.

Two clarifications worth folding in so readers don't trip on edge cases:
- `urlPath` matching is segment-boundary-aware — `/api` matches `/api` and `/api/x` but not `/apinews` — it's not a raw string prefix.
- Host matching ignores the port.

Kris's host case-sensitivity thread is worth resolving too. Approving; these are non-blocking.

PR #597, opened 2026-07-18
docs(models): local-development setup matching production

heskew:
## Summary

Closes #596 — the last unwritten piece of **HarperFast/harper#510**'s docs acceptance ("Documented: writing a custom backend, configuring fallback, local-dev setup matching production"). The other two-thirds shipped in #554 (custom backends) and #558 (routing & fallback, per HarperFast/harper#1326's conclusion that fallback is a router concern layered on the routing seam, not core facade behavior).

- `reference/models/local-development.md` — new page: logical-name portability (same app code local and production), a local Ollama configuration with its matching production configuration (`${VAR}` credential indirection), parity caveats (the `ollama` backend does not advertise tools — workaround via `openai` backend `baseUrl` against an OpenAI-compatible local server; embedding vectors are not comparable across models, which matters when `@embed` data or HNSW indexes move between environments), per-environment configuration via per-instance config files or `HARPER_CONFIG` merge, and offline/CI stubs via `models.registerBackend` (registered backends shadow same-name config entries).
- `reference/models/overview.md` — one-line cross-link from the "configuration change, not a code change" sentence.
- `sidebarsReference.ts` — sidebar entry under AI & Models.

## Verification

- `npm run build` passes; no broken links/anchors from the new page (the one broken-anchor warning is pre-existing on `release-notes/v5-lincoln/5.1`).
- Prettier clean (`format:check`).
- Claims verified against harper `resources/models/bootstrap.ts` (env-var expansion, per-entry skip-on-error), `backends.md`/`routing.md` shipped docs, and the registered-backend-shadows-config precedence documented in #554.

## Version

Page badge is **v5.1.0** (the models surface it documents); the Offline and CI stubs section carries its own **v5.1.15** badge (custom-backend registration).

---

https://claude.ai/code/session_01Vq8zum5E6fbjzmLsdDDkJh

## Cross-model review (standard: Gemini + Harper-domain adjudication)

No blockers. Both findings addressed in d7c8e878:

- **[Claude+Gemini, significant] `HARPER_CONFIG` partial-override hazard** — the original example overrode a baked ollama entry to openai. `HARPER_CONFIG` merges leaf keys and cannot remove them, so ollama's `host` would survive on the openai entry — an unrecognized field that fails startup validation (`configValidator` `.unknown(false)`). The page now teaches: supply a `models` block the file doesn't define, or override same-backend fields; never switch backends via partial override, and cover both `embedding` and `generative` maps.
- **[Gemini, suggestion] stub scoping** — added: `defineBackend` derives `generate` from `generateStream` but not the reverse (a stream-calling suite needs `generateStream` on the stub), and a `'generative'` stub doesn't serve `embed()` (embedding tests need an `'embedding'`-kind stub).

All other page claims were verified against harper `origin/main` (`bootstrap.ts`, `backendRegistry.ts`, `configValidator.ts`, `ollama`/`openai` components): registered-backend shadowing, ollama `tools: false`, `${VAR}` expansion under `HARPER_CONFIG`, the `defineBackend` return shape, and the validation-vs-registration startup split.

### DIFF
--- a/reference/models/local-development.md
+++ b/reference/models/local-development.md
@@ -71,13 +71,15 @@ Two things do not automatically carry across a backend swap:
 
 ## Per-environment configuration
 
-Each Harper instance reads its own configuration file, so the simplest arrangement is a development config with local backends and a production config with hosted ones. Where baking a config file into an image is awkward — containerized or orchestrated deployments — the `models` block can be overridden from the environment with [`HARPER_CONFIG`](../configuration/overview#harper_config), which merges exactly the keys it names over the config file:
+Each Harper instance reads its own configuration file, so the simplest arrangement is a development config with local backends and a production config with hosted ones. Where baking a config file into an image is awkward — containerized or orchestrated deployments — the `models` block can be supplied from the environment with [`HARPER_CONFIG`](../configuration/overview#harper_config), which merges exactly the keys it names over the config file:
 
 ```bash
-export HARPER_CONFIG='{"models":{"generative":{"default":{"backend":"openai","apiKey":"${OPENAI_API_KEY}","model":"gpt-4o"}}}}'
+export HARPER_CONFIG='{"models":{"embedding":{"default":{"backend":"openai","apiKey":"${OPENAI_API_KEY}","model":"text-embedding-3-small"}},"generative":{"default":{"backend":"openai","apiKey":"${OPENAI_API_KEY}","model":"gpt-4o"}}}}'
 ```
 
 The `${OPENAI_API_KEY}` placeholder is resolved by Harper at startup, not by the shell — the single quotes are deliberate.
+
+Because the merge is key-by-key and cannot remove keys, use `HARPER_CONFIG` to supply a `models` block the config file does not define (as above), or to override fields within an entry that keeps the same backend — a model name, a credential. Do not use it to switch an existing entry to a different backend: keys the override does not name stay in place, and a leftover field from the old backend — `host` from an `ollama` entry, say — is an unrecognized field on the new backend's entry, which [fails configuration validation](./overview#startup-behavior) and prevents startup. To swap backends between environments, swap the config file (or define every model entry through `HARPER_CONFIG` and keep the file's `models` block empty), and cover both the `embedding` and `generative` maps — an override that names only one leaves the other pointed wherever the file pointed it.
 
 While iterating on configuration, remember the [startup behavior](./overview#startup-behavior) split: a structurally invalid entry fails configuration validation and prevents startup, while a registration-time error — such as an unreachable backend module — is logged and skipped, and calls to that logical name then throw as unconfigured. If a model works in one environment and throws "not configured" in another, check the startup log for a skipped registration.
 
@@ -103,3 +105,5 @@ models.registerBackend(
 ```
 
 The stub still exercises routing, accounting, and [analytics](./analytics) — only the inference itself is faked.
+
+Two scoping notes. `defineBackend` derives `generate` from a supplied `generateStream` (by draining the stream), but not the reverse — if the suite calls `generateStream()`, give the stub a `generateStream` implementation. And a backend registered under `'generative'` serves only generation: `embed()` resolves the `'embedding'` registry, so embedding tests need their own stub registered with `models.registerBackend('embedding', 'default', …)`.
--- a/reference/models/local-development.md
+++ b/reference/models/local-development.md
@@ -65,7 +65,25 @@ The [`@embed` schema directive](../database/schema#embed), [tool calling](./tool
 
 Two things do not automatically carry across a backend swap:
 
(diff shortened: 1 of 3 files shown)


reference/models/local-development.md
```diff
@@ -0,0 +1,105 @@
...
+models.registerBackend(
+	'generative',
+	'default',
+	models.defineBackend({
+		name: 'stub',
+		async generate() {
+			return { status: 'completed', output: { content: 'stub reply', finishReason: 'stop' } };
+		},
+	})
+);
```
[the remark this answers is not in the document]
heskew: Addressed in 7455458b — added: register the stub during component initialization (e.g. `handleApplication`) rather than at a test file's top level, since each worker thread keeps its own registry; cross-linked `registerBackend()`'s guidance. (Matches the existing per-thread registration note in `backends.md`.)

🤖 Posted by Claude (Fable 5) on Nathan's behalf

dawsontoth (approved):

reference/models/local-development.md
```diff
@@ -0,0 +1,111 @@
...
+const reply = await models.generate('Summarize this record.');
+```
+
+The [`@embed` schema directive](../database/schema#embed), [tool calling](./tool-calling), [fallback groups](./routing#fallback-groups), and [analytics](./analytics) all address models the same way, so they carry across the swap too — subject to the parity caveats below.
+
+## Parity caveats
+
+Two things do not automatically carry across a backend swap:
+
+**Tool support.** The `ollama` backend does not advertise the `tools` capability, so a `generate()` call that declares tools fails up front against it (see [Backends](./backends#ollama)). If your application uses [tool calling](./tool-calling), run an OpenAI-compatible local server (such as vLLM) and select it with the [`openai` backend's `baseUrl` field](./backends#openai) instead — the local backend then advertises tools the way production does.
```
kriszyp: The `openai` backend does advertise `tools`, but that alone does not make a default vLLM server tool-capable. vLLM's automatic tool choice is disabled by default and requires a tool-capable model plus `--enable-auto-tool-choice` and a matching `--tool-call-parser` (and some models need a tool chat template). Harper also requires a non-empty `apiKey` for every `openai` entry, even if the local server itself does not authenticate. As written, someone can follow this paragraph, pass Harper's up-front capability check, and then fail at vLLM. Could we include a minimal `vllm serve ...` command and Harper YAML example (including the matching/dummy API key), or narrow the wording and link directly to vLLM's tool-calling setup?

kriszyp (commented):
I think it makes sense to provide the vLLM instruction.

reference/models/local-development.md
heskew: Addressed in edaa988e — reworked the paragraph around both requirements: (1) the server must have tool calling enabled (the `openai` backend always advertises `tools`, so Harper's up-front check passes regardless and an unconfigured server fails at request time), with a runnable `vllm serve Qwen/Qwen2.5-7B-Instruct --enable-auto-tool-choice --tool-call-parser hermes` example and a link to vLLM's tool-calling guide for the model–parser pairings; (2) the non-empty `apiKey` requirement, shown in matching YAML with a placeholder value and a note that the literal-credential startup warning is expected there.

🤖 Posted by Claude (Fable 5) on Nathan's behalf

reference/models/local-development.md
```diff
@@ -0,0 +1,129 @@
...
+
+When tests must run with no model server at all, register an in-process stub as a [custom backend](./backends#custom-backends) under the logical names the application uses. A registered backend shadows a configuration entry with the same name, so the stub wins regardless of what the config file says:
+
+```javascript
+import { models } from 'harper';
+
+models.registerBackend(
+	'generative',
+	'default',
+	models.defineBackend({
```
kriszyp, line 116: The section frames this stub as an offline replacement under the application's existing logical name, but `defineBackend()` defaults `tools` to `false`. If the application declares tools, `models.generate()` automatically requires that capability and throws before this `generate()` method runs. Could the scoping notes call this out as a third case? Tool-calling tests need `tools: true` plus a stub implementation that returns the tool-call sequence the test expects; I would avoid adding the flag to this simple reply stub without that qualification.

kriszyp (approved):

Ethan-Arrowood (approved):
LGTM — verified the setup steps against main and confirmed all the cross-doc anchors resolve.

reference/models/local-development.md
heskew: Good catch — this merged before your note was addressed. Follow-up in #621: third scoping note added, with the qualification you suggested (tool tests need `tools: true` plus a `generate()` returning the expected tool-call sequence; the simple reply stub stays flagless). Verified against `backendRegistry.ts` (`tools = false` default) and `Models.ts` (tools auto-require) on harper main.

🤖 Posted by Claude (Fable 5) on Nathan's behalf

PR #599, opened 2026-07-21
docs(deploy): two-phase deploy, revert, by-reference deploys, and OIDC trusted publishing

dawsontoth:
> **Draft.** Documentation companion to the v5.3.0 deploy work. Stacked on [#630](https://github.com/HarperFast/documentation/pull/630), which carries the v5.2.0 CLI-token content that used to live here. Everything remaining describes unreleased surface — see the dependency table.

Depends-on: HarperFast/harper#1849

## What this documents

Six files, not the two the previous description listed. The cluster-wide deploy lifecycle, rollback, deploy-by-reference, sealed deploy credentials, server-side SSH keygen, and OIDC trusted publishing.

| Surface | Upstream | State |
| --- | --- | --- |
| Surface | Upstream | Ships in |
| --- | --- | --- |
| `deploy by_ref=true` / `ref=` / `credential=` | [harper#1850](https://github.com/HarperFast/harper/pull/1850) | **v5.2.3 — shipped** |
| `deploy setup=true` | [harper#1851](https://github.com/HarperFast/harper/pull/1851) | **v5.2.3 — shipped** |
| `add_ssh_key generate: true` | [harper-pro#594](https://github.com/HarperFast/harper-pro/pull/594) | **v5.2.4 — shipped** |
| OIDC trusted publishing | [harper#2173](https://github.com/HarperFast/harper/pull/2173) | merged, unreleased → v5.3.0 |
| Two-phase deploy, `revert_component`, staging retention | [harper#1849](https://github.com/HarperFast/harper/pull/1849) | **still a draft** → v5.3.0 |

## Two things changed since the last review pass

**The versions have moved twice, and are no longer uniform.** harper `main` is at **5.2.4**, and `git tag --contains` on each merge commit puts by-reference deploys and sealed credentials in **v5.2.3** and server-side SSH keygen in **v5.2.4** — all three are in users' hands already. OIDC merged 2026-08-21 and is unreleased; two-phase is still a draft. Both of those carry `v5.3.0`, confirmed by the `5-3-0.ts` upgrade directive on main. Release notes are split the same way: the shipped features moved to `5.2.md` labelled with their patch, and `5.3.md` keeps two-phase and OIDC.

**The one piece that genuinely shipped in v5.2.0 has left.** `login --for-ci` and the token env vars (harper#1876, in v5.2.0 since 2026-07-31) are now [#630](https://github.com/HarperFast/documentation/pull/630), against `main`, mergeable today. That answers @Ethan-Arrowood's first sequencing question — the shipped half no longer waits on four unmerged PRs — and it puts the token rules inside `main`'s canonical numbered **Authentication Precedence** section rather than in a competing paragraph beside it.

## Review feedback

**@kriszyp's eight findings** (CHANGES_REQUESTED), each re-checked against upstream as it stands now rather than as it stood when filed. Four had been fixed at the source in the meantime.

| # | Finding | Resolution |
| --- | --- | --- |
| 1 | "All-or-nothing at go-live" is wrong | **Docs fixed.** Reframed as an all-nodes staging barrier. |
| 2 | Deploy modes presented as independent knobs | **Fixed upstream**, now documented as a matrix. |
| 3 | Explicit `ref` not resolved to a SHA | **Fixed upstream** (`5c85714b8`, `26cce986d`, `21043f994`). |
| 4 | Rollback scope and restart claims too broad | **Docs fixed.** Scoped to the two-phase path; `restart` documented as optional. |
| 5 | Absolute SSH-key encryption guarantee | **Docs fixed.** Qualified — the plaintext fallback is deliberate upstream. |
| 6 | `auth_*` vs. ordinary `username=` precedence | **Fixed upstream**; `main`'s canonical list covers it, #630 extends it. |
| 7 | Plain "Available since" instead of badges; release notes | **Docs fixed.** Badges + new `5.3.md`. |
| 8 | Missing badges, summary-table row, release notes | **Docs fixed.** Plus the `revert_component` row. |

Worth expanding on three:

- **Finding 1 was a self-contradiction**, not just imprecision: the page claimed all-or-nothing in its opening paragraph and then correctly described a version split twenty lines later. The staging barrier is what the two phases buy — it eliminates the common failure, a node that can't fetch or install — but activation is still per node. The activation-failure discussion is now its own `#### Activation failures` section rather than sitting mid-parameter-list, and the opening paragraph points at it.
- **Finding 2 became documentable** because core added the validation @kriszyp asked for. `activate: false` and `deployment_id` are rejected when combined with `two_phase: false`, `replicated: false`, or an unreplicated `system` database, so a request that says stage-only can no longer activate. There is now a valid-combinations table, including that `revert_on_failure` is refused rather than ignored.
- **Finding 5 stands as a docs problem** because core chose to keep the fallback: `sealSSHKey` still stores and replicates the private key in plaintext with a WARN when no secret custody is registered, deliberately, since SSH keys predate custody. So encryption at rest is a property of the deployment's configuration, not of the operation — stated as a warning, while noting custody is present by default so the risk doesn't read as likely.

**@Ethan-Arrowood's structural hold** is addressed:

1. **Sequencing** — split, as above. The remaining content is uniformly unshipped v5.3.0, and the machine-readable marker at the top of this description arms it for [#629](https://github.com/HarperFast/documentation/pull/629)'s companion check.
2. **The conflict with `main`** — resolved by hand, and it was the hazard he predicted. The ops table needed a three-way merge: `main`'s seven backup rows (`create_backup`, `list_backups`, `verify_backup`, `delete_backup`, `purge_backups`, `restore_backup`, `get_backup`), this branch's four deploy rows, and this branch's recategorization. Taking either side wholesale would have silently dropped one set. I compared both sides row by row first and confirmed no description text differed, so the width churn is only Prettier re-padding. `authentication.md` no longer states precedence twice, since #630 owns that file's content.
3. **Duplicate precedence, stale body, missing config keys, `(draft)` in the title** — the first is gone with the split, this description is rewritten, the title marker is dropped in favour of GitHub's own draft state.
4. **The `system`-replication precondition** (his second question) — now documented. A plain deploy on a cluster with `system` excluded from replication silently takes the one-shot path, which also means no rollback target; the staged parameters are rejected rather than downgraded.

@cb1kenobi's five findings were fixed in the previous pass and remain so.

## Still open

- **`deployment_stagingRetention_maxCount` and `deployment_payloadRetention_*` are documented inline, not in `reference/configuration/options.md`.** @Ethan-Arrowood is right that operators look there, and right that the original justification ("matching how `payloadRetention_maxSize` is surfaced") was false — it isn't surfaced anywhere. Left as-is deliberately: three config keys for an unreleased feature seem better added when the options page can be updated in one pass against the shipping release. Happy to add them here instead.
- **Most of this is no longer gated.** Only two-phase and `revert_component` depend on an unmerged PR, so the marker names harper#1849 alone. Everything else is shipped or merged — which makes splitting the ungated content into its own PR worth considering rather than holding it behind a draft.

## Verification

`npm run build` and `npm run format:check` clean. The build reports two broken anchors; both are pre-existing on `main` (confirmed by building `main` with these changes stashed) and neither is in a file this PR touches. Claims re-verified against the upstream branches; harper#2173's description has since been updated by its author and no longer diverges. Fourteen findings from the third cross-model round remain unaddressed and unverified — see the review-coverage note below.


reference/cli/operations-api-commands.md
```diff
@@ -34,87 +34,94 @@ The following operations are available through the CLI. Operations that require
...
+| `deploy_component`               | Deploy a component                            | [Components](../operations-api/operations.md#components)               | v4.3.0          |
+| `deploy` (alias)                 | Alias for `deploy_component`                  | [Components](../operations-api/operations.md#components)               | v4.3.0          |
+| `stage` (alias)                  | Stage a component without going live          | [Components](../operations-api/operations.md#deploy_component)         | v5.2.0          |
+| `activate` (alias)               | Activate a previously-staged deploy           | [Components](../operations-api/operations.md#deploy_component)         | v5.2.0          |
+| `revert_component`               | Roll a component back to its previous version | [Components](../operations-api/operations.md#revert_component)         | v5.2.0          |
+| `revert` (alias)                 | Alias for `revert_component`                  | [Components](../operations-api/operations.md#revert_component)         | v5.2.0          |
```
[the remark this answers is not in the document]
dawsontoth: Done — pointed the `stage`/`activate`/`revert_component`/`revert` category links at `#components` to match every other row in the table.

reference/operations-api/operations.md
```diff
@@ -624,26 +637,58 @@ Response:
...
+{ "operation": "deploy_component", "project": "my-app", "package": "my-org/my-app#semver:v1.2.3", "activate": false }
+```
+
+```json
+{ "deployment_id": "a3f8c2d1...", "project": "my-app", "staged": true, "message": "Staged component: my-app" }
+```
+
+```json
+{ "operation": "deploy_component", "project": "my-app", "deployment_id": "a3f8c2d1...", "restart": "rolling" }
+```
```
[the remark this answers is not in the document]
dawsontoth: Done — labeled the three blocks **Stage request** / **Stage response** / **Activation request** (with a one-line note on each) so the stage→activate flow reads clearly, matching the "A normal deploy" example above.

reference/operations-api/operations.md
```diff
@@ -602,10 +602,27 @@ Detailed documentation: [Components Overview](../components/overview.md)
...
+:::
+
 Additional parameters:
 
 - `urlPath` — override the HTTP URL path the component is mounted at (e.g. `"/api/v2"`)
 - `install_allow_scripts` — set to `true` to allow npm pre/post install scripts (disabled by default)
+- `activate` — set to `false` to **stage only** and stop before go-live. The build is prepared and verified on every node and the response returns a `deployment_id` in a `staged` state; nothing goes live. Activate it later by calling `deploy_component` again with that `deployment_id` (see below). Useful for pre-staging a release and flipping it live in a separate, fast step.
+- `deployment_id` — activate a previously-staged deployment (from an `activate: false` call). No new payload is fetched or installed; the already-staged build is swapped live cluster-wide. `project` is still required.
```
cb1kenobi: **Medium: `deployment_id` activate doesn't persist root config for a package deploy**

In the reference implementation (harper#1849, `deployComponentActivateExisting` in `components/operations.js`), `writeComponentRootConfig` only runs `if (req.package)`. The activation example just below (line 714) omits `package` — consistent with this bullet's own "no new payload is fetched or installed" framing — so a real activate-by-`deployment_id` call for a `package` deploy will *not* update `harperdb-config.yaml`'s stored `package:` reference for the component. A node that restarts without ever having this version installed, or a new peer joining the cluster, re-installs from that stale root config — silently reverting to the pre-activation package. Worth a caveat here (or requiring `package` again on the activate call) so this isn't assumed to be as durable as a normal single-shot deploy.

Suggested fix: add a line such as "Note: for a `package` deploy, also pass `package` on this call to durably update `harperdb-config.yaml` — omitting it activates the build cluster-wide but leaves the stored package reference unchanged, so a future restart or new peer may reinstall the older version."

—
_Generated by Barber AI_

reference/operations-api/operations.md
```diff
@@ -673,26 +690,75 @@ Response:
...
+
+**Activation request** (take the staged build live by passing its `deployment_id`):
+
+```json
+{ "operation": "deploy_component", "project": "my-app", "deployment_id": "a3f8c2d1...", "restart": "rolling" }
+```
+
+### `revert_component`
+
+Swaps a component's live version back to its **retained previous version** across the cluster, then restarts. Every `deploy_component` activation retains the version it replaced (one previous version is kept per component), so `revert_component` is a fast rollback that does not re-fetch or re-install. The swap is bidirectional — reverting a revert rolls forward again.
```
cb1kenobi, line 722: **Medium: `revert_component` won't reconverge a new peer to the reverted version**

`revertApplication` only swaps directories on disk — `writeComponentRootConfig` is called from the deploy/activate paths only, never from `revertComponent`. Boot-time `installApplications` installs strictly from root config's `package:` reference, so a node that joins the cluster (or has its components directory rebuilt) *after* a revert installs the newer, reverted-away version, not the one every existing node just reverted to. That's a gap in the "across the cluster" framing here — revert is a live-instance swap, not a config-level rollback, and a not-yet-provisioned node won't pick it up.

Suggested fix: add a caveat that a node provisioned after a revert (new peer, rebuilt components dir) installs the current root-config package reference, not the reverted-to version — so revert doesn't durably change what a future node will run.

—
_Generated by Barber AI_

reference/operations-api/operations.md
```diff
@@ -705,6 +771,11 @@ Returns a list of deployment records, newest first. All filter parameters are op
 
 Response includes a `deployments` array and a `total` count. The `payload_blob` field is stripped from list responses for size; use `get_deployment_payload` to retrieve the tarball.
 
+Deployment statuses fall into two groups:
+
+- **Terminal** — `success`, `failed`, `rolled_back`, and `staged` (an `activate: false` stage-and-stop resting until it is activated or ages out of the staging-retention window).
```
cb1kenobi: **Medium: "Terminal" bucket incorrectly includes `staged`**

The implementation's own terminal-status set — `TERMINAL_STATUSES` in both `components/deploymentRecorder.ts` and `components/deploymentOperations.ts` — is `{success, failed, rolled_back}`; `staged` is explicitly excluded. That set gates `get_deployment_payload` ("... is not in a terminal state") and payload-retention pruning eligibility. Listing `staged` under "Terminal" here contradicts the unchanged `get_deployment_payload` section a few paragraphs down, which requires `success`, `failed`, or `rolled_back` — not `staged`.

Suggested fix: give `staged` its own "resting" bucket (the parenthetical already describes it that way) instead of grouping it with the true terminal statuses.

—
_Generated by Barber AI_

reference/operations-api/operations.md
```diff
@@ -718,22 +789,22 @@ Returns a single deployment record by `deployment_id`. When called on an in-prog
...
-| `completed_at`       | Timestamp when deployment finished                                      |
-| `user`               | User who initiated the deployment                                       |
-| `rollback_of`        | `deployment_id` of the deployment this rolls back, if applicable        |
-| `error`              | Error message for failed deployments                                    |
+| Field                | Description                                                                                        |
+| -------------------- | -------------------------------------------------------------------------------------------------- |
+| `deployment_id`      | Unique identifier (content hash)                                                                   |
+| `project`            | Component project name                                                                             |
+| `package_identifier` | Package reference or `payload` for tar uploads                                                     |
+| `status`             | `pending`, `staging`, `staged`, `activating`, `success`, `failed`, or `rolled_back`                |
```
cb1kenobi: **Low: `status` field list is incomplete**

The "Deployment statuses" section just above correctly lists all in-flight statuses (`pending`, `extracting`, `installing`, `staging`, `loading`, `replicating`, `activating`, `reverting`, `restarting`), but this field description only picked up `staging`/`staged`/`activating` and still omits `extracting`, `installing`, `loading`, `replicating`, `reverting`, and `restarting`. Someone building tooling against `get_deployment`'s `status` field (e.g. during a revert) won't find the value documented here.

Suggested fix:
```suggestion
| `status`             | `pending`, `extracting`, `installing`, `staging`, `staged`, `loading`, `replicating`, `activating`, `reverting`, `restarting`, `success`, `failed`, or `rolled_back` |
```

—
_Generated by Barber AI_

reference/operations-api/operations.md
```diff
@@ -673,26 +690,75 @@ Response:
 ```json
 {
 	"deployment_id": "a3f8c2d1...",
-	"message": "Component deployed successfully"
+	"message": "Successfully deployed: my-app"
```
cb1kenobi: **Low: example response doesn't match the request's `restart: "rolling"`**

The request right above sets `"restart": "rolling"`, but per `deployComponentTwoPhase`'s rolling-restart branch the response includes a `restartJobId` and a message ending in `, restarting Harper` (`"Successfully deployed: my-app, restarting Harper"`). The plain `"Successfully deployed: my-app"` shown here is only returned when no restart was requested.

Suggested fix: either drop `"restart": "rolling"` from the request to match this response, or update the response to include `restartJobId` and the `, restarting Harper` suffix.

—
_Generated by Barber AI_

reference/cli/operations-api-commands.md
```diff
@@ -212,6 +219,30 @@ harper get_components
...
+**Roll a component back to its previous version**:
+
+```bash
+harper revert project=my-app
+```
+
+**Mint a deploy key for a private repository** (Harper generates the ed25519 keypair server-side and returns only the public half — register it with your git host):
+
+```bash
+harper add_ssh_key name=my-key generate=true host=my-key.github.com hostname=github.com
```
cb1kenobi, line 243: **Low: `add_ssh_key generate` documents an untracked cross-repo dependency**

This example (and the mirrored section in `operations.md`) documents `generate: true`, implemented in `HarperFast/harper-pro#594` — a separate, still-open draft PR. This PR's own "hold merge until #1849 lands" note only tracks `HarperFast/harper#1849`; it doesn't mention harper-pro#594. If #1849 merges (and this docs PR follows) before harper-pro#594 ships, these docs advertise an operation that isn't yet available.

Suggested fix: add harper-pro#594 to the merge-gating note in the PR description, or split this section into a follow-up PR tied to that dependency.

—
_Generated by Barber AI_

reference/operations-api/operations.md
dawsontoth: Fixed at the source rather than as a caveat: harper#1849 (525dfd76d) now recovers the staged deployment's `package_identifier` and credential references at activation and persists root config on the origin and every peer, so an activate-by-id no longer leaves a stale package reference. Updated the `deployment_id` bullet to say so explicitly — you don't need to repeat `package`.

reference/operations-api/operations.md
dawsontoth: Good catch — verified in code (`revertComponent` never calls `writeComponentRootConfig`, and `installApplications()` installs from root config's `package:`). Added a caution to the `revert_component` section: reverting swaps live directories only, so a node provisioned after a revert installs the reverted-away version; deploy the older `package` explicitly for a durable rollback.

reference/operations-api/operations.md
dawsontoth: Fixed — `staged` now has its own **Resting** bucket, with Terminal limited to `success`/`failed`/`rolled_back` to match `TERMINAL_STATUSES`, and a note on why a resting deployment's payload is deliberately still held (it's the source the pending activation needs).

reference/operations-api/operations.md
dawsontoth: Fixed — the `status` row now lists all 13 values and points at the grouped explanation under `list_deployments` so the two stay in step.

reference/operations-api/operations.md
dawsontoth: Fixed — the response now shows the rolling shape (`restartJobId` plus the `, restarting Harper` suffix), with a line noting what the no-restart response looks like instead.

reference/cli/operations-api-commands.md
dawsontoth: Not gating this, per the repo owner: harper-pro#594 is planned to merge and release ahead of these docs, so the sequencing is handled outside the PR description. Flagging the dependency here for traceability — #594 is rebased, green, and carries the `generate` implementation these docs describe.

reference/operations-api/operations.md
```diff
@@ -602,10 +602,27 @@ Detailed documentation: [Components Overview](../components/overview.md)
 
 Deploys a component. The `package` option accepts any valid NPM reference including GitHub repos (`HarperDB/app#semver:v1.0.0`), tarballs, or NPM packages. The `payload` option accepts a base64-encoded tar string from `package_component`. Supports `"replicated": true` and `"restart": true` or `"restart": "rolling"`.
 
+Across a cluster, `deploy_component` runs in two phases so a deploy is all-or-nothing at go-live:
```
kriszyp, line 605: `deploy_component` is not actually all-or-nothing at go-live. The stage barrier prevents fetch/install failures from touching live directories, but activation still occurs per node; if one swap fails, other nodes can already be live, and `revert_on_failure` is off by default. Please describe this as an all-nodes staging barrier (or reduced-risk activation), and explicitly say that an activation failure can split versions unless automatic rollback is enabled.

reference/operations-api/operations.md
```diff
@@ -602,10 +602,27 @@ Detailed documentation: [Components Overview](../components/overview.md)
...
+:::
+
 Additional parameters:
 
 - `urlPath` — override the HTTP URL path the component is mounted at (e.g. `"/api/v2"`)
 - `install_allow_scripts` — set to `true` to allow npm pre/post install scripts (disabled by default)
+- `activate` — set to `false` to **stage only** and stop before go-live. The build is prepared and verified on every node and the response returns a `deployment_id` in a `staged` state; nothing goes live. Activate it later by calling `deploy_component` again with that `deployment_id` (see below). Useful for pre-staging a release and flipping it live in a separate, fast step.
```
kriszyp, line 620: Please define and enforce the deployment modes before presenting these as independent knobs. The companion core validation currently permits `deployment_id` with `activate: false` or with a fresh `package`/`payload`; dispatch prioritizes `deployment_id`, so the request activates despite `activate: false` and ignores fresh build input. Likewise, `two_phase: false` plus `activate: false`—or any staged request when `system` is excluded from replication—takes the legacy one-shot path and goes live. Please reject conflicting/unsupported combinations in core, then document the valid matrix; a request that says stage-only should never activate.

reference/components/applications.md
```diff
@@ -130,6 +130,79 @@ Harper generates a `package.json` from component configurations and uses a form
...
+
+```sh
+# Deploy a specific tag
+harper deploy ref=v1.2.0 restart=true replicated=true
+
+# Roll back by deploying an older commit
+harper deploy ref=9f8c2a1 restart=true replicated=true
+```
+
+**A reference is pinned to a SHA, not to the name you typed.** Tags and branches are resolved locally and the full commit SHA is what ships. This matters on a cluster: peers resolve the package independently, so a tag that moves mid-deploy — or a branch that advances — could otherwise leave nodes running different code.
```
kriszyp, line 163: The companion CLI does not currently provide this guarantee for an explicit `ref`: `resolveGitCommittish` returns the supplied string unchanged, so `ref=v1.2.0` produces a package ending in `#v1.2.0` rather than a full SHA. Each peer can therefore resolve a mutable tag or branch independently, recreating the mixed-version risk this paragraph says is eliminated. Please resolve explicit refs locally to a commit (for example, `git rev-parse <ref>^{commit}`) before building the package, or require a full SHA and qualify the docs accordingly.

reference/operations-api/operations.md
```diff
@@ -668,31 +685,89 @@ The response includes a `deployment_id` that can be used to query the deployment
...
+
+**Activation request** (take the staged build live by passing its `deployment_id`):
+
+```json
+{ "operation": "deploy_component", "project": "my-app", "deployment_id": "a3f8c2d1...", "restart": "rolling" }
+```
+
+### `revert_component`
+
+Swaps a component's live version back to its **retained previous version** across the cluster, then restarts. Every `deploy_component` activation retains the version it replaced (one previous version is kept per component), so `revert_component` is a fast rollback that does not re-fetch or re-install. The swap is bidirectional — reverting a revert rolls forward again.
```
kriszyp, line 722: This is narrower than documented. The retained `.deploy-previous` directory is created by two-phase activation; the legacy one-shot path used by `two_phase: false` (and by the no-system-replication fallback) does not retain it, so a later `revert_component` fails even after repeated deploys. Restart is also optional in the implementation. Please scope rollback to versions activated through the two-phase path and say it restarts only with `restart: true`/`"rolling"`, or change core so these guarantees hold on every path.

reference/operations-api/operations.md
```diff
@@ -775,7 +856,11 @@ The deployment must be in a terminal status (`success`, `failed`, or `rolled_bac
 
 ### `add_ssh_key`
 
-Adds an SSH key (must be ed25519) for authenticating deployments from private repositories.
+Adds an SSH key (must be ed25519) for authenticating deployments from private repositories. Supply the private key with `key`, or omit it and pass `generate: true` to have Harper mint the keypair itself.
+
+The stored private key is encrypted at rest and only ever crosses the cluster as ciphertext; `list_ssh_keys` and the logs never return key material.
```
kriszyp, line 861: This absolute security guarantee does not match the Pro implementation. `sealSSHKey` deliberately falls back to storing and replicating the private key in plaintext when no secret custody is registered; it logs a warning but does not fail. Please either make `add_ssh_key` (including `generate: true`) reject that configuration, or qualify this text to say encryption/ciphertext requires configured custody and explain the fallback, so operators do not assume a guarantee they may not have.

reference/cli/authentication.md
```diff
@@ -129,6 +133,51 @@ harper deploy target=https://prod-server.com:9925 replicated=true
...
+
+```yaml
+- name: Deploy
+  run: harper deploy project=my-app restart=true replicated=true
+  env:
+    HARPER_CLI_TARGET: ${{ secrets.HARPER_CLI_TARGET }}
+    HARPER_CLI_REFRESH_TOKEN: ${{ secrets.HARPER_CLI_REFRESH_TOKEN }}
+```
+
+**Precedence**: an explicitly supplied username (arguments or `HARPER_CLI_USERNAME`/`HARPER_CLI_PASSWORD`) wins; otherwise env-var tokens are used; otherwise the token saved by `harper login`. Env-var tokens deliberately outrank the saved credentials file so a runner that has both behaves predictably. A token refreshed from an env var is held in memory for that invocation only — nothing is written to `~/.harperdb/credentials.json`.
```
kriszyp, line 171: Please name the argument form precisely here. In the companion CLI, only `auth_username=`/`auth_password=`, target-URL userinfo, or a complete username/password environment pair beat env/saved bearer tokens. Ordinary `username=`/`password=` arguments are a legacy fallback after those tokens, so a reader following Method 3 can explicitly name one account while the saved account still executes the operation. Document the dedicated `auth_*` override and clarify where the ordinary payload fields sit in precedence.

reference/components/applications.md
```diff
@@ -130,6 +130,79 @@ Harper generates a `package.json` from component configurations and uses a form
 
 For SSH-based private repos, use the [Add SSH Key](#add_ssh_key) operation to register keys first.
 
+### Deploying by Reference
+
+Available since: v5.2.0
```
kriszyp, line 135: Current reference docs use `<VersionBadge version="v5.2.0" />` directly below headings for newly added minor-version surface. Please replace the plain `Available since` lines under both new sections with badges, and add these user-facing workflows—by-reference deploys, sealed deploy credentials, and CI token authentication/`--for-ci`—to `release-notes/v5-lincoln/5.2.md` so users can discover them from the release.

reference/operations-api/operations.md
```diff
@@ -668,31 +685,89 @@ The response includes a `deployment_id` that can be used to query the deployment
...
 
+`revert_component` fails with "no previous version is retained" for a component that has only ever been deployed once (nothing to revert to).
+
+:::caution
+Reverting swaps the **live directories** on the nodes currently in the cluster; it does not rewrite the component's stored `package:` reference in `harperdb-config.yaml`. For a `package` deploy that means a revert is not a config-level rollback: a node provisioned _after_ the revert — a newly joined peer, or an existing node whose components directory is rebuilt — installs from the stored package reference at boot, so it comes up on the version you reverted away from rather than the one every other node is running.
+
+To roll back durably for a `package` deploy, deploy the older version explicitly (`deploy_component` with the previous `package` reference) instead of, or after, reverting. `revert_component` is the fast live-instance swap; an explicit deploy is what changes what a future node will install.
+:::
+
 ### Deployment Operations
```
kriszyp, line 742: Please finish the v5.2 discoverability/versioning work for this new surface: add a changed `<VersionBadge>` under `deploy_component`, added badges under `revert_component` and the `generate` subsection, add `revert_component` to the Components operation summary table, and cover these user-facing features in `release-notes/v5-lincoln/5.2.md`. The CLI table records v5.2 availability, but the API reference currently reads as though these behaviors exist throughout v5.

cb1kenobi (approved):

kriszyp (changes requested):
This is looking good, but some comments worth addressing.

Ethan-Arrowood (commented):
Holding rather than stacking another CHANGES_REQUESTED on top of @kriszyp's — his eight findings already cover the prose, all are still open, and I reproduced 1, 2, 4 and 5 in local harper/harper-pro checkouts with no disagreement. My additions are structural, and they're the actual reason for the hold.

**1. Merging this publishes four unmerged upstream features.** There is no "next" tree for reference docs — `docusaurus.config.ts` sets the reference plugin to `lastVersion: 'current'`, `includeCurrentVersion: false`, `versions: { current: { label: 'v5', path: 'v5' } }`. `reference/` *is* the published `/reference/v5` site, so merging is publishing. Upstream state as of this pass:

| Documented surface | Implementation | State |
| --- | --- | --- |
| two-phase, `revert_component`, staging retention | harper#1849 | open |
| `deploy by_ref=true` / `ref=` / `credential=` | harper#1850 | open |
| `deploy setup=true` | harper#1851 | open |
| `add_ssh_key generate: true` | harper-pro#594 | open |
| `login --for-ci`, `HARPER_CLI_REFRESH_TOKEN` | harper `main` | shipped (5.2.0) |

One of five has landed. @cb1kenobi raised harper-pro#594 and the answer was that sequencing is handled outside the PR — that doesn't scale to four dependencies, and the merge gate in the body still names only #1849.

**2. The conflict with `main` can silently drop content.** `git merge-tree origin/main 94152378` conflicts in `reference/cli/authentication.md` and `reference/cli/operations-api-commands.md`. (`operations.md` and `applications.md` auto-merge cleanly — I checked the merged blob and `main`'s new `host` badge survives.) Two hazards:

- **The ops table** is rewritten wholesale here (column re-padding plus the Users/Roles and SSH recategorisation). `main` has since added seven rows this branch has never seen: `create_backup`, `list_backups`, `verify_backup`, `restore_backup`, `delete_backup`, `purge_backups`, `get_backup`. Resolving by taking "ours" drops all seven — this needs a hand merge and re-pad, not picking a side. (The recategorisation itself is right; I confirmed `#users--roles` and `#components` both exist and the old `#certificate-management` targets for user ops were wrong.)
- **Authentication precedence is about to be stated twice.** `main` landed a canonical numbered "Authentication Precedence" section in `authentication.md` after this branch forked. This PR adds a second, prose precedence statement in the same file, so post-merge the file gives the order twice, differently — and `HARPER_CLI_REFRESH_TOKEN` never gets slotted into the canonical list at all. Folding the token rules into the numbered section instead of adding a competing paragraph also resolves @kriszyp's `auth_*` vs. ordinary `username=` point in one place. Same file re-adds the `(or CLI_TARGET_*)` bullet style that `main` rewrote away.

**3. Smaller, all new:**
- The PR body is stale. It describes two files; #616 merged into this branch on 07-29, so the diff is six. Someone reading the description alone won't know they're approving CI-credential and private-source-deploy docs.
- `deployment_stagingRetention_maxCount` and `deployment_payloadRetention_*` are documented only inline in `operations.md`. `grep -rn "payloadRetention\|stagingRetention" reference/` on `main` returns nothing, so the body's "matching how `deployment_payloadRetention_maxSize` is currently surfaced" doesn't hold — it isn't surfaced anywhere. Three operator-facing config keys with no entry in `reference/configuration/options.md`, which is where operators look.
- Title still says "(draft)" but `isDraft: false`. Same on #616. If the marker is load-bearing use GitHub's draft state; otherwise drop it.

**Two questions before I'd move off the hold:**

1. **Sequencing.** Merge only once all four upstream PRs ship in the same release, or split so the `--for-ci`/token half (already on harper `main` at 5.2.0) can land now?
2. **The `system`-replication precondition.** Independent of whether core tightens the flag matrix per @kriszyp, does the prose get this caveat now? `if (req.two_phase === false || isReplicatedExecution || !isSystemDatabaseReplicated())` takes the one-shot path, so a cluster with `system` excluded from replication silently gets one-shot deploys and no revert target while reading docs that promise both as unconditional cluster behavior.

Verified correct and not worth re-litigating: the status buckets against `TERMINAL_STATUSES`, `deployment_timeout` default 120000 vs. `DEFAULT_AWAIT_ROW_TIMEOUT_MS`, retention defaults (5 / 10 MiB / 1), the rolling-restart response carrying `restartJobId` and the `, restarting Harper` suffix, and every new internal anchor. @cb1kenobi's five findings are genuinely fixed — two at the source rather than papered over with a caveat, which was the right call.

reference/operations-api/operations.md
dawsontoth: Fixed. You were right, and the page was contradicting itself — it claimed all-or-nothing in the opening paragraph and then correctly described a version split twenty lines below.

Reframed around what the two phases actually buy: an **all-nodes staging barrier**, which eliminates the common failure (a node that cannot fetch or install) but explicitly not a per-node activation failure. The activation discussion is now its own `#### Activation failures` section rather than sitting mid-parameter-list, and the opening paragraph links to it.

`revert_on_failure` is gone entirely rather than off-by-default — core now rejects it, and the docs say why there is no automatic rollback: past the barrier, a peer reporting failure does not prove it did not activate, so reverting "the failed nodes" can split the cluster three ways instead of converging it.

reference/operations-api/operations.md
dawsontoth: Resolved at the source, which is the better outcome you were pointing at. Core now enforces the modes rather than leaving them as independent knobs: `activate: false` and `deployment_id` are rejected outright when combined with `two_phase: false`, `replicated: false`, or a cluster whose `system` database is not replicated, and `two_phase: true` is rejected without the replication to support it.

So the request that says stage-only can no longer activate. The docs now carry the valid combinations as a table under `#### Deploy modes`, including that `revert_on_failure` is refused rather than accepted and ignored.

One case the table states explicitly because it is the quiet one: a plain deploy on a cluster with `system` excluded from replication silently takes the one-shot path — which also means no rollback target. That was @Ethan-Arrowood's second question on this PR.

reference/components/applications.md
dawsontoth: Fixed at the source — the guarantee is now real. harper commits `5c85714b8`, `26cce986d` and `21043f994` made an explicit `ref=` resolve to a SHA: locally when the checkout has it, from the remote when it does not (a shallow CI clone usually has neither), and the deploy **fails closed** if neither can name a commit, rather than sending the name for peers to resolve independently — exactly the divergence you described.

There is also a namespace check now: a ref outside `refs/heads/*` and `refs/tags/*` is rejected up front, because a clone fetches only those, so `refs/pull/123/head` would resolve locally and still be uncheckoutable on the cluster.

The by-reference content has since moved to #636, which documents all of that — it ships in **v5.2.3**, so it no longer belongs behind this PR's gate on harper#1849.

reference/operations-api/operations.md
dawsontoth: Fixed. Scoped to the two-phase path with a `:::caution`, in the terms you described: the retained copy is created by the activate phase, so a component deployed with `two_phase: false` — or on a cluster where `system` is excluded from replication — has nothing to revert to however many times it has been deployed. The docs previously implied repeated deploys accumulate a rollback target.

The failure message is quoted so it is greppable ("no previous version is retained"), and the `#### Deploy modes` table now names both one-shot paths as retaining nothing.

Also documented, since it was adjacent and equally invisible: `restart` on `revert_component` is optional, so a reverted component keeps serving the rolled-away-from code until something restarts it.

reference/operations-api/operations.md
dawsontoth: Fixed by qualifying the docs rather than changing the implementation — core kept the fallback deliberately, so the docs had to stop asserting the absolute.

The page now says encryption at rest is a property of your configuration, not a guarantee of the operation: with no secret custody registered, `add_ssh_key` stores and replicates the private key in **plaintext**, logs a WARN, and succeeds. It also says custody is present by default (the file tier generates a cluster keypair on first boot), so the risk reads as the exception rather than the likely case, and it tells the reader to check the logs for that warning after adding a key.

You specifically called out `generate: true`, and that was the right instinct — it reads as though the key can never be exposed, so it carries the caveat explicitly rather than inheriting it.

That content has since moved to #636: `add_ssh_key generate` ships in **v5.2.4** (harper-pro#594 merged 2026-08-20), so it is no longer gated behind this PR.

reference/cli/authentication.md
dawsontoth: Resolved at the source, and more thoroughly than this PR could have. `main` grew a canonical numbered **Authentication Precedence** section that states exactly the distinction you drew — `auth_username=`/`auth_password=`, target-URL userinfo, and a complete environment pair outrank bearer tokens, while ordinary `username=`/`password=` are a legacy fallback *below* them — plus a dedicated "Legacy `username` and `password` Fallback" section explaining that those names are also legitimate payload fields for `add_user` and friends.

#630 then folded the token variables into that same numbered list rather than restating the order beside it, which is what this PR had been doing and what @Ethan-Arrowood flagged as the page giving the order twice, differently.

This PR now only adds the OIDC exchange as entry 7 in that canonical list.

reference/components/applications.md
dawsontoth: Both done, and the version turned out to be the interesting part.

The plain `Available since` lines are now `<VersionBadge>` standalone under their headings, per CONTRIBUTING. But the version is not `v5.2.0` — that placeholder was wrong in both directions before it was right. `git tag --contains` on the merge commits puts by-reference deploys and sealed credentials in **v5.2.3** and server-side SSH keygen in **v5.2.4**, so all three were already in users' hands while the docs sat unmerged. The badges say so, and the release notes went to `5.2.md` labelled by the patch that shipped each one, since this repo keeps one page per minor and points patches at GitHub releases — a patch-shipped feature is otherwise undiscoverable from the docs.

All of that content now lives in #636 against `main`, precisely because it is shipped and this PR is still gated on harper#1849.

The lesson is recorded in `AGENTS.md`: derive the badge from core's git tags, not from the feature branch's `package.json`, and re-check on every refresh pass — a release cut between passes invalidates a badge that was correct when written.

reference/operations-api/operations.md
dawsontoth: All four done, for the surface that remains here.

- `deploy_component` carries a **changed** badge — existing surface whose cluster behavior changed.
- `revert_component` carries an added badge, and the row it was missing from the Components summary table.
- The `generate` subsection got its badge too, then moved to #636 with the rest of the shipped SSH-keygen content.
- Release notes: a new `5.3.md` covering two-phase deploys, stage-now/activate-later, component rollback, and OIDC.

All at **v5.3.0** rather than v5.2 — two-phase is still unmerged (harper#1849) and OIDC merged 2026-08-21 but is unreleased, so v5.3.0 is confirmed by the `5-3-0.ts` upgrade directive on `main`.

Each release-note entry states its limits alongside the capability, since those are what decide whether to adopt: the barrier is not an all-or-nothing go-live, it silently degrades to one-shot without `system` replication, and rollback reaches back exactly one two-phase activation.

dawsontoth:
Closing this. Its upstream was abandoned, and the one piece that came back shipped in a different shape and is already documented on `main`.

**The two-phase work is gone.** [harper#1849](https://github.com/HarperFast/harper/pull/1849) and [harper#2301](https://github.com/HarperFast/harper/pull/2301) were both closed without merging on 2026-08-25. Checked against `harper` main: `revert_component`, `two_phase`, `to_deployment_id`, `activateStagedApplication` and `deployComponentTwoPhase` do not exist, and neither do the `staging`, `activating` or `reverting` statuses this PR documents. `DeploymentStatus` is `pending`, `extracting`, `installing`, `loading`, `replicating`, `restarting`, `success`, `failed`, `rolled_back`, `staged` — so the three-way status grouping here would have shipped three invented values. `phase` is still `prepare`/`load`/`replicate`/`restart`, not `stage`/`activate`.

`deployment_payloadRetention_maxCount` does not exist either. Only `maxSize` is real (`utility/hdbTerms.ts`), and the only `maxCount` is `stagingRetention`'s. That settles the major finding this PR's description recorded as unverified — the documented retention bounds did not match pruning behavior because half the table was a key nobody built.

**Stage-then-activate came back separately.** [harper#2605](https://github.com/HarperFast/harper/pull/2605) merged 2026-09-16 as `d256ffb21`, and `main` documents it under [Staging a build and activating it later](https://github.com/HarperFast/documentation/blob/main/reference/operations-api/operations.md) — more accurately than this branch did: `404` vs `409` refusals, `restart` rejected alongside `activate: false`, a staged build outliving later deploys, `file:` sources, and `activated_from`. Nothing here improves on it.

**Everything shipped already left**, as the description says: #636, #630 and #637 all merged.

**Two things were still worth keeping**, now in #673:

1. The `deploy_component` response example on `main` shows `"Component deployed successfully"` — a string that is not in core. This PR had the real one.
2. `restartRequired` after a deploy without a restart was not documented anywhere in `reference/`. This PR's draft of it was stale (it claimed the route-miss 404 unconditionally, and said a redeploy of a live component never sets the flag, which the v5.2.1 package-metadata condition contradicts), so #673 re-derives it from core.

If two-phase deploys and `revert_component` are revived, this branch is a starting point for the prose but should be re-verified line by line against whatever lands rather than reused.

PR #600, opened 2026-07-22
Document get_deployment_payload / delete_deployment_payload response contracts

kriszyp:
Companion to [HarperFast/harper#1898 "Implement get_deployment_payload and delete_deployment_payload operations"](https://github.com/HarperFast/harper/pull/1898), which implements the two operations (documented since 5.1 but returning `Operation not found` — [harper#1893](https://github.com/HarperFast/harper/issues/1893)).

Adds the previously-unspecified behavioral contracts:

- `get_deployment_payload` — response is raw tarball bytes (octet-stream + download filename), never JSON/base64; 404 when the payload was reclaimed.
- `delete_deployment_payload` — cluster-wide reclaim note, terminal-status requirement (409 otherwise), idempotency (`freed_bytes: 0`), response shape, and the `payload_dropped` audit event.

Hold merge until harper#1898 lands.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Opus 4.8)

### DIFF
--- a/reference/operations-api/operations.md
+++ b/reference/operations-api/operations.md
@@ -697,9 +697,11 @@ Returns the raw tarball for a deployment. Useful for inspecting or re-deploying 
 }
 ```
 
+The response is the raw tarball bytes (`Content-Type: application/octet-stream`, with a `Content-Disposition` download filename) - not JSON and not base64-encoded, so payloads of any size stream without inflation. Returns `404` if the deployment does not exist or its payload has already been reclaimed (by payload retention or `delete_deployment_payload`).
+
 ### `delete_deployment_payload`
 
-Removes the tarball blob from a deployment record. The deployment record itself is retained; only the binary payload is deleted. Use this to reclaim storage after confirming a deployment is stable.
+Removes the tarball blob from a deployment record. The deployment record itself is retained; only the binary payload is deleted. Use this to reclaim storage after confirming a deployment is stable. The deletion replicates, so one call frees the payload's storage on every node in the cluster.
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


reference/operations-api/operations.md
```diff
@@ -697,9 +697,11 @@ Returns the raw tarball for a deployment. Useful for inspecting or re-deploying
 }
 ```
 
+The response is the raw tarball bytes (`Content-Type: application/octet-stream`, with a `Content-Disposition` download filename) - not JSON and not base64-encoded, so payloads of any size stream without inflation. Returns `404` if the deployment does not exist or its payload has already been reclaimed (by payload retention or `delete_deployment_payload`).
```
cb1kenobi, line 700: **Medium: get_deployment_payload's super_user requirement is stricter than delete_deployment_payload's, but the doc doesn't say so**

Both ops are registered with `requires_su: true` in `operation_authorization.ts`, and that map has a Gate-2 bypass: an op that is SU-only by default can still be granted to a non-super_user role via that role's `operations` allowlist. `delete_deployment_payload` (and nearly every other `super_user` op in this table) goes through that path unmodified. `get_deployment_payload` is the exception — harper#1898 adds a `requireSuperUser()` check directly inside `handleGetDeploymentPayload`, so it 403s even for a role that explicitly lists it in `operations` (matching the secrets-store ops, per that PR's own comment). An admin who grants `get_deployment_payload` the same way they successfully granted `delete_deployment_payload` will hit an unexplained 403.

This PR's stated goal is documenting "previously-unspecified behavioral contracts" for these two ops, so this asymmetry belongs here rather than staying implicit in the `super_user` table label both ops share.

Suggested fix: append a sentence to this paragraph, e.g. "Unlike most other `super_user` operations, this check is enforced directly in the handler and cannot be satisfied by granting the operation through a role's `operations` allowlist — only an actual `super_user` role can call it."

—
_Generated by Barber AI_

Ethan-Arrowood (approved):
LGTM — with harper#1898 merged, I verified the documented response contracts field-for-field against the handler source (field names, status values, error codes, the octet-stream/Content-Disposition contract, cluster-wide reclaim, idempotency, and the audit note) — all accurate.

One net-new nit to fold in: the idempotent-case `message` differs from the example shown, worth aligning. And cb1kenobi's point is real — `get_deployment_payload` self-enforces super_user (non-delegable) while `delete_deployment_payload` doesn't — worth documenting that asymmetry.

Issue #582, opened 2026-07-13
@relationship examples: singular from mismatch (querying.md) + scalar from+to mislabeled many-to-many (schema.md)


kriszyp:
Two `@relationship` example inaccuracies in the docs propagate into the generated agent rules (`HarperFast/skills`) on every regen. Surfaced during re-review of skills#70 (`docs: regenerate rules from documentation@de2aaf1`); the second regen in a row to ship a relationship-example bug (skills#69, skills#70). Fixing upstream here stops the drift.

Verified against live `~/dev/harper` `graphql.ts` / `search.ts`.

## 1. Blocker — many-to-many array example references a nonexistent attribute

`reference/rest/querying.md:227-228`:

```graphql
resellerIds: [Long] @indexed
resellers: [Reseller] @relationship(from: "resellerId")
```

The array attribute is `resellerIds` (plural) but the directive says `from: "resellerId"` (singular).

**Why it's broken:** `graphql.ts` stores directive args verbatim with no name normalization (`relationshipDefinition[arg.name.value] = arg.value.value`), and `search.ts` resolves the array many-to-many path by literal property lookup — `record[attribute.relationship.from]?.filter?.(...)` (`search.ts:578`) and `{ attribute: attribute.relationship.from, value: relatedEntry }` (`search.ts:187`). With `from: "resellerId"` that reads `record.resellerId`, which is undefined since the schema only defines `resellerIds`. An agent copying this gets a relationship that silently resolves to nothing.

**Fix:** `from: "resellerIds"`.

## 2. Significant — scalar `from`+`to` join mislabeled as many-to-many

`reference/database/schema.md:398-407`:

```graphql
### `@relationship(from: attribute, to: attribute)` — foreign key to foreign key
...
This is useful for many-to-many relationships that join on non-primary-key attributes.
...
product: Product @relationship(from: productSku, to: sku) # join on sku, not primary key
```

Both `productSku` and `sku` are scalar fields — that join is many-to-**one**, not many-to-many. Cardinality in Harper is decided by whether the *referenced* attribute is an array: `search.ts:180` sets `isManyToMany = Boolean(findAttribute(relatedTable.attributes, attribute.relationship.to)?.elements)`. A scalar `from`/`to` join can't be many-to-many. The correct many-to-many-via-array pattern is already shown in the same file at the `@relationship(from: arrayAttribute)` section (`networks: [Network] @relationship(from: networkIds)`).

[...]

PR #602, opened 2026-07-22
docs(schema): correct from+to relationship cardinality to many-to-one

kriszyp:
## What
Fixes the second bug from #582: the `@relationship(from: attribute, to: attribute)` example in `reference/database/schema.md` joins two scalar fields (`productSku` -> `sku`) but was described as "useful for many-to-many relationships." Reworded the section to state that cardinality follows the `to` attribute (scalar -> many-to-one, array -> many-to-many via the array pattern already shown in the `from`-only section above), and relabeled the example's inline comment as many-to-one.

## Why
Per #582: Harper's `search.ts` decides many-to-many purely by whether the *target* `to` attribute is an array (`isManyToMany = Boolean(findAttribute(relatedTable.attributes, attribute.relationship.to)?.elements)`). A scalar-to-scalar `from`+`to` join can never be many-to-many, so the old text mislabeled the example and could mislead readers (and downstream agent-rules generation in `HarperFast/skills`) into modeling it wrong.

The issue's first finding (`reference/rest/querying.md` pointing `from` at singular `resellerId` instead of `resellerIds`) was already fixed on `main` by 7e373fa3 prior to this dispatch — no change needed here.

## Test plan
- [x] `npx prettier --check reference/database/schema.md reference/rest/querying.md` — passes
- [x] Verified against `~/dev/harper`'s `search.ts` `isManyToMany` logic per the issue's citation

Refs #582

### DIFF
--- a/reference/database/schema.md
+++ b/reference/database/schema.md
@@ -397,14 +397,14 @@ type Network @table @export {
 
 ### `@relationship(from: attribute, to: attribute)` — foreign key to foreign key
 
-Both `from` and `to` can be specified together to define a relationship where neither side uses the primary key — a foreign key to foreign key join. This is useful for many-to-many relationships that join on non-primary-key attributes.
+Both `from` and `to` can be specified together to define a relationship where neither side uses the primary key — a foreign key to foreign key join. Cardinality follows the same rule as the `to`-only form above: a scalar `to` attribute gives a many-to-one join (as below); an array `to` attribute gives a many-to-many join, using the array pattern shown in `@relationship(from: attribute)` above.
 
 ```graphql
 type OrderItem @table @export {
 	id: Long @primaryKey
 	orderId: Long @indexed
 	productSku: Long @indexed
-	product: Product @relationship(from: productSku, to: sku) # join on sku, not primary key
+	product: Product @relationship(from: productSku, to: sku) # many-to-one join on sku, not primary key
 }
 
 type Product @table @export {
--- a/reference/database/schema.md
+++ b/reference/database/schema.md
@@ -397,7 +397,7 @@ type Network @table @export {
 
 ### `@relationship(from: attribute, to: attribute)` — foreign key to foreign key
 
-Both `from` and `to` can be specified together to define a relationship where neither side uses the primary key — a foreign key to foreign key join. Cardinality follows the same rule as the `to`-only form above: a scalar `to` attribute gives a many-to-one join (as below); an array `to` attribute gives a many-to-many join, using the array pattern shown in `@relationship(from: attribute)` above.
+Both `from` and `to` can be specified together to define a relationship where neither side uses the primary key — a foreign key to foreign key join. Cardinality is determined by the target (`to`) attribute: a scalar `to` attribute gives a many-to-one join (as below), while an array `to` attribute gives a many-to-many join.
 
 ```graphql
 type OrderItem @table @export {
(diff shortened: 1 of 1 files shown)


reference/database/schema.md
```diff
@@ -397,14 +397,14 @@ type Network @table @export {
...
-Both `from` and `to` can be specified together to define a relationship where neither side uses the primary key — a foreign key to foreign key join. This is useful for many-to-many relationships that join on non-primary-key attributes.
+Both `from` and `to` can be specified together to define a relationship where neither side uses the primary key — a foreign key to foreign key join. Cardinality is determined by the target (`to`) attribute: a scalar `to` attribute gives a many-to-one join (as below), while an array `to` attribute gives a many-to-many join.
 
 ```graphql
 type OrderItem @table @export {
 	id: Long @primaryKey
 	orderId: Long @indexed
 	productSku: Long @indexed
-	product: Product @relationship(from: productSku, to: sku) # join on sku, not primary key
+	product: Product @relationship(from: productSku, to: sku) # many-to-one join on sku, not primary key
```
cb1kenobi: **High: the corrected example does not actually work in Harper**

Relabeling the join as many-to-one is directionally right for the *intended* cardinality, but the schema as written doesn't resolve at all on current `main` (`513f87471`), regardless of label. `Table.ts`'s attribute-wiring (`updatedAttributes()`, around `resources/Table.ts:4414-4450`) is `if (relationship.to) { if (attribute.elements?.definition) {...} } else if (relationship.from) {...}`. Since `to: sku` is set, it takes the *first* branch unconditionally — and because `product: Product` is a scalar field (no `attribute.elements`), that branch's `else` fires: `console.error('...must have an array type referencing a table as the elements')`, and `attribute.resolve` is left `null`. The `else if (relationship.from)` branch — the one that actually knows how to resolve a single scalar relationship — is never reached when `to` is also set.

I verified this empirically against a live instance built from `origin/main` with this exact schema: `GET /OrderItem/:id` (no select) silently omits `product` from the response entirely (not even `null`), and `GET /OrderItem/:id?select(...,product{...})` (the standard way to expand a relationship, as used elsewhere in these docs/tests) 500s with `TypeError: Cannot read properties of undefined (reading 'propertyResolvers')` in `Resource.ts` (`transformForSelect`), because it dereferences `attribute.resolve.definition` on a `null` resolve.

The combined `from`+`to` resolver path only registers when the field type is an **array** (`attribute.elements?.definition`) — there is no code path that resolves a *scalar* foreign-key-to-foreign-key relationship by a non-primary-key `to` attribute. So the fix that actually makes this example work is `products: [Product] @relationship(from: productSku, to: sku)` (array field), not a comment/prose relabel — and once it's array-typed, it's structurally a one-to-many/many-to-many resolver in the code regardless of how unique `sku` is in practice.

This is a pre-existing bug in the example (not introduced by this PR), but the PR's stated purpose is specifically to make this example accurate, and its test plan only checked `search.ts`'s `isManyToMany` (the query-filter cardinality calc), not `Table.ts`'s resolver wiring (the code path that determines whether the field resolves at all). Worth fixing before merge so the docs don't keep shipping a schema that 500s when a reader follows it.

Suggested fix: change the field to an array type (`products: [Product] @relationship(from: productSku, to: sku)`) and update the surrounding prose (line 400) accordingly, or drop the `to` argument and this section's non-primary-key framing until Harper actually supports scalar-target foreign-key-to-foreign-key resolution.

—
_Generated by Barber AI_

cb1kenobi: This issue sounds legit.

cb1kenobi (approved):

Ethan-Arrowood (approved):
Verified against the resolver in Table.ts on main — the corrected array-typed `to` example matches actual behavior. One nit: title still says "many-to-one".

PR #605, opened 2026-07-24
docs(resources): document static properties as a first-class MCP/OpenAPI schema source

### LINKED ISSUE #1923: [docs] Document programmatic Resource static properties for MCP/OpenAPI schema authoring
Follow-up from #1920 / PR #1921.

`static properties` (a `Record<string, JsonSchemaFragment>`) was introduced in #1167 as the canonical class-level metadata API. PR #1921 extends it so a **programmatic** Resource declaring a bare `static properties` literal (no table-backed `attributes` Array) now drives rich MCP tool `inputSchema`s, the OpenAPI document, and the `harper://schema` introspection resource — types, per-property descriptions, `enum`/`format`/`const`, arrays, and nested objects.

The HarperFast/documentation site should document this authoring path:
- Declaring `static description` + `static properties` on a custom Resource for quality MCP/OpenAPI output (with a JSON-Schema-fragment example incl. enum/nested).
- Note that `static properties` uses JSON Schema vocabulary (lowercase types), distinct from `Table.attributes`.
- Cross-link to the MCP tool-selection metadata docs from #1095/#1167.

Needs a companion PR to HarperFast/documentation (see its AGENTS.md for versioning/mechanics).

_Issue generated by kAIle (Claude Opus 4.8)._

kylebernhardy:
Companion docs for [harper#1921 — `feat(mcp/openapi): drive schemas from a programmatic Resource's static properties`](https://github.com/HarperFast/harper/pull/1921) and [harper#1933 — `fix(resources): URL attribute-suffix routing for programmatic static-properties Resources`](https://github.com/HarperFast/harper/pull/1933). Closes [HarperFast/harper#1923](https://github.com/HarperFast/harper/issues/1923).

## What changed in core

`static properties` has been the canonical class-level metadata API since #1167, and the docs already described it — but only table-backed Resources actually got rich schemas from it. The MCP tool builder and the OpenAPI generator both read the internal `attributes` Array, so a **programmatic** Resource declaring a bare `static properties` literal produced an empty property set. 5.2.0 fixes that, and also resolves `GET /Resource/id.property` against `static properties`.

So this is less "document a new API" than "the documented API now works for the case the docs implied it did" — plus the vocabulary, constraint, and per-surface details an author actually needs.

## Changes

- **`reference/resources/resource-api.md`** — the bulk of it. Under `static properties`: which surfaces derive from it (and which explicitly don't), a **JSON Schema vocabulary** subsection, a table of every fragment key Harper reads with its per-surface behavior, resolution notes for unions / item-less arrays / optional properties / `static primaryKey`, and a nested-object + array-of-object example.
- **`learn/developers/mcp-and-openapi-metadata.mdx`** — Path B gets the emitted `tools/list` JSON, a vocabulary warning, a Path-B authoring rubric, and debugging guidance.
- **`reference/mcp/tool-metadata.md`**, **`reference/mcp/tools-and-resources.md`**, **`reference/mcp/overview.md`** — the three places that said input schemas come from `Table.attributes`.
- **`reference/rest/overview.md`**, **`reference/rest/querying.md`** — the `GET /MyTable/123.propertyName` rule, in both places it's stated.

## Review history worth knowing about

The first commit (`7fea797`) was fact-checked against the implementation and **15 of its claims were wrong**; `6cfe041` and `556106b` correct them. If you reviewed the first commit, please re-read rather than diffing. The corrections that matter most, because each is a per-surface asymmetry a reader would not guess:

- `harper://schema/{db}/{table}` is keyed by database/table, so a table-less programmatic Resource never appears there. It is not one of the surfaces.
- Only `get_*` has a record-shaped `outputSchema`. `search_*` has none; the write verbs advertise fixed `{id}` / `{ok}` / `{deleted}` envelopes. `update_*` and `patch_*` are mutually exclusive.
- Nullability: MCP re-expands to `type: ["string","null"]` and never emits a `nullable` keyword; OpenAPI 3.0.3 drops it at the top level and emits it only inside nested objects.
- `enum` / `format` / `const` and per-property `hidden` are honored at the **top level** only; inside a nested object or an array's `items`, MCP keeps them and OpenAPI drops them.
- The OpenAPI path parameter and the verb-description sentence read the class-level `static primaryKey` (default `'id'`), **not** the fragment's `primaryKey: true`. Both examples now declare both.
- Capitalized GraphQL type names (`String`, `Int`) do map correctly; the real hazard is a name in neither vocabulary (`'Text'`), where MCP coerces to `string` and OpenAPI emits `{}`.

## Open items for a human

- **Two pre-existing docs claims were false and are corrected here**, which widens this PR beyond #1923. `mcp/overview.md` and the metadata guide both said `attribute_permissions` narrows MCP input schemas per user — every derivation call site passes `undefined`; schemas are built once at registration and are caller-agnostic (RBAC filters the tool *list*, and is enforced at call time). Shout if you'd rather split those out.
- **`<VersionBadge version="v5.2.0" />` is the first v5.2.0 badge in the repo**, and there's no `release-notes/v5-lincoln/5.2.md` yet — I didn't create one for a single entry. Happy to add it once that page exists.
- **Three core behaviors documented here look like bugs, not intent** — filed separately if the team agrees: verb tools for table-less Resources are listed only to super-users (`makeVisibleTo` returns `false` without a db/table), `hidden` on a nested sub-property isn't suppressed (OpenAPI even emits `hidden: true` as a schema key), and the two surfaces disagree on unknown type names. The docs describe current behavior; if any of these get fixed, these pages need a follow-up.

Verified with `npm run build` (clean; the one reported broken anchor is pre-existing on `release-notes/v5-lincoln/5.1`) and `npm run format:check`.

_PR description generated by kAIle (Claude Opus 4.8)._


reference/resources/resource-api.md
```diff
@@ -536,7 +538,73 @@ export class ProductInventory extends Resource {
...
+				},
+				required: ['sku', 'quantity'],
+				additionalProperties: false,
+			},
+		},
+	};
+
+	async get(id) {
+		/* ... */
+	}
```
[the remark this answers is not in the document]
kylebernhardy: Good catch, and it went further than the flagged line. Fixed in 10bc616.

The instance form wasn't unique to the new example — all five examples in the "Class-level metadata for MCP and OpenAPI" section used it, while this same page's primary examples (and the Resource Static Methods section above) use `static get(target)` as the documented override point. Converted all five plus the matching example in the MCP/OpenAPI metadata guide, so the page is internally consistent rather than just the one line.

Comment generated by kAIle (Claude Opus 4.8).

dawsontoth (approved):

reference/mcp/tools-and-resources.md
```diff
@@ -72,7 +72,7 @@ The Resource's path is sanitized into a valid tool name: `/` and `.` become `_`.
 
 ### Input schema derivation
 
-Input schemas come from `Table.attributes`:
+Input schemas come from `Table.attributes` — or, for a programmatic Resource that declares no attributes, from its [`static properties`](/reference/v5/resources/resource-api#static-properties-recordstring-jsonschemafragment) <VersionBadge type="changed" version="v5.2.0" />, whose JSON Schema types pass through unchanged:
```
kriszyp, line 75: The bullets immediately below still say `attribute_permissions` narrows these schemas per requesting user, which contradicts this PR's corrected model. Registration derives every descriptor with no caller permissions, so a user whose role excludes (for example) `secret` still receives the same schema containing `secret`; the restriction is enforced when the tool runs. Only table/verb permissions affect whether the tool itself appears in `tools/list`. Please update lines 81-83 accordingly, and also reword `reference/mcp/overview.md:49`, where the current grammar still says `attribute_permissions` filters the tool list.

learn/developers/mcp-and-openapi-metadata.mdx
```diff
@@ -137,15 +141,73 @@ export class ProductInventory extends Resource {
...
+
+### Authoring rubric, Path B edition
+
+- **Describe every property.** The rubric from Path A applies verbatim — meaning over type, units and formats spelled out, short.
+- **Use `enum` wherever the value set is closed.** It's the single highest-leverage hint for an LLM: it turns "pass a status string" into "pass one of these four."
+- **Add `format`** (`date-time`, `uuid`, `email`, ...) where it applies. It reaches Swagger UI and gives the LLM a concrete shape to emit.
+- **Declare a `primaryKey` property _and_ `static primaryKey`.** The fragment flag types and describes the `id` argument on `get_*` / `update_*` / `delete_*`; the class static drives the OpenAPI path parameter and the tool-description sentence.
+- **Reach for `static outputSchemas`** when a verb returns something other than the full record — `get_*` is the only verb whose output schema is derived from `static properties`, so this is how you describe what `create_*` or a custom `search_*` actually returns.
```
kriszyp, line 209: `static outputSchemas` cannot currently describe a custom `search_*` result: the search registration path deliberately omits `outputSchema` and never reads `outputSchemas.search`. An author following this advice will add an override that has no effect in `tools/list`. Please limit this recommendation to supported verbs (such as `get`/`create`), or explicitly state that `search_*` output schemas are not supported yet.

reference/resources/resource-api.md
```diff
@@ -530,13 +532,87 @@ export class ProductInventory extends Resource {
...
+	static async get(target) {
 		/* ... */
 	}
 }
 ```
 
-For complex types and nested structures, JSON Schema vocabulary applies (`type`, `enum`, `required`, `additionalProperties`, etc.). Per-property `description` flows into both MCP `inputSchema.properties[*].description` and OpenAPI `components.schemas[*].properties[*].description`.
+As of v5.2.0, a Resource that declares `static properties` with no table backing drives the same schema derivation a table-backed Resource gets:
+
+- **MCP verb tools** — the `inputSchema` on every registered verb tool, and the `outputSchema` on `get_*` (the full record shape). `search_*` has no `outputSchema`; `create_*` / `update_*` / `patch_*` / `delete_*` advertise fixed result envelopes (`{ id }`, `{ ok }`, `{ deleted }`) that don't vary with your properties. A Resource implementing both `put` and `patch` registers `update_*` only — `patch_*` appears when it implements `patch` without `put`.
```
kriszyp, line 543: This fixed-envelope description is correct, but the `static outputSchemas` section later says that, when no override is supplied, `create`/`update`/`patch` fall back to `static properties`. They do not: only `get_*` derives its output from the record properties; the write verbs default to `{ id }` or `{ ok }` (and delete to `{ deleted }`). Please update that later fallback sentence so this page gives authors one consistent response contract.

reference/resources/resource-api.md
```diff
@@ -530,13 +532,87 @@ export class ProductInventory extends Resource {
...
+| `hidden`                                         | Suppresses a **top-level** property from MCP and OpenAPI (see [`static hidden`](#static-hidden-boolean) for the whole-class equivalent)                                                                       |
+| `assignCreatedTime`, `assignUpdatedTime`         | Marks a Harper-assigned timestamp: dropped from the MCP write-verb input schemas and marked `required` on the `get_*` output schema                                                                           |
+
+Notes on how a few of these resolve:
+
+- **Nullability is emitted differently per surface.** A `'null'` member in a union folds into the same internal flag `nullable: true` sets. MCP then re-expands it — `type: ['string', 'null']` emits `{ "type": ["string", "null"] }`, never a `nullable` keyword. OpenAPI 3.0.3 emits `{ "type": "string" }` for a top-level property (the flag isn't represented there) and `{ "type": "string", "nullable": true }` for a property inside a nested object. A union of two non-null types isn't expressible on the attribute form, so the first member wins — prefer a single type where you can.
```
Ethan-Arrowood, line 574: Two claims in this bullet are stale against #1921's head:

1. **Top-level OpenAPI `nullable` is now emitted** — the docs say it's dropped. As written, an author reads this after 5.2.0 ships with `f34cedef` and believes a working feature is missing.
2. **Multi-type unions are preserved** (MCP emits a type array, OpenAPI emits `oneOf`) — the docs say the first member wins, and the "prefer a single type" advice steers authors away from a shape that now works correctly.

Both need rewriting against the current head.

reference/resources/resource-api.md
```diff
@@ -530,13 +532,87 @@ export class ProductInventory extends Resource {
...
+- **`const` is spelled differently per surface.** Harper's OpenAPI document declares 3.0.3, whose Schema Object is the JSON Schema draft-04 subset — `const` only arrived in draft-06, so it is not a keyword there. MCP receives `const` as written; OpenAPI receives the equivalent single-value `enum`. `{ kind: { type: 'string', const: 'order' } }` reaches Swagger UI as `{ "type": "string", "enum": ["order"] }`. The constraint is identical either way; only the spelling changes. For the same reason a bare `type: 'null'` becomes an untyped `nullable` schema in the OpenAPI document, since 3.0 has no `null` type.
+- `{ type: 'array' }` with no `items` is valid and means "array of anything"; it emits a bare array schema rather than guessing an element type.
+- `hidden` and `assignCreatedTime` / `assignUpdatedTime` are honored only at the top level, and the timestamp keys only by MCP — OpenAPI request bodies reference one shared component schema per resource, so a Harper-assigned property still appears there.
```
Ethan-Arrowood, line 577: Also stale: nested `hidden` **is** honored in the OpenAPI static-properties path at #1921's head. This needs a per-surface split, since MCP nested still leaks (that's #1941) — as one combined statement it's wrong for OpenAPI and right for MCP.

reference/resources/resource-api.md
```diff
@@ -530,13 +532,87 @@ export class ProductInventory extends Resource {
...
+
+- **Nullability is emitted differently per surface.** A `'null'` member in a union folds into the same internal flag `nullable: true` sets. MCP then re-expands it — `type: ['string', 'null']` emits `{ "type": ["string", "null"] }`, never a `nullable` keyword. OpenAPI 3.0.3 emits `{ "type": "string" }` for a top-level property (the flag isn't represented there) and `{ "type": "string", "nullable": true }` for a property inside a nested object. A union of two non-null types isn't expressible on the attribute form, so the first member wins — prefer a single type where you can.
+- **`const` is spelled differently per surface.** Harper's OpenAPI document declares 3.0.3, whose Schema Object is the JSON Schema draft-04 subset — `const` only arrived in draft-06, so it is not a keyword there. MCP receives `const` as written; OpenAPI receives the equivalent single-value `enum`. `{ kind: { type: 'string', const: 'order' } }` reaches Swagger UI as `{ "type": "string", "enum": ["order"] }`. The constraint is identical either way; only the spelling changes. For the same reason a bare `type: 'null'` becomes an untyped `nullable` schema in the OpenAPI document, since 3.0 has no `null` type.
```
Ethan-Arrowood, line 575: Related precision fix while you're in here: `type: 'null'` now emits `{ nullable: true, enum: [null] }` at every depth (`d1065816`/`eaa755b4`), not just "an untyped `nullable` schema". Worth folding into the nullability refresh above.

Ethan-Arrowood (changes requested):
The writing here is precise and the Path-A/Path-B framing is genuinely useful — my concern is purely that the facts have moved. You refreshed the docs the same day as #1921's `d1065816` (the `const`→`enum` fix), but #1921 has landed four commits since, and three of the divergence bullets are now wrong in the reader's favor — they warn authors away from shapes that work.

Requesting changes on the stale bullets (inline). Separately, a merge-ordering question that may make some of this text moot: harper#1944 removes the #1941/#1942 divergences these v5.2.0-badged sections document. If #1944 lands in 5.2.0, these passages need a same-release rewrite — so it's worth deciding now whether to write them against the unified emitter instead. (I reviewed #1944 today; it still has an outstanding union-narrowing blocker.)

Two smaller notes: this is the repo's first `v5.2.0` badge and no `release-notes/v5-lincoln/5.2` page exists yet — you flagged that yourself; worth tracking so the 5.2 entry lands with the feature. And an OSPO thought: the Path-B authoring rubric is agent-facing guidance, but `harper-best-practices` doesn't cover `static properties` schema authoring at all — worth a skill section once the 5.2.0 semantics are final.

Is a final fact-check pass against #1921's merge SHA already planned? That would catch this class automatically.

kriszyp (commented):
Looks good, some fixes to apply.

reference/mcp/tools-and-resources.md
Ethan-Arrowood: Fixed in c35e4c57, and you were right about the scope — it needed a third page.

Verified against v5.2.6: every derivation call site in `components/mcp/tools/application.ts` passes `undefined` for permissions (`deriveGetSchema(attributes, undefined)` and so on for all six verbs), so the schema a restricted user receives is byte-identical to a super-user's.

- **`tools-and-resources.md`** — the "narrow the schema **per requesting user**" bullet is gone, along with the "schema narrowing is a UX optimization" paragraph that followed it. Replaced with: schemas are derived once at registration with no caller permissions in scope; `attribute_permissions` is enforced when the tool runs; and what RBAC filters is the tool _list_, via table-level permissions (`read`/`describe` for `get_*`/`search_*`, `insert`/`update`/`delete` for the writes — that's `makeVisibleTo`).
- **`overview.md`** — same correction on the one-line version. I also dropped the "`attribute_permissions` ... filters which tools `tools/list` returns" phrasing the earlier commit introduced, since that attributes the list filtering to the wrong permission type.
- **`tool-metadata.md`** — the `inputSchema` row said "and the caller's `attribute_permissions`"; now "once at registration time and identical for every caller".
- **`mcp-and-openapi-metadata.mdx`** — the RBAC section said `attribute_permissions` already filters the schema per-user. Rewritten.

Since the schema is caller-agnostic, each page now also says to treat attribute names and descriptions in a descriptor as visible to any authenticated caller who can see the tool.

learn/developers/mcp-and-openapi-metadata.mdx
Ethan-Arrowood: Correct, and it's now stated as a contract rather than a recommendation.

The Path-B rubric bullet that gave this advice is gone — see the top-level comment for why the whole Path-B section came out (c35e4c57) — but the underlying fact needed to live somewhere, so `resource-api.md`'s `static outputSchemas` section now carries a per-verb table, and it says outright that `outputSchemas.search` is not read at all and a `search` entry has no effect on `tools/list`. Confirmed against v5.2.6: the `search_*` registration passes only `inputSchema`, with no `outputSchema` key and no `overrideOutput('search')` call.

reference/resources/resource-api.md
Ethan-Arrowood: Fixed in c35e4c57. The sentence was wrong on both halves, so it's now a table:

| Verb | Output schema with no override |
| --- | --- |
| `get_*` | record shape from the Resource's attributes |
| `create_*` | `{ id }`, typed by the primary-key attribute |
| `update_*`, `patch_*` | `{ ok }` |
| `delete_*` | `{ deleted }` |
| `search_*` | none |

Two corrections beyond what you flagged: `delete_*` is `{ deleted }` alone — `deriveDeleteOutputSchema` builds a single boolean property with `additionalProperties: false`, so the `<pk>` the old text promised is not there. And since only `get_*` is record-shaped, the section now says `static outputSchemas` is the only way to describe what a write verb returns.

I also added the case this PR was originally about: a programmatic Resource with no attributes gets an empty `get_*` record schema, because the derivation reads the internal `attributes` Array rather than `static properties` (harper#1923). Until that closes, the guidance is to describe the result with `static outputSchemas.get`.

reference/resources/resource-api.md
Ethan-Arrowood: Withdrawing this — the bullet is gone rather than rewritten, and my review premise was wrong about which release we were tracking.

I checked #1921 against the tags rather than against its own head: its merge into main hasn't happened (still open, changes requested, head `eaa755b4`), and v5.2.0 through v5.2.6 all shipped without it. At v5.2.6 `components/mcp/tools/application.ts` still reads `const attributes = (ResourceClass?.attributes ?? [])`, where #1921's head reads `resolveAttributes(ResourceClass)`. So the entire v5.2.0-badged section these bullets lived in was describing unreleased behavior, and fact-checking the bullets against #1921's head — which is what I asked for — would have produced text that was still wrong for every shipped release.

It's removed in c35e4c57 and comes back, badged against whichever release it actually ships in, once #1921 lands. Details in the top-level comment.

reference/resources/resource-api.md
Ethan-Arrowood: Same disposition as the nullability thread — removed rather than split, since the passage documented unreleased behavior. Once #1921 lands, this needs the per-surface split you described (or none at all, if harper#1944 lands with it and unifies the emitters). Worth re-deriving from the merged code at that point rather than from either PR's head.

reference/resources/resource-api.md
Ethan-Arrowood: Also removed with the surrounding section. Noting the fact here so it isn't lost when this content is rewritten: at #1921's head, `type: 'null'` emits `{ nullable: true, enum: [null] }` at every depth (`d1065816` / `eaa755b4`), not the untyped `nullable` schema the old text described.

Ethan-Arrowood:
Filed [#640](https://github.com/HarperFast/documentation/issues/640) (Task, `content`, on the OSPO board) to re-add the removed sections once harper#1921 lands. It carries the verification commands, the per-file list of what to restore, the badge caution (not `v5.2.0` — and no `v5.3` milestone exists in this repo yet), and the harper#1944 caveat that the divergence content should be written against the unified emitter rather than restored as-is if #1944 lands with it. `cfdbab61` is still an ancestor of this branch, so `git show cfdbab61:<file>` recovers the text verbatim.

Also force-pushed to correct the committer email on my commits (`ethan@harperdb.io` -> `ethan@arrowood.dev`), which is why the SHAs changed — content is byte-identical. They now show as verified.

This is ready for review as far as I can take it: all checks pass and it is `MERGEABLE`. The only thing blocking is my own `CHANGES_REQUESTED` review from July 29, which was about bullets that no longer exist — I have not dismissed it myself since I wrote the replacement content.

reference/resources/resource-api.md
```diff
@@ -540,7 +540,7 @@ export class ProductInventory extends Resource {
 		},
 	};
 
-	async get(id) {
+	static async get(target) {
 		/* ... */
 	}
 }
```
kriszyp: This still states the behavior that the new caveat at `reference/resources/resource-api.md:583` says is not shipped: for a purely programmatic Resource, MCP and OpenAPI read the internal `attributes` array rather than `static properties`, so these per-property descriptions do not reach either surface. An author following this reference will add `static properties` and still receive empty MCP/OpenAPI property schemas. Please scope this sentence to table-backed attributes (which are derived from GraphQL), or state that `static properties` currently drives REST only until harper#1923 lands.

<sub>(`reference/resources/resource-api.md:549` is not part of this PR's diff — anchored to the nearest line this PR's diff can hold)</sub>

reference/mcp/tool-metadata.md
```diff
@@ -48,15 +48,15 @@ Operations registered outside core (for example, `cluster_status` from harper-pr
...
+| `annotations.readOnlyHint`    | `true` on `get_*` and `search_*`                                                                                                                                                              |
+| `annotations.destructiveHint` | `true` on `delete_*`                                                                                                                                                                          |
+| `annotations.idempotentHint`  | `true` on `update_*` (PUT semantics); other verbs default off                                                                                                                                 |
 
 `static description` and `static properties` on the Resource class override the auto-derived values. `static outputSchemas[verb]` overrides per-verb output schemas. `static mcp.annotations[verb]` overrides annotations per verb. `static hidden === true` suppresses the entire Resource from MCP listing.
```
kriszyp: Please remove or qualify `static properties` here. The table immediately above correctly says MCP schemas derive from `ResourceClass.attributes`; for a programmatic Resource, `static properties` therefore cannot override descriptor properties. Keeping it in this override list reintroduces the v5.2 behavior claim that the Path B correction just removed. `static description`, `static outputSchemas`, annotations, and class-level `static hidden` can remain.

learn/developers/mcp-and-openapi-metadata.mdx
```diff
@@ -190,13 +192,13 @@ type Customer @table @export {
```
kriszyp: The per-property `hidden: true` alternative is not current MCP/OpenAPI behavior. The preceding caveat at `learn/developers/mcp-and-openapi-metadata.mdx:149` says those surfaces ignore a programmatic Resource's `static properties`; likewise, overriding a table Resource's `properties` does not alter the `attributes` MCP reads. That can leave a field advertised after an author follows this security guidance. Keep `static hidden = true` for whole-resource suppression, but direct field-level suppression to GraphQL `@hidden` until harper#1923 ships.

<sub>(`learn/developers/mcp-and-openapi-metadata.mdx:189` is not part of this PR's diff — anchored to the nearest line this PR's diff can hold)</sub>

learn/developers/mcp-and-openapi-metadata.mdx
```diff
@@ -101,7 +101,7 @@ MCP `tools/list` now returns:
 }
 ```
 
-And `/openapi.json` picks up the same data: schema-level `description`, per-property `description`, and prepended path-level descriptions for every verb on `/Product`.
+And `/openapi` picks up the same data: schema-level `description`, per-property `description`, and prepended path-level descriptions for every verb on `/Product`.
```
dawsontoth: Nit, admittedly tangential: maybe add `/Product/`. The `/Product` endpoint returns metadata, but without reading our docs / skills, it's really easy to assume that responds with a GET list. Most people won't want the metadata, so really consistently suffixing...

kriszyp (approved):
LGTM

learn/developers/mcp-and-openapi-metadata.mdx
Ethan-Arrowood: Not tangential at all — you found a factual error, and following it up surfaced three more on the same page.

**`/Product` is not a path in the generated document.** `resources/openApi.ts` keys the collection as `` `/${path}/` `` (line 192) and the record as `'/' + path + '/{' + primaryKey + '}'` (line 248), so the OpenAPI doc contains `/Product/`, `/Product/{sku}`, and `/Product/{sku}.{property}` — never a bare `/Product`. So the sentence was naming a path that does not exist. It now names the real ones and says why the slash matters:

> ...prepended path-level descriptions for every verb on the generated paths — `/Product/` for the collection and `/Product/{sku}` for a single record. The trailing slash is load-bearing: a bare `/Product` returns the resource description rather than a record list, and is not itself a path in the document.

**And three stale quotes of the generated tool description.** Checking that sentence against core meant reading `verbDescription`, which at v5.2.6 composes `` `${prefix}${sentence} Runtime RBAC (${allowMethod}) enforces operation access at call time.` ``:

- `"get on resource '/Product' (table Product)."` is the pre-#1167 format. It survives in core *only as a comment* saying what it replaced; the live output is `"Fetches a single Product record by sku."` Both the prose in "Why descriptions matter" and the Before JSON block were quoting the dead format.
- **`allowGet` is not a Harper predicate.** `get` gates on `allowRead` (`VERB_TO_RBAC_METHOD.get`). That was wrong in all three places.
- The note says `operation access`, not `per-record access` — which matters beyond wording, since operation-scoped admission is exactly what the 5.2 authorization change is about.

Fixed in `b8784d08`. The page's argument is unchanged: with no docstring you still get just the mechanical sentence, still differing between resources only by table name.

PR #606, opened 2026-07-27
Fix the v5 migration guide's transaction example

kriszyp:
Adds the missing `transaction` import so the Transactions and Context example is self-contained.

Generated by Codex.


cb1kenobi (approved):

Ethan-Arrowood (approved):
Verified the fix against the harper module exports at main — the synthetic `harper` module does export `transaction` alongside `getContext`, and the callback overload matches the example's usage.

PR #612, opened 2026-07-27
Route applications by host/urlPath in the root config

kriszyp:
## Summary

Follow-up to [Document middleware host and path routing (#595)](https://github.com/HarperFast/documentation/pull/595), which is already merged.

#595 documented `host`/`urlPath` as a component-`config.yaml` setting. That described the mechanism accurately but put the routing in the wrong place: where an application is served is a deployment concern, and a value checked into the application can't be remapped per environment (the env-config overlay is root-config-only). On a root-config *application* entry those keys were also silently inert.

Core now treats the application's root-config entry as authoritative — [Route applications by host/urlPath declared in the root config (HarperFast/harper#1964)](https://github.com/HarperFast/harper/pull/1964) — so these docs lead with that placement.

## Changes

- **`reference/http/overview.md`** — mount an application from the root `harper-config.yaml`; a plugin's own `urlPath` positions it *within* the app and the mount is prefixed onto it; a root-config `host` overrides one the app shipped. Adds a "what a mount does not do" section: it is not a resource namespace, and it cannot constrain legacy Fastify routes by host.
- **`reference/http/api.md`** — `host` matching is **case-insensitive**, not case-sensitive as previously documented. Hostnames are case-insensitive (RFC 4343) and clients send them lowercased, so the old behavior made a configured `API.example.com` unmatchable; core fixed the compare. Also notes bare (unbracketed) IPv6 literals.
- **`reference/operations-api/operations.md`** — documents `host` on `deploy_component` and the accepted forms for both it and `urlPath`.
- **5.2 release notes** — lead with the application mount.

## Note

The `HttpOptions` case-sensitivity correction describes behavior that changes in harper#1964, so this shouldn't merge ahead of it.

`npm run format:check` and `npm run build` pass (one pre-existing broken anchor in the 5.1 release notes, unrelated).

_Generated by Claude Opus 5._


dawsontoth (approved):

reference/http/overview.md
```diff
@@ -30,18 +30,50 @@ Request and response objects follow the [WHATWG Fetch API](https://developer.moz
...
+An application can also set `host` per plugin, but a `host` on the root-config entry overrides it — the operator's choice of hostname wins over one the application shipped.
+
+Handlers see the request with the mount already removed from the pathname, so application code addresses itself mount-relative and does not need to know where it is mounted.
+
+#### What a mount does not do
+
+A mount is a routing prefix, not an isolation boundary:
+
+- **It does not namespace resources.** Exported tables live in one instance-wide registry, so a table exported by one application is reachable through any mounted REST route, not only its own.
+- **It does not constrain legacy Fastify routes by host.** `fastifyRoutes` registers as a global fallback outside the routed middleware chain, so those routes answer on every hostname. A `urlPath` mount does apply (it becomes the Fastify route prefix); a `host` mount does not, and Harper logs a warning when a host-mounted application declares `fastifyRoutes`. Port them to `server.http()` for host routing.
```
cb1kenobi: **Medium: "logs a warning" understates the actual behavior**

In `harper#1964`, a host-mounted application that declares `fastifyRoutes` doesn't just get a warning — `server/fastifyRoutes.ts`'s `handleApplication` throws (`Application '<name>' is mounted on host '<host>', but its fastifyRoutes cannot be constrained by host...`), which `componentLoader.ts` catches and reports via `componentLifecycle.failed(...)`. The component fails to load entirely; it doesn't continue running with a warning logged.

Someone deploying a host-mounted app with `fastifyRoutes`, reading this line, would expect degraded-but-working behavior and instead gets an app that won't start.

Suggested fix:
```suggestion
- **It does not constrain legacy Fastify routes by host.** `fastifyRoutes` registers as a global fallback outside the routed middleware chain, so those routes answer on every hostname. A `urlPath` mount does apply (it becomes the Fastify route prefix); a `host` mount does not, and Harper refuses to load a host-mounted application that declares `fastifyRoutes` rather than silently serving it unconstrained. Port them to `server.http()` for host routing.
```

—
_Generated by Barber AI_

reference/http/overview.md
```diff
@@ -30,18 +30,50 @@ Request and response objects follow the [WHATWG Fetch API](https://developer.moz
...
+
+Handlers see the request with the mount already removed from the pathname, so application code addresses itself mount-relative and does not need to know where it is mounted.
+
+#### What a mount does not do
+
+A mount is a routing prefix, not an isolation boundary:
+
+- **It does not namespace resources.** Exported tables live in one instance-wide registry, so a table exported by one application is reachable through any mounted REST route, not only its own.
+- **It does not constrain legacy Fastify routes by host.** `fastifyRoutes` registers as a global fallback outside the routed middleware chain, so those routes answer on every hostname. A `urlPath` mount does apply (it becomes the Fastify route prefix); a `host` mount does not, and Harper refuses to load a host-mounted application that declares `fastifyRoutes` rather than silently serving it unconstrained. Port them to `server.http()` for host routing.
+
```
Ethan-Arrowood, line 75: This overstates the failure scope. `fastifyRoutes.ts` does throw in #1964, but the loader's catch is **per plugin** inside the app's config loop ([`componentLoader.ts#L737`](https://github.com/HarperFast/harper/blob/3a8218c4202f499915e250649e7c7191260f1c16/components/componentLoader.ts#L737)) — it marks the `fastifyRoutes` component failed and continues; the code comment says as much ("contained to this component by the loader's per-component try/catch, so it does not take down the rest of the application"). So the app's other handlers (rest, static, …) still load and serve on the host mount; only the legacy routes fail.

As written, a reader expects the whole app to refuse to start and would misdiagnose a partially-running app — the same class of mismatch @cb1kenobi already caught once in the opposite direction (docs said "warns", code throws). Suggested wording: "Harper fails the application's `fastifyRoutes` with a load error rather than silently serving them unconstrained; the application's other handlers still load."

Ethan-Arrowood (commented):
Cross-checked against harper#1964. The substance is accurate — mount-strip semantics, per-environment remounting, the `harper deploy … host= urlPath=` example, the payload-deploy rejection note, bare-IPv6 forms in all three files, and the resource-registry caveat all verified, and correcting the previously-documented case-sensitive host matching is a good catch.

Two things before merge, one of them a hold.

**Hold behind #1964.** I reviewed that PR earlier today and raised two findings that bear directly on this text: host-mounted apps currently don't match under HTTP/2 (`matchesRoute` never reads `:authority`), and a hand-edited root-config `host` is never validated. If the h2 gap gets fixed in core, this doc needs no change; if h1-only is accepted as a design limitation, these docs must say so. Worth re-reading the two host-matching sentences (`api.md:51`, `overview.md:77`) once #1964 settles.

Minor: link style is inconsistent between two of the touched files — `operations.md:609` uses `../http/overview.md#middleware-routing` while `plugin-api.md:47` uses extensionless `../http/overview#middleware-routing`. Build passes either way; extensionless matches the surrounding files.

cb1kenobi (approved):

PR #613, opened 2026-07-27
Document CLI authentication precedence and dedicated credentials

kriszyp:
Documents the dedicated `auth_username=`/`auth_password=` parameters and revised CLI credential precedence from [HarperFast/harper#1873](https://github.com/HarperFast/harper/pull/1873), including atomic environment-variable pairs, CI/CD user-management examples, application deployment guidance, and stale-token recovery. Also adds the Harper 5.2 release-note entry and replaces the applications guide's legacy credential example.

Refs [HarperFast/harper#1872](https://github.com/HarperFast/harper/issues/1872).

Generated with Codex (GPT-5).

### DIFF
--- a/reference/cli/authentication.md
+++ b/reference/cli/authentication.md
@@ -32,6 +32,23 @@ Available since: v4.1.0; expanded in: v4.3.0
 
 For remote operations (operations executed on a remote Harper instance via the `target` parameter), you must provide authentication credentials.
 
+### Authentication Precedence
+
+<VersionBadge type="changed" version="v5.2.0" />
+
+For remote Operations API commands, the CLI uses the first complete authentication source in this order:
+
+1. Dedicated `auth_username=` and `auth_password=` command parameters
+2. Credentials embedded in the `target` URL
+3. `HARPER_CLI_USERNAME` and `HARPER_CLI_PASSWORD`
+4. Legacy `CLI_TARGET_USERNAME` and `CLI_TARGET_PASSWORD`
+5. A token saved by `harper login`
+6. `username=` and `password=` operation parameters (legacy fallback)
+
+Credentials are resolved as a pair and are never combined across sources. An incomplete pair supplied with dedicated authentication parameters or in the target URL causes the command to fail. An incomplete environment-variable pair is skipped with a warning so that a saved login token can still be used.
+
+Before v5.2.0, `username=` and `password=` operation parameters took precedence over environment variables and saved login tokens. This could authenticate an operation as the wrong user when those fields were part of the operation payload, such as the user being created by `add_user`.
+
 ### Authentication Methods
 
 #### Method 1: Persistent Login (Recommended for Local Development)
@@ -62,15 +79,23 @@ harper logout https://server.com:9925
 - Simplifies frequent remote operations
 - No need to maintain environment variables in multiple terminal sessions
 
+A complete environment-variable credential pair takes precedence over a saved login token. Check the [authentication precedence](#authentication-precedence) when a project `.env` file and `harper login` are both configured for the same target.
+
 #### Method 2: Environment Variables (Recommended for CI/CD)
+
+<VersionBadge type="changed" version="v5.2.0" />
 
 The CLI supports loading environment variables from your shell environment (or optionally from a `.env` file in the current directory). This is the recommended method for CI/CD pipelines and for pre-populating the `target` parameter for specific projects.
 
+Starting in v5.2.0, a complete environment-variable credential pair takes precedence over a saved login token and the legacy `username=` and `password=` fallback. This makes the configured CI identity authoritative even when the operation payload also contains username or password fields.
+
 **Supported Variables**:
 
-- `HARPER_CLI_TARGET` (or `CLI_TARGET`) - Sets the default `target` for CLI commands.
-- `HARPER_CLI_USERNAME` (or `CLI_TARGET_USERNAME`) - Harper admin username for the target.
-- `HARPER_CLI_PASSWORD` (or `CLI_TARGET_PASSWORD`) - Harper admin password for the target.
+- `HARPER_CLI_TARGET` - Sets the default `target` for CLI commands. `CLI_TARGET` is the legacy equivalent.
+- `HARPER_CLI_USERNAME` and `HARPER_CLI_PASSWORD` - Preferred credential pair for the target.
+- `CLI_TARGET_USERNAME` and `CLI_TARGET_PASSWORD` - Lower-priority legacy credential pair.
+
+Each credential namespace is independent. For example, the CLI never combines `HARPER_CLI_USERNAME` with `CLI_TARGET_PASSWORD`. If either namespace supplies only a username or only a password, that incomplete pair is skipped with a warning.
 
 **Example `.env` file**:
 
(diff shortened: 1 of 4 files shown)


reference/cli/authentication.md
```diff
@@ -32,6 +32,23 @@ Available since: v4.1.0; expanded in: v4.3.0
...
+### Authentication Precedence
+
+<VersionBadge type="changed" version="v5.2.0" />
+
+For remote Operations API commands, the CLI uses the first complete authentication source in this order:
+
+1. Dedicated `auth_username=` and `auth_password=` command parameters
+2. Credentials embedded in the `target` URL
+3. `HARPER_CLI_USERNAME` and `HARPER_CLI_PASSWORD`
+4. Legacy `CLI_TARGET_USERNAME` and `CLI_TARGET_PASSWORD`
```
cb1kenobi: Maybe it's overly verbose, but it might be nice to state the context for these variables.
```suggestion
3. `HARPER_CLI_USERNAME` and `HARPER_CLI_PASSWORD` environment variables
4. Legacy `CLI_TARGET_USERNAME` and `CLI_TARGET_PASSWORD` environment variables
```

cb1kenobi (approved):

reference/cli/authentication.md
```diff
@@ -172,6 +229,14 @@ target=http://localhost:9925
 target=https://server.example.com:8080
 ```
 
+The target URL can also contain a complete username and password, which the CLI ranks above environment variables:
+
+```bash
+target=https://username:password@server.example.com:9925
+```
+
+Percent-encode reserved characters in either value (`@` becomes `%40`, for example). Avoid this form when possible: the URL can be exposed in shell history and process listings, and `harper login` can save its target to the project `.env` file. Prefer environment variables or a saved login for routine use.
```
Ethan-Arrowood, line 238: This advice can't work against current harper main. WHATWG `URL` getters return userinfo **still percent-encoded** — `new URL('https://admin:p%40ss@host').password === 'p%40ss'`, and a raw `@` gets re-encoded the same way — and the CLI feeds `parsedTarget.username`/`password` verbatim into the Basic header ([`cliOperations.ts:399`](https://github.com/HarperFast/harper/blob/74224938ff7d7670c038429b4c37c3396cdd80d7/bin/cliOperations.ts#L399) → [`basicAuthHeader`](https://github.com/HarperFast/harper/blob/74224938ff7d7670c038429b4c37c3396cdd80d7/bin/cliOperations.ts#L156)) with no `decodeURIComponent` anywhere; `normalizeTarget` preserves the encoding too.

So a real password `p@ss` goes out as `p%40ss` and authentication fails — encoded or not, the URL form simply can't carry reserved-character passwords right now. I verified the `URL` behavior with Node directly. harper#1873's tests don't cover encoded userinfo.

Two ways forward: harper decodes userinfo the way curl does (feasible before 5.2 ships), or this sentence is replaced with an explicit limitation note scoping the URL form to reserved-character-free credentials. Either way it shouldn't ship as-is.

reference/cli/authentication.md
```diff
@@ -127,31 +151,64 @@ export HARPER_CLI_PASSWORD=$SECURE_PASSWORD  # from secret manager
 # Execute operations without passing credentials or target (if set)
 harper deploy target=https://prod-server.com:9925 replicated=true
 harper restart target=https://prod-server.com:9925 replicated=true
+
+# The environment variables authenticate the admin; username/password remain payload fields.
+# NEW_USER_PASSWORD is provided by a secret manager.
+harper add_user \
```
Ethan-Arrowood, line 157: Small hardening thought: the auth credentials here are properly protected, but the *new user's* `password=` still lands in argv — `"$NEW_USER_PASSWORD"` keeps it out of shell history, yet the expanded value is visible in process listings. That's inherent to CLI `add_user` rather than anything this PR introduces, but since this page is the security guidance, a one-line caveat under these examples would keep it airtight.

Ethan-Arrowood (commented):
I traced every precedence claim in this page against `bin/cliOperations.ts` on harper main — the ordering, the pre-5.2 behavior note, the `auth_`-parameter semantics, and the stale-token troubleshooting are all accurate. Nice page.

One substantive problem, inline: the percent-encoding advice for target-URL credentials describes behavior the CLI doesn't have today, so following it produces auth failures rather than fixing them. That's really an upstream harper bug, not just a docs wording issue — flagging it here because the sentence should either come out or be replaced with a limitation note until harper decodes userinfo.

reference/cli/authentication.md
DavidCockerill: Corroborating this — I verified it independently and it holds.

`extractTargetCredentials` (`bin/cliCredentials.ts:28`) returns WHATWG `URL.username`/`.password`, which stay percent-encoded, and `resolveTransportCredentials` passes them verbatim to `basicAuthHeader` with no `decodeURIComponent` anywhere on the path. Confirmed in Node: `new URL('https://admin:p%40ss@host').password === 'p%40ss'`, and a raw `@` re-encodes to the same thing.

So a password of `p@ss` goes out as the literal `p%40ss` and 401s — following this advice causes the failure it claims to prevent, and not following it fails too.

Either harper decodes userinfo before 5.2 ships, or the sentence becomes a limitation note scoping the URL form to credentials without reserved characters. I'd block on it not shipping as written rather than on which of the two fixes you pick.

reference/cli/authentication.md
```diff
@@ -32,6 +32,23 @@ Available since: v4.1.0; expanded in: v4.3.0
...
+
+<VersionBadge type="changed" version="v5.2.0" />
+
+For remote Operations API commands, the CLI uses the first complete authentication source in this order:
+
+1. Dedicated `auth_username=` and `auth_password=` command parameters
+2. Credentials embedded in the `target` URL
+3. `HARPER_CLI_USERNAME` and `HARPER_CLI_PASSWORD` environment variables
+4. Legacy `CLI_TARGET_USERNAME` and `CLI_TARGET_PASSWORD` environment variables
+5. A token saved by `harper login`
```
DavidCockerill, line 45: The list omits a rung. `cliOperations.ts` resolves **seven** tiers: after the four basic-credential sources it reads `HARPER_CLI_OPERATION_TOKEN`/`HARPER_CLI_REFRESH_TOKEN` (falling back to the `CLI_TARGET_*` namespace, chosen as a unit) and only then `~/.harperdb/credentials.json`. So the env-token tier sits between your #4 and #5.

These vars are new in 5.2 — `origin/v5.1` has no `OPERATION_TOKEN` — and `harper login --for-ci` prints `HARPER_CLI_REFRESH_TOKEN` as the durable CI secret. Yet they appear in **zero** markdown files in this repo, so this page is the only place a reader would look.

Concretely: a CI runner holding both a refresh token and `HARPER_CLI_USERNAME`/`PASSWORD` gets basic auth, not the token, and this list can't tell them that. One extra rung plus a sentence naming the vars closes it.

DavidCockerill (changes requested):
This covers all three things I asked for on harper#1873 and gets them right — I verified the precedence order, the pair-atomicity rule, the pre-5.2 note and the stale-token entry against `cliOperations.ts` rather than taking the page's word for it. Good, careful doc.

Requesting changes on two things:

1. **@Ethan-Arrowood's percent-encoding thread** — I've replied there with an independent confirmation. The URL-credentials advice can't work as written, and since this page is where people will copy from, it shouldn't ship in that form.
2. **The precedence list is missing a rung** (thread below): the `HARPER_CLI_OPERATION_TOKEN`/`HARPER_CLI_REFRESH_TOKEN` tier sits between #4 and #5, and those vars are documented nowhere else in the repo despite `--for-ci` printing one as the CI secret.

Both are small. Happy to re-review quickly.

— Reviewed by DAIvid (Claude Opus 5)

PR #614, opened 2026-07-28
docs(http): document request.connectionInfo (forwarded PROXY v2 TLS facts)

kriszyp:
Companion docs for **[harper#1985](https://github.com/HarperFast/harper/pull/1985)** (closes harper#1983).

## What
Documents the new **`request.connectionInfo`** accessor on the HTTP `Request` reference (`reference/http/api.md`): the TLS facts a fronting TLS-terminating proxy (Harper Fabric's Symphony) forwards over PROXY protocol v2 — negotiated ALPN, SNI authority, TLS version/cipher, the client's JA3/JA4 fingerprint, and the verified mTLS client certificate chain. Adds:
- a row in the `Request` **Properties** table (with `<VersionBadge version="v5.2.0" />`),
- a **Connection info** subsection with the `ConnectionInfo` TypeScript shape, a JA4 usage example, and the trust / `tls.verified`-vs-mTLS caveat,
- a clarification on the existing **`protocol`** row: it reflects the client's original scheme (`https`) behind a v2-forwarding proxy even though the proxy→Harper hop is plaintext.

## Notes for review
- `v5.2.0` badge — the feature ships in core 5.2 (currently `5.2.0-beta.2`), per the versioning convention.
- The trust caveat is deliberate: `connectionInfo.tls.verified` reflects the proxy's verify bit and can be `true` with no `clientCertChain` (proxy verified but omitted an oversized chain), so the docs steer mTLS gating to `request.peerCertificate` + the authenticated user, matching the core PR's field doc.
- `npm run format:check` clean; `npm run build` succeeds with no broken links/anchors on this page (the one broken-anchor warning is a pre-existing `/release-notes/v5-lincoln/5.1` issue, untouched here).

Generated by an LLM (Claude Opus 4.8) pairing with Kris.


reference/http/api.md
```diff
@@ -78,17 +78,53 @@ A `Request` object is passed to HTTP middleware handlers and direct static REST
...
+class Origin {
+	get(request) {
+		const ja4 = request.connectionInfo?.ja4;
+		if (ja4 && abuseList.has(ja4)) throw new Error('blocked');
+		return fetch(request);
+	}
+}
+```
+
+**Trust and mTLS.** `connectionInfo` is populated only from the trusted proxy-protocol channel (Harper Fabric's per-worker Unix domain socket, writable only by the local proxy) — never from a request header, so a client cannot forge it. `tls.verified` reflects the proxy's verify bit: it can be `true` even when `clientCertChain` is absent (a proxy may verify a certificate but omit an oversized chain). For gating on a verified mTLS client identity, use the parsed [`request.peerCertificate`](../security/certificate-verification.md) and the connection's authenticated user, not `connectionInfo.tls.verified` alone.
```
Ethan-Arrowood: This link sends readers to the wrong page: [`certificate-verification.md`](https://github.com/HarperFast/documentation/blob/main/reference/security/certificate-verification.md) covers CRL/OCSP revocation checking and contains no mention of `peerCertificate` — and nothing else in the docs repo documents `request.peerCertificate` either. The property is real (harper `server/serverHelpers/Request.ts`), but the reader following this advice gets no help. `mtls-authentication.md` at least covers the authenticated-user flow:

```suggestion
**Trust and mTLS.** `connectionInfo` is populated only from the trusted proxy-protocol channel (Harper Fabric's per-worker Unix domain socket, writable only by the local proxy) — never from a request header, so a client cannot forge it. `tls.verified` reflects the proxy's verify bit: it can be `true` even when `clientCertChain` is absent (a proxy may verify a certificate but omit an oversized chain). For gating on a verified mTLS client identity, use the parsed [`request.peerCertificate`](../security/mtls-authentication.md) and the connection's authenticated user, not `connectionInfo.tls.verified` alone.
```

Worth also documenting `peerCertificate` properly somewhere, since this is the second page to reference it without defining it.

reference/http/api.md
```diff
@@ -78,17 +78,53 @@ A `Request` object is passed to HTTP middleware handlers and direct static REST
...
+| `pathname`       | string                                                                | Path portion of the URL, without query string                                                                                                                                                                                                                                                                                        |
+| `protocol`       | string                                                                | `http` or `https`. Behind a TLS-terminating proxy that forwards PROXY protocol v2, this reflects the client's original scheme even though the proxy-to-Harper hop is plaintext.                                                                                                                                                      |
```
Ethan-Arrowood: Per AGENTS.md this row's behavior changed in 5.2, so it should carry a changed-badge (already raised by the Gemini bot — corroborating it, the convention holds elsewhere in this file):

```suggestion
| `protocol`       | string                                                                | `http` or `https`. Behind a TLS-terminating proxy that forwards PROXY protocol v2, this reflects the client's original scheme even though the proxy-to-Harper hop is plaintext. <VersionBadge type="changed" version="v5.2.0" />                                                                                                     |
```

Ethan-Arrowood (commented):
Traced every field, type, and presence condition against harper#1985 (which I approved earlier today) — the content is accurate, including the `tls.verified`-without-`clientCertChain` caveat. One real problem plus a few polish items, all with suggestions inline.

The one that matters: the `request.peerCertificate` link points at `certificate-verification.md`, which is about CRL/OCSP revocation checking and never mentions `peerCertificate` — in fact nothing in this repo documents it. Anyone following the mTLS-gating advice lands on a page that can't help them.

Merge-ordering: this should land after harper#1985, since it documents an accessor that doesn't exist on main yet. A `connectionInfo` entry in `release-notes/v5-lincoln/5.2.md` is also still missing — fine as a follow-up when #1985 lands.

On the Gemini bot's other suggestion (make the example's `get` `static`): I'd leave it as-is — it matches this page's existing instance-method examples, and I found no such rule in AGENTS.md or CONTRIBUTING.md.

One question: the core PR's own field doc recommends `request.authorized` as the standard verified-mTLS gate, while these docs recommend `peerCertificate` + the authenticated user. Intentional, or worth mentioning `request.authorized` too?

reference/http/api.md
```diff
@@ -78,17 +78,53 @@ A `Request` object is passed to HTTP middleware handlers and direct static REST
...
+		cipher?: string; // negotiated cipher suite
+		verified?: boolean; // a client certificate was presented and the proxy verified it
+	};
+	ja3?: string; // JA3 fingerprint (32-char MD5 hex)
+	ja4?: string; // JA4 fingerprint
+	clientCertChain?: Buffer[]; // verified mTLS client certificate chain (DER, leaf first)
+}
+```
+
+`request.connectionInfo` is `undefined` unless a PROXY v2 header was decoded; each field is present only when the proxy actually forwarded it. Reading the client's JA4 fingerprint — e.g. to make a bot or abuse decision — is simply:
```
Ethan-Arrowood: Two small precision fixes. The implementation leaves `connectionInfo` `undefined` not just when there's no v2 header, but also when a v2 header carries no relevant TLVs — it's only set once at least one field decodes. And it's Node-runtime only: the Bun and uWS request classes never populate it (`tls.md` sets the precedent for calling out runtime caveats).

```suggestion
`request.connectionInfo` is `undefined` unless a PROXY v2 header carrying TLS facts was decoded; each field is present only when the proxy actually forwarded it. On the Bun and uWebSockets runtimes it is always `undefined`. Reading the client's JA4 fingerprint — e.g. to make a bot or abuse decision — is simply:
```

reference/http/api.md
```diff
@@ -78,17 +78,53 @@ A `Request` object is passed to HTTP middleware handlers and direct static REST
...
+
+```javascript
+class Origin {
+	get(request) {
+		const ja4 = request.connectionInfo?.ja4;
+		if (ja4 && abuseList.has(ja4)) throw new Error('blocked');
+		return fetch(request);
+	}
+}
```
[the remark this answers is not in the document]
kriszyp: Left as-is: this file's other Resource example (`async get(request)` at what's now line ~148) also uses an instance method, so `static` would be inconsistent with the existing convention here rather than more consistent with it. Also checked AGENTS.md/CONTRIBUTING.md directly and found no static-method rule for these examples — matches what Ethan found. Not applying the suggestion.

<sub>sent with Claude Sonnet 5</sub>

reference/http/api.md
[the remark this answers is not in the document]
kriszyp: Fixed — added `<VersionBadge type="changed" version="v5.2.0" />` to the `protocol` row.

<sub>sent with Claude Sonnet 5</sub>

kriszyp: Fixed — applied your suggestion, `<VersionBadge type="changed" version="v5.2.0" />` on the `protocol` row.

<sub>sent with Claude Sonnet 5</sub>

reference/http/api.md
kriszyp: Fixed the link. Also went a step further based on an independent pre-push review pass: per `server/serverHelpers/proxyProtocol.ts`'s own docblock ("For the standard 'verified mTLS client' gate use `request.authorized`, which is chain-gated"), the whole gating recommendation was imprecise — `request.authorized` is the correct primary gate here, not `request.peerCertificate` + "the authenticated user". Also caught and fixed a real accuracy bug while I was in there: `clientCertChain` was documented as "verified," but the decoder attaches it whenever the client *presented* a cert, independent of the proxy's verify result — chain presence and `tls.verified` are separate signals, and I made that explicit in the trust paragraph, the property table, and the `ConnectionInfo` interface comment. Neither `request.authorized` nor `request.peerCertificate` has a real home in this repo yet (your point stands) — left that as a follow-up rather than expanding this PR's scope further; noted in Findings for the EM to triage.

<sub>sent with Claude Sonnet 5</sub>

reference/http/api.md
kriszyp: Fixed both — undefined condition now says "unless a PROXY v2 header carrying TLS facts was decoded" and added the Node-only (not Bun/uWS) runtime caveat, matching `tls.md`'s precedent.

<sub>sent with Claude Sonnet 5</sub>

reference/http/api.md
```diff
@@ -78,17 +78,53 @@ A `Request` object is passed to HTTP middleware handlers and direct static REST
...
+| `pathname`       | string                                                                | Path portion of the URL, without query string                                                                                                                                                                                                                                                                                                                                                                                                              |
+| `protocol`       | string                                                                | `http` or `https`. On the Node runtime, behind a TLS-terminating proxy that forwards the PROXY v2 SSL TLV, this reflects the client's original scheme even though the proxy-to-Harper hop is plaintext; a proxy forwarding only address info (no SSL TLV) still reads as `http`. Bun and uWebSockets derive this from `X-Forwarded-Proto`/the listener's secure flag instead, independent of the SSL TLV. <VersionBadge type="changed" version="v5.2.0" /> |
```
DavidCockerill, line 87: Precision nit, non-blocking. "Bun and uWebSockets derive this from `X-Forwarded-Proto`/the listener's secure flag" is true of uWS (`Request.ts:515-520` reads the header first, then falls back to `#isSecure`) but not Bun — `BunRequest.protocol` (line 401) is only `this.#isSecure ? 'https' : 'http'` and never consults the header.

So a Bun user behind a TLS-terminating proxy setting `X-Forwarded-Proto: https` still gets `http`. Suggest splitting them: "uWebSockets reads `X-Forwarded-Proto`, falling back to the listener's secure flag; Bun uses the listener's secure flag only."

Only inaccuracy I found on the page.

DavidCockerill (approved):
Approving. I traced every field, type, presence condition and runtime caveat against `proxyProtocol.ts`, `Request.ts` and `http.ts` — all accurate, including the presented-vs-verified distinction, which is the part most likely to be got wrong and is right here.

My one worry going in was merge ordering: this documents an API added by harper#1985, and a doc describing something no released version exposes is worse than no doc. That's resolved — #1985 merged 2026-07-30 into main at `5.2.0-beta.4`, so the API exists and the version gating on the page matches.

One precision nit in the thread about the Bun protocol derivation.

— Reviewed by DAIvid (Claude Opus 5)

PR #617, opened 2026-07-30
docs(deploy): by_ref ref pinning, credential=true, and CI behavior

dawsontoth:
> **Draft.** Refines the `by_ref` documentation in [#599](https://github.com/HarperFast/documentation/pull/599) to match [harper#1850](https://github.com/HarperFast/harper/pull/1850) as merged. All edits are in the `by_ref` section of `reference/components/applications.md`.

**Its own dependency has landed.** harper#1850 merged 2026-08-14 — the previous description here said it was open and approved. Everything this PR documents is now on harper `main`, verified against the merged code rather than against the PR. It carries no companion marker for that reason; sequencing is handled by its base branch.

**It shipped in v5.2.3**, not v5.3.0 as an earlier pass of this stack assumed. `git tag --contains` on harper#1850's merge commit returns v5.2.3 and v5.2.4, so `by_ref` has been in users' hands since that patch. The badge is corrected in #599, and the release note moved to `5.2.md`. Nothing here is gated on an unmerged PR.

## What changed

**Refs are pinned or the deploy stops.** Resolution covers the remote as well as the local checkout, since a shallow CI clone usually doesn't have the ref, and annotated tags resolve to the commit they point at. If a ref can't be resolved either way the deploy stops rather than sending the name — peers resolve the package independently, so an unresolvable name is exactly the divergence the SHA pin exists to prevent. A full commit SHA is always accepted, since it can't move.

**A ref must name something a clone can fetch.** New in this pass, and newer than the review: `refs/heads/*`, `refs/tags/*`, or a bare branch or tag name. Anything else — `refs/pull/123/head` being the realistic case — is rejected up front *even when the local checkout resolves it happily*, because the cluster could resolve that commit and still never check it out. Worth documenting precisely because the failure looks wrong: it works locally, so the error reads as a bug.

**`credential=true`.** The credential host is derived from the package being deployed, so it always matches the clone it authenticates. `credential=github.com` still works but is compatibility rather than a choice, and a host that doesn't match the package's is rejected — the clone would never ask for it, so the deploy would otherwise fail as though no credential were configured.

**The unpushed-commit warning is skipped under GitHub Actions.** @Ethan-Arrowood's finding on the last pass, still valid in merged `main` (`bin/cliOperations.ts:538`): the dirty-tree warning is unconditional, the unpushed-commit one is not. "Warns in both directions" read as unconditional, so an Actions user expecting the second warning silently wouldn't get it. Now stated, with the reason — the runner's checkout isn't a branch `git branch -r --contains` can see.

**What `by_ref` deploys under GitHub Actions.** On a `pull_request` run it's the pull request's head commit, not the merge commit the runner checks out — that one lives under `refs/pull/<n>/merge`, which a plain clone can't fetch. For a fork PR the head repository is the fork. Includes the `ref=${{ github.event.pull_request.head.sha }}` escape hatch the CLI points to when the event payload isn't readable.

## Review notes

- **Stacked on [#599](https://github.com/HarperFast/documentation/pull/599)**, which is now itself stacked on [#630](https://github.com/HarperFast/documentation/pull/630). This branch is a strict superset, so targeting `main` would re-show all of #599's still-unmerged content. GitHub retargets automatically as the stack lands.
- Merging #599 in also picked up its later corrections to the shared text — this branch had been two commits behind, so its copy of the `revert_component` note was stale and would have reverted #599's improvement.
- **On @gemini-code-assist's `<VersionBadge type="changed" />` suggestion**, @Ethan-Arrowood's reasoning to skip it held, but the conclusion has changed with the version: `by_ref` is unreleased surface, so the whole section now carries an *added* `v5.3.0` badge in #599 rather than a changed one here. Nothing to add mid-sentence.
- Approvals from @kriszyp and @cb1kenobi predate this pass; the net-new content is the two items above.
- `npm run build` and `npm run format:check` clean.


reference/components/applications.md
```diff
@@ -162,7 +162,7 @@ harper deploy ref=9f8c2a1 restart=true replicated=true
 
 **A reference is pinned to a SHA, not to the name you typed.** Tags and branches are resolved locally and the full commit SHA is what ships. This matters on a cluster: peers resolve the package independently, so a tag that moves mid-deploy — or a branch that advances — could otherwise leave nodes running different code.
 
-**Commit and push first.** The cluster clones from the remote, so it only sees commits that have been pushed. `by_ref` warns when the working tree is dirty, since uncommitted changes won't be part of the deploy.
+**Commit and push first.** The cluster clones from the remote, so it only sees commits that have been pushed. `by_ref` warns in both directions: when the working tree is dirty (those changes won't be part of the deploy) and when the commit being deployed isn't on any remote branch (the cluster won't be able to clone it). The second check reads your local remote-tracking refs, so run `git fetch` if you get it for a commit you know you pushed.
```
cb1kenobi, line 165: **Medium: the adjacent SHA-pinning guarantee (line 163) is not absolute, and this PR doesn't flag the gap**

The unpushed-commit warning added here is accurate to `warnIfCommitNotPushed` in [harper#1850](https://github.com/HarperFast/harper/pull/1850) (`bin/cliOperations.ts`) — it checks `git branch -r --contains <committish>`, i.e. local remote-tracking refs, and the `git fetch` advice matches the code's own comment.

But the sentence right above this one (line 163) states unconditionally: *"A reference is pinned to a SHA, not to the name you typed. Tags and branches are resolved locally and the full commit SHA is what ships."* That's only true when `resolveGitCommittish()` can resolve the ref locally. Its `catch` branch does the opposite:

```ts
try {
  return runGit(['rev-parse', `${refStr}^{commit}`]);
} catch {
  // Not resolvable locally (e.g. a ref that only exists on the remote). Pass it through
  // and let the cluster resolve it — losing the pin, ...
  return refStr;
}
```

So a `ref=<branch-or-tag-name>` that isn't fetched locally (e.g. a teammate's remote-only branch) is deployed **unpinned** — the raw mutable name, not a SHA — which is exactly the cluster-divergence hazard line 163 says `by_ref` prevents. This PR is a good place to add that caveat, since it's already revisiting this paragraph for the sibling unpushed-commit warning: e.g. "...full commit SHA is what ships **when the ref resolves locally; an unresolvable ref (one that exists only on the remote) is passed through unpinned by name instead.**"

—
_Generated by Barber AI_

kriszyp (approved):

Ethan-Arrowood (commented):
Holding rather than approving or blocking. This targets `claude/two-phase-deploy-docs` (#599), which is currently CONFLICTING with `main` and carries an unresolved CHANGES_REQUESTED, and it documents behavior from harper#1850, still open. Merging the stack as-is publishes docs for unshipped CLI behavior into the live `/reference/v5` tree. The content here is fine — the hold is purely sequencing.

**The line itself is accurate.** I checked the new wording against `warnIfCommitNotPushed` in harper#1850 (`bin/cliOperations.ts`): the gate is `git branch -r --contains <committish>`, so "reads your local remote-tracking refs" is exact, and the `git fetch` advice mirrors the CLI's own warning text. Nothing to change.

**One net-new gap.** The unpushed check is skipped under CI:

```ts
if (!process.env.GITHUB_SHA) warnIfCommitNotPushed(committish);
```

The dirty-tree warning is not skipped. "Warns in both directions" reads as unconditional, so a GitHub Actions user expecting the second warning silently won't get it. One clause covers it, or call it an implementation detail and leave it — your call.

**On @gemini-code-assist's `<VersionBadge type="changed" />` suggestion, I'd skip it**, on two grounds. Wrong type: `by_ref` is itself documented as v5.2.0 surface that hasn't shipped (harper's `package.json` on the 1850 branch is `5.2.0-beta.3`), so this warning lands in the *same* unreleased version as the feature — it isn't a behavior change to existing surface. Wrong placement: `CONTRIBUTING.md:232` reserves `<VersionBadge>` for standalone placement after headings precisely because mid-sentence use is awkward, and the bot's example puts it mid-paragraph.

Agreeing with @cb1kenobi on line 163 — I reached the same read of `resolveGitCommittish()` independently: the `catch` returns the raw `refStr`, so a remote-only ref deploys unpinned by name. Worth folding in while the paragraph is already open, even though the sentence itself belongs to #599.

The plain-text `Available since: v5.2.0` line on this section is also #599's, not this diff — I've raised it there rather than here.

cb1kenobi (approved):

dawsontoth:
Closing as absorbed by #636.

Every change here — `ref` resolving to a SHA with the local-then-remote fallback and fail-closed path, the `refs/heads`/`refs/tags` namespace restriction, `credential=true` with host derivation and mismatch rejection, the Actions-skipped unpushed-commit check, and `pull_request` head resolution — is in #636 against `main`, folded into the by_ref section rather than layered on top of it.

The reason for the move rather than the merge: everything this PR documents shipped in **v5.2.3** (`git tag --contains` on harper#1850's merge commit), while the branch it was stacked on is still gated on harper#1849, a draft. It was also correcting text in the PR beneath it, so a reviewer of that PR read `credential=<host>` — a form the CLI now rejects — and only saw the correction by opening this one. Folding the corrections in removes that window entirely.

The `APPROVED` here is from 2026-07-30 and predates dozens of commits on content that has since changed substantially, so it is not coverage worth preserving. #636 carries its own verification against harper v5.2.4 source.

Thanks @kriszyp and @cb1kenobi for the reviews — both findings from them survive in #636: the SHA-pinning guarantee is now stated with its actual local/remote/fail-closed behavior, and the adjacent gap cb1kenobi flagged at what was line 163 is closed at the source upstream and documented accordingly.

_🤖 Closed by Claude Code_

PR #618, opened 2026-08-03
Transaction log deletion is database-wide on RocksDB; table-scoped requests are rejected

### LINKED ISSUE #2049: RocksDB: table-scoped delete_transaction_logs_before destroys the entire database's transaction log
# RocksDB: table-scoped `delete_transaction_logs_before` destroys the entire database's transaction log

## Summary

On RocksDB, `delete_transaction_logs_before({ database, table, timestamp })` silently ignores the `table` scope and purges the **whole database's** transaction/audit log. An operator asking to prune one table's history permanently destroys every sibling table's history in the same database. The operation reports success with an `entries_deleted` count reflecting the wider deletion, and there is no warning and no way to recover.

The parameter is annotated `// lmdb only` in the signature, so this is known-not-supported on RocksDB — but the code path doesn't *reject* it, it accepts it and does something far more destructive than what was asked.

## Mechanism

`dataLayer/harperBridge/ResourceBridge.ts:481-519`:

```ts
async deleteTransactionLogsBefore(deleteObj: {
    schema?: string;
    database?: string;
    table?: string; // lmdb only
    timestamp: Date | number | string;
    cleanup_deleted_records?: boolean; // lmdb only
}): Promise<DeleteTransactionLogsBeforeResults> {
    ...
    const table = getTable(deleteObj);
    if (!table) {
        // no table, check if any of the tables are RocksDB
        // since all tables share the same transaction log store, we break after the first
        // RocksDB table is found
        const tables = getDatabases()[deleteObj.database];
        if (tables) {
            for (const table of Object.values(tables)) {
                if (table.primaryStore instanceof RocksDatabase) {
                    const deleted = table.primaryStore.purgeLogs({ before, includeEntryCounts: true });
                    ...
                    break;
                }
            }
        }
    } else if (table.primaryStore instanceof RocksDatabase) {
        const deleted = table.primaryStore.purgeLogs({ before, includeEntryCounts: true });
        ...
    }
    ...
}
```

The two RocksDB branches — "caller named a table" and "caller named no table" — do **the identical thing**. The comment in the second branch states why: *all tables share the same transaction log store*. So naming a table only selects which table's `primaryStore` handle is used to reach the shared per-database log store; the purge itself has no table dimension.

**Per-table purging isn't merely unimplemented — it isn't expressible at the storage layer.** `rocksdb-js` 2.5.0's `PurgeLogsOptions` is:

```ts
type PurgeLogsOptions = {
  before?: number;
  destroy?: boolean;
  includeEntryCounts?: boolean;
  name?: string;   // log name, e.g. "local" — not a table
};
```

There is no table/tableId filter, and the log is purged in whole-file units, which is fundamentally at odds with per-table granularity — a per-table purge would require rewriting log files rather than deleting them.

## Measured

Purging one 5,000-row table in a database with three 5,000-row tables:

- reported `entries_deleted: 15000` (i.e. all three tables' entries)
- after restart, the two sibling tables read **0** rows of history each (5,000 before)

Manager-verified by source read **and** an independent cold rerun (6/6 runs, identical numbers).

A detail that makes this worse in practice: a stale in-process cache keeps the sibling data *appearing* intact until restart, so an operator's immediate "did that do what I wanted?" check sees healthy tables. The loss surfaces on the next restart, decoupled in time from the action that caused it.

## Why this is worse than an unsupported-parameter bug

The failure mode is unrecoverable destruction of data the operator never asked to delete, on an operation whose whole purpose is bounded deletion. Ignoring an unsupported scope parameter is normally benign; here it widens the blast radius of a destructive operation by the number of tables in the database.

## Suggested direction

Short term, and I'd argue this is the whole fix: **reject `table` on RocksDB with a 400** rather than silently widening the scope. The signature already documents it as LMDB-only; make the validator enforce what the comment claims. `validation/transactionLogValidator.ts` is the natural place, though it needs the storage engine in scope to make the decision.

If per-table pruning on RocksDB is wanted as a feature, that's a `rocksdb-js` change first (a table filter in `PurgeLogsOptions`, plus a story for whole-file purge granularity) and should be a separate issue — but the silent-widening behaviour should not wait on it.

Worth deciding alongside: whether `cleanup_deleted_records` (also annotated `// lmdb only`) has the same silent-ignore shape. It looks inert rather than destructive on the RocksDB path, but it's the same pattern in the same signature.

## Provenance

Found by the qa-explorer exploratory QA loop. Source lines verified against `main` @ `b852a722c`; `PurgeLogsOptions` verified against the pinned `@harperfast/rocksdb-js` 2.5.0.

cb1kenobi:
Companion to [HarperFast/harper#2064](https://github.com/HarperFast/harper/pull/2064) (fixes [HarperFast/harper#2049](https://github.com/HarperFast/harper/issues/2049) — RocksDB: table-scoped delete_transaction_logs_before destroys the entire database's transaction log).

Updates `reference/database/transaction.md`:

- Adds a `delete_transaction_logs_before` section: on RocksDB (the default engine) deletion is database-wide because all tables in a database share one transaction log; as of v5.2.0 a request naming a `table` is rejected with a 400 (previously the table scope was silently ignored and the whole database's log was purged), and a nonexistent `table` returns a 404. On LMDB, `table` scopes deletion to that table, unchanged.
- Notes on `delete_audit_logs_before` that it is deprecated in favor of `delete_transaction_logs_before`, and that on RocksDB it always errors since it requires `table`.

`npm run build` and `npm run format:check` pass locally.

Generated by Claude (Fable 5).

### DIFF
--- a/reference/database/transaction.md
+++ b/reference/database/transaction.md
@@ -11,33 +11,25 @@ title: Transaction Logging
 
 # Transaction Logging
 
-Harper provides two complementary mechanisms for recording a history of data changes on a table: the **audit log** and the **transaction log**. Both record and expose history for individual tables, but their underlying storage differs: on RocksDB (the default storage engine) the transaction log is physically a single database-wide log shared by all tables — history is read per table, while deletion operates on the whole database's log.
+Harper maintains a **transaction log** for every database: a record of every data change, capturing the operation type, the user who made the change, the timestamp, and both the new and original record values. There is one transaction log per database, shared by all tables.
 
-| Feature                       | Audit Log                         | Transaction Log                                                         |
-| ----------------------------- | --------------------------------- | ----------------------------------------------------------------------- |
-| Storage                       | Standard Harper table (per-table) | Shared per-database log (RocksDB); clustering streams, per-table (LMDB) |
-| Requires clustering           | No                                | No (RocksDB, native WAL); Yes (LMDB, clustering streams)                |
-| Available since               | v4.1.0                            | v4.1.0                                                                  |
-| Stores original record values | Yes                               | No                                                                      |
-| Query by username             | Yes                               | No                                                                      |
-| Query by primary key          | Yes                               | No                                                                      |
-| Used for real-time messaging  | Yes (required)                    | No                                                                      |
-
-## Audit Log
+:::info Audit log and transaction log are the same thing
+Some operations and settings still carry the older "audit log" name — `read_audit_log`, `delete_audit_logs_before`, the [`logging.auditLog`](../logging/configuration.md) setting, and the `@table(audit:)` directive — but they all act on this single transaction log. The "audit log" is not a separate mechanism; the distinct-log terminology is a historical artifact.
+:::
 
 Available since: v4.1.0
 
-The audit log is a data store that tracks every transaction across all tables in a database. Harper automatically creates and maintains a single audit log per database. The audit log captures the operation type, the user who made the change, the timestamp, and both the new and original record values.
+On RocksDB (the default storage engine) the transaction log is physically a single database-wide log shared by all tables: history is read per table, while deletion operates on the whole database's log.
 
-The audit log is **enabled by default**. To disable it, set [`logging.auditLog`](../logging/configuration.md) to `false` in `harper-config.yaml` and restart Harper.
+The transaction log is **enabled by default**. To disable it, set [`logging.auditLog`](../logging/configuration.md) to `false` in `harper-config.yaml` and restart Harper.
 
-> The audit log is required for real-time messaging (WebSocket and MQTT subscriptions) and replication. Do not disable it if real-time features or replication are in use.
+> The transaction log is required for real-time messaging (WebSocket and MQTT subscriptions) and replication. Do not disable it if real-time features or replication are in use.
 
-### Audit Log Operations
+## Operations
 
-#### `read_audit_log`
+### `read_audit_log`
 
-Queries the audit log for a specific table. Supports filtering by timestamp, username, or primary key value.
+Queries the transaction log for a specific table. Supports filtering by timestamp, username, or primary key value.
 
 **By timestamp:**
 
@@ -114,9 +106,9 @@ Timestamp behavior:
 
 The `original_records` field contains the record state before the operation was applied.
 
-#### `delete_audit_logs_before`
+### `delete_audit_logs_before`
 
-Deletes audit log entries older than the specified timestamp. Deprecated in favor of [`delete_transaction_logs_before`](#delete_transaction_logs_before).
+Deletes transaction log entries older than the specified timestamp. Deprecated in favor of [`delete_transaction_logs_before`](#delete_transaction_logs_before).
 
 <VersionBadge type="changed" version="v4.3.0" /> — Audit log cleanup improved to reduce resource consumption during scheduled cleanups
 
@@ -133,9 +125,7 @@ Deletes audit log entries older than the specified timestamp. Deprecated in favo
 }
 ```
 
-### Transaction Log Operations
-
-#### `delete_transaction_logs_before`
+### `delete_transaction_logs_before`
 
 <EngineBadge engines="RocksDB, LMDB" />
 
@@ -156,7 +146,7 @@ Parameters:
 
 On LMDB, the table-scoped deletion scans the database's full audit history (and `cleanup_deleted_records: true` adds a second full scan of the table's records), so the cost grows with total history depth — schedule accordingly on databases with deep audit history.
 
-The operation runs as a background job: the request itself returns `200` with a job ID, and any rejection surfaces when the job runs. Poll [`get_job`](jobs.md#get-job) to observe the outcome — a rejected request ends with job status `ERROR` and a message describing the failure (for example, the RocksDB table-scope rejection).
+Request validation runs first and synchronously: a request that omits both `database` and `schema` is rejected immediately with an error and no job ID (there is nothing to poll). Once accepted, the operation runs as a background job that returns `200` with a job ID, and operation-time failures surface through [`get_job`](jobs.md#get-job) — the job ends with status `ERROR` and a message describing the failure (for example, the RocksDB table-scope rejection, or a `table`/`database` that does not exist).
 
 **RocksDB (database-wide — omit `table`):**
 
@@ -203,9 +193,9 @@ Response:
 
 ---
 
-## Enabling Audit Log Per Table
+## Enabling the Transaction Log Per Table
 
-You can enable or disable the audit log for individual tables using the `@table` directive's `audit` argument in your schema:
+You can enable or disable the transaction log for individual tables using the `@table` directive's `audit` argument in your schema:
 
 ```graphql
 type Dog @table(audit: true) {
@@ -218,7 +208,7 @@ This overrides the [`logging.auditLog`](../logging/configuration.md) global conf
 
 ## Related Documentation
 
-- [Logging](../logging/overview.md) — Application and system logging (separate from transaction/audit logging)
-- [Replication](../replication/overview.md) — Clustering setup required for transaction logs
-- [Logging Configuration](../logging/configuration.md) — Global audit log configuration (`logging.auditLog`)
+- [Logging](../logging/overview.md) — Application and system logging (separate from the transaction log)
+- [Replication](../replication/overview.md) — Replication and clustering, which consume the transaction log
+- [Logging Configuration](../logging/configuration.md) — Global transaction log configuration (`logging.auditLog`)
 - [Operations API](../operations-api/overview.md) — Sending operations to Harper
--- a/reference/database/transaction.md
+++ b/reference/database/transaction.md
@@ -16,7 +16,7 @@ Harper provides two complementary mechanisms for recording a history of data cha
 | Feature                       | Audit Log                         | Transaction Log                                                         |
 | ----------------------------- | --------------------------------- | ----------------------------------------------------------------------- |
 | Storage                       | Standard Harper table (per-table) | Shared per-database log (RocksDB); clustering streams, per-table (LMDB) |
-| Requires clustering           | No                                | Yes                                                                     |
+| Requires clustering           | No                                | No (RocksDB, native WAL); Yes (LMDB, clustering streams)                |
 | Available since               | v4.1.0                            | v4.1.0                                                                  |
 | Stores original record values | Yes                               | No                                                                      |
 | Query by username             | Yes                               | No                                                                      |
@@ -139,20 +139,42 @@ Deletes audit log entries older than the specified timestamp. Deprecated in favo
 
 <EngineBadge engines="RocksDB, LMDB" />
 
-<VersionBadge type="changed" version="v5.2.0" /> — On RocksDB, a request that includes `table` now fails; previously the `table` scope was silently ignored and the entire database's transaction log was purged. On either engine, a `table` that does not exist now fails with a not-found error (previously this was a silent no-op on LMDB). On LMDB, a valid `table` continues to scope the deletion to that table's history, unchanged.
+<VersionBadge type="changed" version="v5.2.0" /> — On RocksDB, a request that includes `table` now fails; previously the `table` scope was silently ignored and the entire database's transaction log was purged. On either engine, a `table` that does not exist now fails with a not-found error (previously a typo'd `table` fell through to the database-wide purge on RocksDB, and was a silent no-op on LMDB). On LMDB, a valid `table` continues to scope the deletion to that table's history, unchanged.
 
 Deletes transaction log entries older than the specified timestamp.
 
-On RocksDB (the default storage engine), deletion is database-wide: all tables in a database share one transaction log, so omit `table` and pass only `database` (or `schema`) and `timestamp`.
+:::warning Database-wide and irreversible
+On RocksDB (the default storage engine), deletion is database-wide: all tables in a database share one transaction log, so omit `table` and pass only `database` (or `schema`) and `timestamp`. This deletes whole log files, permanently removing [`read_audit_log`](#read_audit_log) history below the timestamp for **every table in the database** — there is no per-table survivor and no undo. The only recovery route is a [backup](../backups/overview.md), which restores the transaction log alongside the data. Purging below a lagging replica's catch-up position does not lose data on that replica — the sender detects that the requested start predates its retained history and forces a full base copy instead of incremental catch-up — but that full resync is far more expensive than incremental replication, so avoid purging below your slowest replica's position.
+:::
 
-`cleanup_deleted_records`: `boolean` (optional) — LMDB only; additionally removes leftover tombstone entries for records deleted before the timestamp, a repair step for tombstones that normal audit log cleanup should already have removed. Ignored on RocksDB.
+Parameters:
 
-The operation runs as a background job: the request itself returns `200` with a job ID, and any rejection surfaces when the job runs. Poll [`get_job`](jobs.md) to observe the outcome — a rejected request ends with job status `ERROR` and a message describing the failure (for example, the RocksDB table-scope rejection).
+- `database` (or the deprecated `schema` alias): `string` (required) — a request naming neither fails validation before a job starts; a `database` that does not exist fails the job with a not-found error.
+- `timestamp`: `number` (required) — epoch milliseconds; entries older than this are deleted.
+- `table`: `string` (LMDB only) — scopes deletion to that table's history. On RocksDB the job fails (see the warning above).
+- `cleanup_deleted_records`: `boolean` (optional) — LMDB only; additionally removes leftover tombstone entries for records deleted before the timestamp, a repair step for tombstones that normal audit log cleanup should already have removed. Ignored on RocksDB.
+
+On LMDB, the table-scoped deletion scans the database's full audit history (and `cleanup_deleted_records: true` adds a second full scan of the table's records), so the cost grows with total history depth — schedule accordingly on databases with deep audit history.
+
+The operation runs as a background job: the request itself returns `200` with a job ID, and any rejection surfaces when the job runs. Poll [`get_job`](jobs.md#get-job) to observe the outcome — a rejected request ends with job status `ERROR` and a message describing the failure (for example, the RocksDB table-scope rejection).
+
+**RocksDB (database-wide — omit `table`):**
 
 ```json
 {
 	"operation": "delete_transaction_logs_before",
 	"database": "dev",
+	"timestamp": 1598290282817
+}
+```
+
+**LMDB (`table` is required — omitting it deletes nothing and reports `entries_deleted: 0` with job status `COMPLETE`):**
+
+```json
+{
+	"operation": "delete_transaction_logs_before",
+	"database": "dev",
+	"table": "dog",
 	"timestamp": 1598290282817
 }
 ```
@@ -164,6 +186,19 @@ Response:
 	"message": "Starting job with id 2fe25039-566e-4670-8bb3-2db3d4e07e69",
 	"job_id": "2fe25039-566e-4670-8bb3-2db3d4e07e69"
 }
+```
+
+`get_job` reports the outcome. A successful job's `result` carries `log_files_deleted` and `entries_deleted` (plus `start_timestamp`/`end_timestamp` on LMDB) — the record of how much was deleted. A rejected request looks like:
+
+```json
+[
+	{
+		"id": "2fe25039-566e-4670-8bb3-2db3d4e07e69",
+		"type": "delete_transaction_logs_before",
+		"status": "ERROR",
+		"message": "There was an error running deleteTransactionLogsBefore job with id 2fe25039-566e-4670-8bb3-2db3d4e07e69 - Table-level transaction log deletion is not supported for RocksDB tables because all tables in a database share one transaction log; to delete the transaction logs for the entire 'dev' database, use delete_transaction_logs_before with only 'database' and 'timestamp'"
+	}
+]
 ```
 
 ---
--- a/reference/database/transaction.md
+++ b/reference/database/transaction.md
@@ -116,17 +116,37 @@ The `original_records` field contains the record state before the operation was 
 
 #### `delete_audit_logs_before`
 
-Deletes audit log entries older than the specified timestamp.
+Deletes audit log entries older than the specified timestamp. Deprecated in favor of [`delete_transaction_logs_before`](#delete_transaction_logs_before).
 
 <VersionBadge type="changed" version="v4.3.0" /> — Audit log cleanup improved to reduce resource consumption during scheduled cleanups
 
 <VersionBadge type="changed" version="v4.5.0" /> — Storage reclamation: Harper automatically evicts older audit log entries when free storage drops below a configurable threshold
+
+<VersionBadge type="changed" version="v5.2.0" /> — This operation requires `table`, but on the RocksDB storage engine (the default) history cannot be deleted for a single table because all tables in a database share one transaction log. On RocksDB this operation now always returns an error directing you to `delete_transaction_logs_before`; it remains usable on LMDB.
 
 ```json
 {
 	"operation": "delete_audit_logs_before",
 	"schema": "dev",
 	"table": "dog",
+	"timestamp": 1598290282817
+}
+```
+
+#### `delete_transaction_logs_before`
+
+Deletes transaction log entries older than the specified timestamp.
+
+<EngineBadge engines="RocksDB, LMDB" />
+
+On RocksDB (the default storage engine), deletion is database-wide: all tables in a database share one transaction log, so omit `table` and pass only `database` (or `schema`) and `timestamp`.
+
+<VersionBadge type="changed" version="v5.2.0" /> — On RocksDB, a request that includes `table` is now rejected with a `400` error, and a `table` that does not exist returns a `404`. Previously the `table` scope was silently ignored and the entire database's transaction log was purged. On LMDB, `table` scopes the deletion to that table's history, unchanged.
+
+```json
+{
+	"operation": "delete_transaction_logs_before",
+	"database": "dev",
 	"timestamp": 1598290282817
 }
 ```
--- a/reference/database/transaction.md
+++ b/reference/database/transaction.md
@@ -143,7 +143,9 @@ Deletes transaction log entries older than the specified timestamp.
 
 On RocksDB (the default storage engine), deletion is database-wide: all tables in a database share one transaction log, so omit `table` and pass only `database` (or `schema`) and `timestamp`.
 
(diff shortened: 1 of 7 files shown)


reference/database/transaction.md
```diff
@@ -131,6 +133,24 @@ Deletes audit log entries older than the specified timestamp.
 }
 ```
```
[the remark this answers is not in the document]
cb1kenobi: Good call — added a `### Transaction Log Operations` header in 22004028 so the operation no longer sits under Audit Log Operations. — Claude (Fable 5)

reference/database/transaction.md
```diff
@@ -131,6 +133,26 @@ Deletes audit log entries older than the specified timestamp.
...
+
+#### `delete_transaction_logs_before`
+
+Deletes transaction log entries older than the specified timestamp.
+
+<EngineBadge engines="RocksDB, LMDB" />
+
+On RocksDB (the default storage engine), deletion is database-wide: all tables in a database share one transaction log, so omit `table` and pass only `database` (or `schema`) and `timestamp`.
+
+<VersionBadge type="changed" version="v5.2.0" /> — On RocksDB, a request that includes `table` is now rejected with a `400` error, and a `table` that does not exist returns a `404`. Previously the `table` scope was silently ignored and the entire database's transaction log was purged. On LMDB, `table` scopes the deletion to that table's history, unchanged.
```
cb1kenobi: **Medium: LMDB's "unchanged" claim is inaccurate — the nonexistent-table check is engine-agnostic**

The new `ResourceBridge.deleteTransactionLogsBefore` (harper#2064) adds `if (deleteObj.table != null && !table) throw ...404...` *before* branching on storage engine. This check fires for LMDB too, not just RocksDB. Previously, a nonexistent `table` on LMDB fell into the same `if (!table)` branch as "no table given" — the RocksDB-detection loop found no RocksDB tables and the call returned success with `entries_deleted: 0`, silently. After this change, a typo'd `table` name on LMDB now throws a `404` instead of that silent no-op.

The doc frames "a `table` that does not exist returns a `404`" as a RocksDB-specific behavior, then says "On LMDB, `table` scopes the deletion to that table's history, unchanged." That's only true for a *valid* table name — for a nonexistent one, LMDB behavior did change (silent success → 404). An operator/script that relied on the old silent no-op for a since-dropped or misspelled table name will now see an error where it previously got a clean (if useless) success.

Suggested fix: scope the "unchanged" claim to valid table names, e.g. "On LMDB, a valid `table` continues to scope the deletion to that table's history; a nonexistent `table` now also returns a `404` on either engine (previously a silent no-op on LMDB)."

—
_Generated by Barber AI_

reference/database/transaction.md
```diff
@@ -131,6 +133,26 @@ Deletes audit log entries older than the specified timestamp.
...
+
+```json
+{
+	"operation": "delete_transaction_logs_before",
```
cb1kenobi, line 155: **Low: `cleanup_deleted_records` (LMDB-only) is undocumented in this newly-added section**

`ResourceBridge.deleteTransactionLogsBefore` accepts `cleanup_deleted_records?: boolean` (LMDB-only — used in the `table.deleteHistory(before, deleteObj.cleanup_deleted_records)` branch), but this is the first real documentation `delete_transaction_logs_before` has ever had (previously just a table-of-contents pointer), and it doesn't mention the parameter at all. Given the section otherwise carefully calls out which params are engine-conditional (`table`), this is worth adding alongside it for completeness.

Suggested fix: add a short line noting `cleanup_deleted_records` (boolean, LMDB-only) alongside the `table` engine-conditional note.

—
_Generated by Barber AI_

reference/database/transaction.md
cb1kenobi: Accurate — the 404 guard in harper#2064 sits before the engine branch, so nonexistent-table behavior changed on LMDB too (silent no-op → 404). Reworded in f3d66404: the 404 is now described as applying on either engine, with the LMDB "unchanged" claim scoped to valid table names. — Claude (Fable 5)

reference/database/transaction.md
cb1kenobi: Added in f3d66404 — documented `cleanup_deleted_records` as LMDB-only (removes leftover delete tombstones older than the timestamp; ignored on RocksDB), verified against Table.deleteHistory. — Claude (Fable 5)

reference/database/transaction.md
```diff
@@ -131,6 +133,28 @@ Deletes audit log entries older than the specified timestamp.
...
+
+#### `delete_transaction_logs_before`
+
+Deletes transaction log entries older than the specified timestamp.
+
+<EngineBadge engines="RocksDB, LMDB" />
+
+On RocksDB (the default storage engine), deletion is database-wide: all tables in a database share one transaction log, so omit `table` and pass only `database` (or `schema`) and `timestamp`.
+
+<VersionBadge type="changed" version="v5.2.0" /> — On RocksDB, a request that includes `table` is now rejected with a `400` error; previously the `table` scope was silently ignored and the entire database's transaction log was purged. On either engine, a `table` that does not exist now returns a `404` (previously this was a silent no-op on LMDB). On LMDB, a valid `table` continues to scope the deletion to that table's history, unchanged.
```
kriszyp: `delete_transaction_logs_before` is dispatched as a background job, so the caller does not receive HTTP 400/404 from this request. The integration path in [core PR #2064, “`delete_transaction_logs_before` with a table on RocksDB now returns an error instead of deleting the entire database's transaction log”](https://github.com/HarperFast/harper/pull/2064), explicitly accepts HTTP 200, extracts `job_id`, and later observes job status `ERROR`; 400/404 are the internal error status codes. As written, clients may branch on the POST status and never poll the job. Please state that the request returns a job ID and the job ends in `ERROR`, and only describe the numeric codes if `get_job` exposes where clients can read them. A short initial-response/`get_job` example would make the contract unambiguous.

— KrAIs (Codex)

reference/database/transaction.md
```diff
@@ -131,6 +133,28 @@ Deletes audit log entries older than the specified timestamp.
...
 
+### Transaction Log Operations
+
+#### `delete_transaction_logs_before`
+
+Deletes transaction log entries older than the specified timestamp.
+
+<EngineBadge engines="RocksDB, LMDB" />
+
+On RocksDB (the default storage engine), deletion is database-wide: all tables in a database share one transaction log, so omit `table` and pass only `database` (or `schema`) and `timestamp`.
```
kriszyp: This correctly says RocksDB uses one database-wide transaction log, but the same page's introduction says both mechanisms are available at table level and its comparison table labels transaction-log storage as `Clustering streams (per-table)`. Readers now get incompatible storage models on one page. Could you update that overview to distinguish logical per-table history/read filtering from RocksDB's physically shared database-wide log and purge scope?

— KrAIs (Codex)

reference/database/transaction.md
```diff
@@ -116,12 +116,14 @@ The `original_records` field contains the record state before the operation was
...
 #### `delete_audit_logs_before`
 
-Deletes audit log entries older than the specified timestamp.
+Deletes audit log entries older than the specified timestamp. Deprecated in favor of [`delete_transaction_logs_before`](#delete_transaction_logs_before).
 
 <VersionBadge type="changed" version="v4.3.0" /> — Audit log cleanup improved to reduce resource consumption during scheduled cleanups
 
 <VersionBadge type="changed" version="v4.5.0" /> — Storage reclamation: Harper automatically evicts older audit log entries when free storage drops below a configurable threshold
 
+<VersionBadge type="changed" version="v5.2.0" /> — This operation requires `table`, but on the RocksDB storage engine (the default) history cannot be deleted for a single table because all tables in a database share one transaction log. On RocksDB this operation now always returns an error directing you to `delete_transaction_logs_before`; it remains usable on LMDB.
```
kriszyp: “Always returns an error directing you” is too broad. The companion core change resolves the table first, so a nonexistent table throws `TABLE_NOT_FOUND`/404 before reaching the RocksDB migration error; the new paragraph below also documents that case. Please qualify this as an existing RocksDB table, or say more generally that the operation is unsupported on RocksDB and direct valid-table callers to `delete_transaction_logs_before`.

— KrAIs (Codex)

reference/database/transaction.md
kriszyp: The contribution guide requires a version badge to appear directly below the heading it annotates. Could you move this v5.2 change badge immediately below `delete_transaction_logs_before` (alongside or adjacent to the engine badge) and place the descriptive prose after it?

— KrAIs (Codex)

reference/database/transaction.md
```diff
@@ -131,6 +133,28 @@ Deletes audit log entries older than the specified timestamp.
...
+
+Deletes transaction log entries older than the specified timestamp.
+
+<EngineBadge engines="RocksDB, LMDB" />
+
+On RocksDB (the default storage engine), deletion is database-wide: all tables in a database share one transaction log, so omit `table` and pass only `database` (or `schema`) and `timestamp`.
+
+<VersionBadge type="changed" version="v5.2.0" /> — On RocksDB, a request that includes `table` is now rejected with a `400` error; previously the `table` scope was silently ignored and the entire database's transaction log was purged. On either engine, a `table` that does not exist now returns a `404` (previously this was a silent no-op on LMDB). On LMDB, a valid `table` continues to scope the deletion to that table's history, unchanged.
+
+On LMDB only, the optional `cleanup_deleted_records` (boolean) parameter additionally removes leftover tombstone entries for records deleted before the timestamp — a repair step for tombstones that normal audit log cleanup should already have removed. It is ignored on RocksDB.
```
kriszyp: The repository's content style requires type annotations in the form `` `<type>` (<modifier>) ``. Could you describe this as `` `cleanup_deleted_records`: `boolean` (optional) `` (or equivalent) rather than putting `(boolean)` after the parameter name?

— KrAIs (Codex)

kriszyp (commented):
Probably should clarify the response, since it is returns a job-id.

reference/database/transaction.md
cb1kenobi: Reworked in ffe30c44 — the section now states the operation runs as a background job (request returns 200 with a job ID), that rejections surface as job status ERROR with a descriptive message via get_job, and includes the initial-response example. Dropped the bare 400/404 numerals since get_job exposes the message, not the HTTP code. — Claude (Fable 5)

reference/database/transaction.md
cb1kenobi: Updated in ffe30c44 — the intro now distinguishes logical per-table history from RocksDB's physically shared database-wide log, and the comparison table's storage cell reads 'Shared per-database log (RocksDB); clustering streams, per-table (LMDB)'. — Claude (Fable 5)

reference/database/transaction.md
cb1kenobi: Qualified in ffe30c44 — now 'unsupported on RocksDB', with the existing-table case (migration error) and nonexistent-table case (not-found error) called out separately. — Claude (Fable 5)

reference/database/transaction.md
cb1kenobi: Moved in ffe30c44 — the v5.2 change badge now sits directly below the delete_transaction_logs_before heading next to the engine badge, with the prose after. — Claude (Fable 5)

reference/database/transaction.md
cb1kenobi: Restyled in ffe30c44 to the content-style form: `cleanup_deleted_records`: `boolean` (optional). — Claude (Fable 5)

reference/database/transaction.md
```diff
@@ -11,17 +11,17 @@ title: Transaction Logging
...
-| Used for real-time messaging  | Yes (required)                    | No                             |
+Harper provides two complementary mechanisms for recording a history of data changes on a table: the **audit log** and the **transaction log**. Both record and expose history for individual tables, but their underlying storage differs: on RocksDB (the default storage engine) the transaction log is physically a single database-wide log shared by all tables — history is read per table, while deletion operates on the whole database's log.
+
+| Feature                       | Audit Log                         | Transaction Log                                                         |
+| ----------------------------- | --------------------------------- | ----------------------------------------------------------------------- |
+| Storage                       | Standard Harper table (per-table) | Shared per-database log (RocksDB); clustering streams, per-table (LMDB) |
+| Requires clustering           | No                                | Yes                                                                     |
```
Ethan-Arrowood: Gap 2 from the body. This row is v4/NATS-era and is false on v5/RocksDB, where the transaction log is the native WAL and exists whether or not replication is configured. It now contradicts the Storage row directly above it, which this PR rewrote — so I think it's fair game here rather than a separate PR.

```suggestion
| Requires clustering           | No                                | No (RocksDB, native WAL); Yes (LMDB, clustering streams)                |
```

Ethan-Arrowood (commented):
I went through this against the merged implementation in harper#2064 rather than taking the prose on trust — the 400 guard for a RocksDB `table` scope, the engine-agnostic 404 that sits *before* the engine branch, `cleanup_deleted_records` against `Table.deleteHistory`, the background-job dispatch, and the response body all check out line by line. Every factual claim in the diff is accurate, and all eight points from the earlier rounds (Gemini's section header, @kriszyp's five Codex findings, Barber AI's two) are correctly resolved in `ffe30c44`. I'm not re-raising any of those.

What's left is what the page *doesn't* say, which on a destructive operation is the part that matters. Four gaps:

1. **On LMDB, omitting `table` is a silent no-op** — and the section's only example omits it. That request resolves `getTable()` to `undefined`, falls into the `if (!table)` branch, finds no `RocksDatabase` instance, and returns `entries_deleted: 0` with job status `COMPLETE`. An LMDB operator copying the example verbatim schedules a retention job that quietly does nothing and reports success. Harper's own validator acknowledges this (`validation/transactionLogValidator.ts:20`: `// \`table\` will need to be required for lmdb, but not for rocksdb`). Inline suggestion below.

2. **`Requires clustering: Yes` in the comparison table is false on v5/RocksDB.** The 5.0 release notes describe the transaction log as Harper's "own native transaction log as a write-ahead log (WAL) for RocksDB" — unconditional, not a clustering feature; `readTransactionLog` just delegates to `harperBridge.readAuditLog` with no stream involvement; and `reference/backups/overview.md` says every managed backup includes the transaction log with no clustering caveat. As it stands the row directly contradicts the Storage row above it, which this PR rewrote. A RocksDB reader concludes they have no transaction log unless they've configured replication, then can't reconcile that with a page telling them how to delete it. Inline suggestion below. (The `Available since v4.1.0` row has the same v4-era problem, but I'd leave that one for a separate pass.)

3. **A database-wide destructive purge is documented in plain body prose.** Line 146 is the sentence that tells an operator who wanted to prune one table to instead purge every table's history, and it sits in an unmarked paragraph. The reference tree uses admonitions for exactly this (`:::warning Data safety` in `reference/database/api.md:54`). Three things are missing beyond the wrapper: what is actually lost (`purgeLogs()` deletes whole log files — no per-table survivor, no undo), and the recovery route (a managed backup restores the transaction log alongside the data, per `reference/backups/operations.md:113`). Inline suggestion below covers both.

   The third piece I couldn't establish myself and it's a genuine question for you or @kriszyp: **what happens if you purge below a lagging peer's replication position?** `purgeLogs` looks like a whole-file delete with no watermark check at the bridge layer. If there's no guard, "do not purge below your slowest replica's position" belongs in that admonition. If rocksdb-js refuses to delete unshipped files, that's reassuring and also belongs there. Either answer improves the page — silence is the only bad option, so I left it out of the suggestion rather than guess.

4. **No entry in `release-notes/v5-lincoln/5.2.md`.** This is breaking in three ways: table-scoped `delete_transaction_logs_before` now errors where it previously returned success; `delete_audit_logs_before` now always errors on the default engine; and a nonexistent `table` now 404s on LMDB where it previously returned a clean success. Anyone with a scheduled retention job hits one of the first two on upgrade. `5.2.md` is actively maintained with prose sections and a reference-page badge isn't a substitute. A short `### Transaction Log Deletion` section would do it. Fine as a follow-up if you'd rather not grow this PR — but it shouldn't ship without one.

Everything else below is a suggestion, take or leave. Items 1-3 are cheap enough that I'd land them here.

reference/database/transaction.md
```diff
@@ -131,6 +133,39 @@ Deletes audit log entries older than the specified timestamp.
 }
 ```
 
+### Transaction Log Operations
+
+#### `delete_transaction_logs_before`
+
+<EngineBadge engines="RocksDB, LMDB" />
+
+<VersionBadge type="changed" version="v5.2.0" /> — On RocksDB, a request that includes `table` now fails; previously the `table` scope was silently ignored and the entire database's transaction log was purged. On either engine, a `table` that does not exist now fails with a not-found error (previously this was a silent no-op on LMDB). On LMDB, a valid `table` continues to scope the deletion to that table's history, unchanged.
```
Ethan-Arrowood: Accurate, but this tells the LMDB half of the story and the omitted half is the dangerous one: on RocksDB a *typo'd* table name previously fell through to the no-table branch and purged the entire database's log. harper#2064's own design note calls this out specifically. On a page about data-loss semantics, the destructive prior behavior is the one readers need in order to recognize it in their incident history — worth adding a clause to the parenthetical.

reference/database/transaction.md
```diff
@@ -131,6 +133,39 @@ Deletes audit log entries older than the specified timestamp.
...
+
+Deletes transaction log entries older than the specified timestamp.
+
+On RocksDB (the default storage engine), deletion is database-wide: all tables in a database share one transaction log, so omit `table` and pass only `database` (or `schema`) and `timestamp`.
```
Ethan-Arrowood: Gap 3 from the body — this is the sentence that redirects a single-table prune into a whole-database purge, and it's unmarked body prose. Suggestion below adds the admonition, states what is unrecoverable, and cross-links backups as the recovery route. I deliberately left the replication-lag sentence out because I couldn't establish the behavior; if you can confirm it either way, add it here.

```suggestion
:::warning Database-wide and irreversible
On RocksDB (the default storage engine), deletion is database-wide: all tables in a database share one transaction log, so omit `table` and pass only `database` (or `schema`) and `timestamp`. This deletes whole log files, permanently removing [`read_audit_log`](#read_audit_log) history below the timestamp for **every table in the database** — there is no per-table survivor and no undo. The only recovery route is a [backup](../backups/overview.md), which restores the transaction log alongside the data.
:::
```

reference/database/transaction.md
```diff
@@ -131,6 +133,39 @@ Deletes audit log entries older than the specified timestamp.
...
+
+`cleanup_deleted_records`: `boolean` (optional) — LMDB only; additionally removes leftover tombstone entries for records deleted before the timestamp, a repair step for tombstones that normal audit log cleanup should already have removed. Ignored on RocksDB.
```
Ethan-Arrowood: Three small parameter gaps worth closing while you're in here:

1. `database` (or `schema`) is **required** — the Joi schema is `.or('schema', 'database')`, so a request missing both is a `400` at validation, before the job starts. The prose implies it but doesn't state it.
2. The **success** result shape is undocumented: `log_files_deleted` and `entries_deleted` (plus `start_timestamp`/`end_timestamp`). For a destructive operation, "how much did I just delete" is the field an operator most wants to see.
3. A nonexistent `database` throws `SCHEMA_NOT_FOUND`/404 from `getTable()` — same class as the documented table 404, currently unmentioned.

Also worth a sentence on cost: on LMDB the table-scoped path walks `auditStore.getRange({start: 0, end: endTime})` for the whole database and filters by `tableId` in the loop, and `cleanup_deleted_records: true` adds a second full `primaryStore` scan. That's O(database history) per call. `reference/operations-api/operations.md` already warns "Costly on tables with deep audit history" elsewhere, and this page is where an operator decides how often to schedule it.

reference/database/transaction.md
```diff
@@ -131,6 +133,39 @@ Deletes audit log entries older than the specified timestamp.
...
+
+The operation runs as a background job: the request itself returns `200` with a job ID, and any rejection surfaces when the job runs. Poll [`get_job`](jobs.md) to observe the outcome — a rejected request ends with job status `ERROR` and a message describing the failure (for example, the RocksDB table-scope rejection).
```
Ethan-Arrowood: Two things here.

The link resolves but lands at the top of the page; `get_job` is under `### Get Job`, so deep-link it. Note `onBrokenAnchors` isn't configured in `docusaurus.config.ts` (only `onBrokenLinks: 'throw'`), so anchors on this page aren't build-enforced — I checked the new `#delete_transaction_logs_before` anchor by hand and the slug matches, it's just unguarded.

```suggestion
The operation runs as a background job: the request itself returns `200` with a job ID, and any rejection surfaces when the job runs. Poll [`get_job`](jobs.md#get-job) to observe the outcome — a rejected request ends with job status `ERROR` and a message describing the failure (for example, the RocksDB table-scope rejection).
```

Separately: @kriszyp asked for "a short initial-response/`get_job` example" and only the initial response landed. The whole point of this paragraph is that the rejection is invisible in the POST response, so the ERROR shape is the more valuable of the two. The integration tests give you the exact form — `get_job` returns an array, with `body[0].status === 'ERROR'` and `body[0].message` carrying the migration text.

reference/database/transaction.md
```diff
@@ -131,6 +133,39 @@ Deletes audit log entries older than the specified timestamp.
...
+
+```json
+{
+	"operation": "delete_transaction_logs_before",
+	"database": "dev",
+	"timestamp": 1598290282817
+}
+```
```
Ethan-Arrowood, line 159: Gap 1 from the body. The section is badged `RocksDB, LMDB` but this is the RocksDB form only — on LMDB it's a silent no-op that reports success. Labelling both forms fixes it without needing new prose:

````suggestion
**RocksDB (database-wide — omit `table`):**

```json
{
	"operation": "delete_transaction_logs_before",
	"database": "dev",
	"timestamp": 1598290282817
}
```

**LMDB (`table` is required — omitting it deletes nothing and reports `entries_deleted: 0` with job status `COMPLETE`):**

```json
{
	"operation": "delete_transaction_logs_before",
	"database": "dev",
	"table": "dog",
	"timestamp": 1598290282817
}
```
````

File: reference/database/transaction.md (hunk 8 above)
Comment (cb1kenobi) [reply reference/database/transaction.md]: Applied your suggestion in 5cc6c517 — the row now reads 'No (RocksDB, native WAL); Yes (LMDB, clustering streams)', consistent with the rewritten Storage row. — Claude (Fable 5)

File: reference/database/transaction.md (hunk 9 above)
Comment (cb1kenobi) [reply reference/database/transaction.md]: Added in 5cc6c517 — the parenthetical now leads with the destructive prior behavior: '(previously a typo'd `table` fell through to the database-wide purge on RocksDB, and was a silent no-op on LMDB)'. — Claude (Fable 5)

File: reference/database/transaction.md (hunk 10 above)
Comment (cb1kenobi) [reply reference/database/transaction.md]: Applied in 5cc6c517 with your admonition text, plus the replication-lag answer — I was able to establish it: there is no purge-side guard, but the sender detects a peer whose requested incremental start predates retained history (shouldForceBaseCopyForRetention, harper-pro replication/replicationConnection.ts, harper-pro#277) and forces a bounded base copy rather than silently skipping purged entries or replaying unbounded history (harper#1114). So purging below a lagging replica's position costs a full resync, not data loss — the admonition now says exactly that and advises against purging below the slowest replica's position. — Claude (Fable 5)

File: reference/database/transaction.md (hunk 11 above)
Comment (cb1kenobi) [reply reference/database/transaction.md]: All four closed in 5cc6c517: a Parameters list marks `database`/`schema` as required (with the no-database validation failure and nonexistent-database not-found cases), the success result fields (log_files_deleted, entries_deleted, start_timestamp/end_timestamp) are documented under the get_job paragraph, and an O(history)-cost sentence covers the LMDB scan plus the cleanup_deleted_records second scan. — Claude (Fable 5)

File: reference/database/transaction.md (hunk 12 above)
Comment (cb1kenobi) [reply reference/database/transaction.md]: Both applied in 5cc6c517 — deep link is now jobs.md#get-job, and a get_job ERROR response example shows the array shape with status ERROR and the prefixed message string (verified against jobRunner's error path: message is a string, 'There was an error running <op> job with id <id> - <error>'). — Claude (Fable 5)

File: reference/database/transaction.md (hunk 13 above)
Comment (cb1kenobi) [reply reference/database/transaction.md:159]: Applied your two-example suggestion in 5cc6c517, with the LMDB caption noting the silent entries_deleted: 0 / COMPLETE outcome when table is omitted. — Claude (Fable 5)

**Review by kriszyp — CHANGES_REQUESTED** [review]
Since we are editing the Audit Log / Transaction Log table... Let's just eliminate it and make it clear there is no such thing as a difference between an "audit log" and a "transaction log". That's an ancient relic of NATS that we don't talk about anymore. We just have a transaction log and that's it.

- `reference/database/transaction.md:159`: The new parameter documentation says that omitting both `database` and `schema` fails validation before a job starts, so “any rejection surfaces when the job runs” is now too broad. A client making that mistake will not receive the job ID described here and cannot poll `get_job`. Could you distinguish request-validation failures from errors raised after a job has been accepted, for example: “Once accepted, the operation runs as a background job; operation-time failures surface through `get_job`”?

— KrAIs (Codex)
- `release-notes/v5-lincoln/5.2.md:22`: “Always fails with the same guidance” conflicts with the next bullet and the operation reference: a nonexistent table fails with a not-found error before the RocksDB guidance is reached. Could you qualify this as an existing RocksDB table, matching `reference/database/transaction.md`, so upgrade guidance does not promise the wrong error for typo'd table names?

— KrAIs (Codex)
- `reference/database/transaction.md:222`: The comparison table now correctly says RocksDB transaction logging uses the native WAL and does not require clustering, but this related-doc description still says clustering setup is required for transaction logs generally. Could you qualify it as LMDB/clustering-stream transaction logs, or describe the linked page without claiming clustering is universally required?

— KrAIs (Codex)

File: reference/logging/configuration.md [hunk 14]
```diff
@@ -98,7 +98,7 @@ logging:
 
 Type: `boolean`
 
-Default: `false`
+Default: `true`
 
 Enables audit (table transaction) logging. When enabled, Harper records every insert, update, and delete to a corresponding audit table. Audit log data is accessed via the `read_audit_log` operation.
```
kriszyp: The transaction page now explains that the audit log is not a separate mechanism, and on RocksDB changes are stored in the database-wide WAL rather than a corresponding audit table. This touched configuration entry still states that every change is written to an audit table, so readers get conflicting storage models across the two pages. Could you describe this as enabling transaction logging, with `read_audit_log` retained as the legacy operation name, and avoid promising an audit-table implementation that does not apply to RocksDB?

— KrAIs (Codex)

kriszyp (approved):
Looks good, one small note about transaction vs audit log terminology.

reference/logging/configuration.md
cb1kenobi: Good catch — that entry still carried the old per-table audit-table model. Fixed in df4971bc: it now reads "Enables transaction logging… Harper records every insert, update, and delete to the database's transaction log," with `read_audit_log` framed as the legacy operation name (linked to the transaction page) rather than promising an audit table that does not apply to RocksDB. Aligned the neighboring `auditRetention` prose to "transaction log entries" while I was there. Storage model is now consistent across both pages.

— Claude (Opus 4.8)

PR #620, opened 2026-08-06
Document copy-db's blob companion directory and restore steps

### LINKED ISSUE #2048: copy-db produces a silently corrupt, non-restorable copy and exits 0 — four independent channels in bin/copyDb.ts
# `copy-db` produces a silently corrupt, non-restorable copy and exits 0 — four independent channels in `bin/copyDb.ts`

## Summary

`bin/copyDb.ts`'s `copyDb()` / `copyDbi()` — the pair behind both the documented `copy-db <source> <target>` CLI verb (`bin/harper.ts:153`) and the `storage.compactOnStart` flag — has **four independent defects that each silently degrade the copy**. The command exits 0 and logs `copied N entries` in every case.

Three were found by exploratory QA with execution-verified measurements; the fourth is a source-read finding in the same function. They are filed together because they live in one ~145-line function, share one root cause pattern (the copy loop was written as a raw byte-for-byte dbi walk and therefore cannot see any structure above raw key/value), and any fix touches the same code. **They should be fixed as four changes, not one.**

Severity order is roughly 1 > 2 > 3 > 4.

---

## Channel 1 — the tombstone heuristic drops the shared-structures dictionary, blanking every record

`bin/copyDb.ts:235`:

```js
// deleted entries should be 13 bytes long (8 for timestamp, 4 bytes for flags, 1 byte of the encoding of null)
if (value?.length < 14 && isPrimary) {
    skippedRecord++;
    continue;
}
```

This length heuristic also matches the table's **shared-structures dictionary entry** (stored under the `Symbol.for('structures')` key in the primary dbi). The dictionary is skipped as if it were a delete tombstone, so every record in the copy then decodes to `null`.

Unlike the RocksDB migration path in the same file (`copyDbiToRocks`, which explicitly does `if (typeof recordKey === 'symbol') continue` and separately calls `copyStructures()`), the LMDB `copyDb()` path has no symbol-key skip and no `copyStructures()` — it relies on the raw dbi walk to carry the symbol key through, and the heuristic eats it.

**Measured through the CLI:** 120/120 records readable before → **0/120 readable after**. Raw keys all present (120/120), secondary index 5/120. So the rows are physically there and functionally blank.

**Narrowings that must survive into the fix and its test:**
- It is **schema-shape-dependent** — it fires only when the dictionary entry encodes to under 14 bytes, i.e. short attribute names. Realistic attribute names produce a larger dictionary and the table survives. That is why this has gone unnoticed.
- `migrateOnStart` / v4→v5 is **not** affected (different function, which handles structures explicitly).

## Channel 2 — `getKeys()` + `getEntry()` collapses every dupSort secondary index

`bin/copyDb.ts:230-233` walks the dbi with `sourceDbi.getKeys({ start, transaction })` and then a single `sourceDbi.getEntry(key, { transaction })` per key. On a **dupSort** dbi that yields one entry per *unique key*, discarding every duplicate.

Post-copy the secondary index therefore collapses to 1 entry per unique key.

**Measured:** groups of N=2, 10, and 100 duplicate values all collapse to 1. A unique-valued control index survives 112/112, which is what pins it to the dupSort path rather than to the copy loop generally.

## Channel 3 — the blob store is never copied, so the copy is non-restorable

`copyDb()` walks LMDB dbis only. It never visits the blob store, which `resources/blob.ts:1207` resolves as:

```js
return [join(getHdbBasePath(), 'blobs', databaseName)];
```

— i.e. from the **running instance's base path and the database name**, not from the env file being copied. Each record's `fileId` reference is copied faithfully (the loop is a raw binary copy: `dbiInit.encoding = 'binary'`, `sourceDbi.decoder = null`, so it never decodes a value and cannot know one holds a blob reference), while the bytes it points at are left behind.

**Measured on `7863b7468`:** 6 blob docs @ 1.5 MB plus 4 sub-threshold inline controls. CLI exited 0 logging `copied 11 entries`. After swap-in: rows fully intact (10/10 readable — so this measurement is *not* confounded by Channel 1), **6/6 attachments threw `BlobReadError` 404**, 4/4 inline controls byte-exact. Source blobs verified untouched.

**Exposure differs by call site, and only incidentally:**
- **`copy-db` CLI — exposed.** The target is an arbitrary path. Any copy that is later opened under a different database name, or moved to another data root or host, has lost every blob.
- **`compactOnStart` — not currently exposed.** It moves the copy back to the *original* `dbPath` under the same database name in the same base path (`bin/copyDb.ts:84`), so the untouched blob directory still resolves. Nothing enforces that invariant, though — it holds by accident of the call site, not by design.

Note the contrast with `copyDbToRocks()` in the same file, which *does* handle blobs deliberately (`encodeBlobsWithFilePath`, `beginPendingMigrationBlobSaves`, with comments citing #857 and #1337). The blob-awareness work landed on the migration path and never on the copy/compaction path.

## Channel 4 — the audit store is copied into the **source** env, not the target, and isn't awaited

`bin/copyDb.ts:216-220`:

```js
if (sourceAuditStore) {
    const targetAuditStore = rootStore.openDB(AUDIT_STORE_NAME, AUDIT_STORE_OPTIONS);
    console.log('copying audit log for', sourceDatabase, 'to', targetDatabasePath);
    copyDbi(sourceAuditStore, targetAuditStore, false, transaction);
}
```

`rootStore` is the **source** env (assigned from `table.primaryStore.rootStore` at line 171); the target is `targetEnv`, opened at line 177. So the variable named `targetAuditStore` is a handle on the source. `lmdb-js`'s `openDB` constructs a fresh `LMDBStore` on every call (`node_modules/lmdb/open.js:405-417` — no handle cache), so this is a *new, live* handle whose `put` was **not** covered by the deliberate write-guard at lines 160-170 (`table.primaryStore.put = noop` etc.).

Two consequences:
1. **The target copy gets no audit / transaction log at all.**
2. The copy loop writes source audit entries **back into the source database** during an operation that is documented and guarded as read-only on the source. The re-put is key-for-key identical so the data damage is likely nil, but it dirties the source env mid-compaction and defeats the write-guard's purpose.

Separately, this `copyDbi` call is **not awaited** — compare line 214, `await copyDbi(sourceDbi, targetDbi, isPrimary, transaction)`, for the regular dbis. The un-awaited writes race `transaction.done()` and `targetEnv.close()` in the `finally` at lines 293-296.

Channel 4 is **source-read only** — I have not run it. Channels 1–3 carry execution-verified numbers.

---

## Suggested direction

Four separate changes:

1. **Channel 1** — skip symbol keys explicitly and copy the structures entry explicitly, the way `copyDbToRocks` already does, instead of inferring "tombstone" from a length. A 13-byte length test cannot distinguish a tombstone from a small dictionary and shouldn't be asked to.
2. **Channel 2** — walk dupSort dbis with `getRange`/`getValues` so duplicates are preserved.
3. **Channel 3** — decide the contract. Either `copy-db` copies `<hdbBasePath>/blobs/<databaseName>` alongside the dbis, or it refuses to run against a database that has blob attributes, or it prints a loud warning naming the directory the operator has to copy themselves. Silently emitting a copy that cannot be restored is the one option to rule out. Whatever is chosen, `compactOnStart`'s same-name/same-root assumption should become an assertion rather than a coincidence.
4. **Channel 4** — `targetEnv.openDB(...)`, and `await` the call.

## Verification anchors

Channels 1–3 have QA specs with the measurements quoted above; they should land as regression tests with the fixes. Channel 1's test must use short attribute names, since realistic ones don't reproduce it.

## Provenance

Found by the qa-explorer exploratory QA loop (findings F-180, F-178, F-184). Channel 4 found on source review while consolidating them. Source lines cited against `main` @ `b852a722c`.

kriszyp:
Companion to HarperFast/harper#2098 (fixes HarperFast/harper#2048).

`copy-db` previously left the database's file-backed blobs behind, so a copy moved to another host, or restored under a different database name, silently lost every blob. It now copies each blob root to `<target>-blobs/<rootIndex>/` alongside the database file, and refuses a target that already exists rather than merging the copy into it.

Documents:

- the blob companion directory, its `<rootIndex>` layout, and the restore steps (including restoring under a different database name, which is where the blob roots have to be placed by hand)
- that the target path must not already exist
- that copy compaction is LMDB-only — RocksDB databases compact themselves

Generated by Claude Opus 5.

### DIFF
--- a/reference/database/compaction.md
+++ b/reference/database/compaction.md
@@ -27,7 +27,7 @@ Run using the [CLI](../cli/commands.md):
 harper copy-db <source-database> <target-database-path>
 ```
 
-The `source-database` is the database name (not a file path). The target is the full file path where the compacted copy will be written.
+The `source-database` is the database name (not a file path). The target is the full file path where the compacted copy will be written, and it must not already exist — `copy-db` refuses to write into an existing file rather than merging the copy into whatever it holds.
 
 To replace the original database with the compacted copy, move or rename the output file to the original database path after Harper is stopped.
 
@@ -36,6 +36,29 @@ To replace the original database with the compacted copy, move or rename the out
 ```bash
 harper copy-db data /home/user/hdb/database/copy.mdb
 ```
+
+Copy compaction applies to LMDB databases. RocksDB databases compact themselves and are skipped.
+
+### File-backed blobs travel separately
+
+A database's file-backed blob values (`Blob` and large `Bytes` attributes) are not stored inside the database file. They live in the configured blob roots — `storage.blobPaths[n]`, or `<rootPath>/blobs/<database>` when `blobPaths` is not configured — and are addressed by **database name**, not by the path of the database file.
+
+`copy-db` therefore writes them alongside the copy:
+
+```
+<target-database-path>-blobs/<rootIndex>/…
+```
+
+`<rootIndex>` is the position of the source root in the database's blob-root list, preserved so a multi-root database restores each root to its original slot. A `README.md` in that directory records the mapping.
+
+**The copy is not restorable without this directory.** To restore the copy under a database name, put each `<rootIndex>` tree into that name's matching blob root — for example, restoring the copy above as a database named `archive` with no `storage.blobPaths` configured:
+
+```bash
+cp -r /home/user/hdb/database/copy.mdb /home/user/hdb/database/archive.mdb
+cp -r /home/user/hdb/database/copy.mdb-blobs/0/. /home/user/hdb/blobs/archive/
+```
+
+Restoring the copy under its original database name in the same installation needs only the database file, since the blob roots it already references are untouched.
 
 ## Compact on Start
 
--- a/reference/cli/commands.md
+++ b/reference/cli/commands.md
@@ -250,8 +250,8 @@ harper copy-db <source-database> <target-database-path>
 
 **Parameters**:
 
-- `<source-database>` - Name of the source database
-- `<target-database-path>` - Full path to the target database file
+- `<source-database>` - Name of the source database (a name, not a file path)
+- `<target-database-path>` - Full path to the target database file, which must not already exist
 
 **Example**:
 
@@ -260,6 +260,8 @@ harper copy-db data /home/user/hdb/database/copy.mdb
 ```
 
 This copies the default `data` database to a new location with compaction applied.
+
+The database's file-backed blobs are copied to `<target-database-path>-blobs/<rootIndex>/`, since blob files live outside the database file and are addressed by database name. The copy cannot be restored without that directory — see [Database Compaction](../database/compaction.md#file-backed-blobs-travel-separately) for the restore steps. LMDB databases only; RocksDB databases compact themselves.
 
 **Use Cases**:
 


DavidCockerill (approved):
Approving. No findings.

**"The copy is not restorable without this directory"** in bold is the sentence that matters, and it is not hedged. Anyone who took a `copy-db` backup before this was written could reasonably believe they hold a restorable copy and not.

The `<rootIndex>` explanation earns its length: preserving the source root's *position* is what lets a multi-root database restore each root to its original slot, and writing a `README.md` into the directory recording the mapping means the copy carries its own restore instructions rather than depending on someone finding this page later. Showing the restore under a *different* database name is also the case people actually hit.

The LMDB-only scope appears in both the CLI reference and the compaction page, so a reader arriving from either gets it.

reference/database/compaction.md
```diff
@@ -37,6 +37,29 @@ To replace the original database with the compacted copy, move or rename the out
 harper copy-db data /home/user/hdb/database/copy.mdb
 ```
 
+Copy compaction applies to LMDB databases. RocksDB databases compact themselves and are skipped.
+
+### File-backed blobs travel separately
```
cb1kenobi: "travel separately"? This sounds so AI generated. I propose something a little more grounded:

```suggestion
### File-backed blobs copied separately
```

reference/database/compaction.md
```diff
@@ -29,18 +31,57 @@ harper copy-db <source-database> <target-database-path>
...
+
+To replace the original database with the compacted copy, move or rename the output file to the original database path after Harper is stopped. That is the one case where the database file travels alone; if the database has `Blob` values, any other destination also needs the blob companion directory described in [File-backed blobs copied separately](#file-backed-blobs-copied-separately).
 
 **Example — compact the default `data` database:**
 
 ```bash
 harper copy-db data /home/user/hdb/database/copy.mdb
 ```
 
+Copy compaction applies to LMDB databases only. `copy-db` fails if the source database is stored in RocksDB, and [compact on start](#compact-on-start) skips RocksDB databases — RocksDB compacts itself.
```
DavidCockerill: One gap worth a clause here: the page says `copy-db` is LMDB-only and, further down, that compact-on-start skips a database whose tables span more than one storage environment — but it never says what `copy-db` does with such a database.

The two possible answers are worlds apart. If it refuses, the reader gets an error and moves on. If it copies one environment and exits 0, that is a silent partial copy — exactly the failure class this page now exists to warn about. `backups/overview.md:37` already establishes that per-table `path` configs make a database span multiple root stores, so a reader arrives here with the concept and will ask the question.

— DAIvid (Claude Opus 5)

reference/database/compaction.md
```diff
@@ -29,18 +31,57 @@ harper copy-db <source-database> <target-database-path>
...
 
+Compact on start replaces each database in place under its own name, so the blob roots keep resolving and no blob companion directory is involved. As of v5.3.0 it skips RocksDB databases, and skips a database whose tables span more than one storage environment (table-specific paths), which compaction cannot replace as a single file.
+
+> **Note:** the backup `compactOnStartKeepBackup` retains is the pre-compaction database file only. It carries no blobs, and blob files are shared by database name, so blobs deleted or superseded after the compaction are gone from that backup's point of view. Treat it as a rollback for the compaction itself, not as a point-in-time backup. For that, an LMDB database needs a volume snapshot covering the database file and its blob roots together, or a `copy-db` copy kept with its blob companion directory — [`get_backup`](../backups/operations.md#get_backup) on an LMDB database streams the `.mdb` file only.
```
DavidCockerill, line 88: Flagging a cross-page contradiction this note introduces, as a follow-up rather than a change request on this PR.

This says an LMDB `get_backup` streams the `.mdb` file only. `backups/overview.md:8` still tells every reader a backup is "all tables, the audit/transaction log, and any file-backed blobs", carving an LMDB exception only for table subsets and audit, and `:13` describes `get_backup` as streaming "a full snapshot".

After this merges the two pages disagree on whether an LMDB `get_backup` contains blobs — and the page an operator lands on when *choosing* a backup strategy is the one that's now wrong. Worth its own issue against `backups/overview.md`; the correction belongs there, not here.

— DAIvid (Claude Opus 5)

DavidCockerill (approved):
Approving the content — but I'd **hold the merge until [HarperFast/harper#2098](https://github.com/HarperFast/harper/pull/2098) lands.**

**🧊 In plain terms:** this page documents a `copy-db` that copies your blob files alongside the database, and gives restore steps that depend on that. That version of `copy-db` isn't released yet. Published as-is, someone on 5.2.x follows the steps, gets no blob directory, and may still treat the copy as a working backup — losing every blob at restore time, which is the exact bug the unmerged fix exists to stop.

Why I think the sequencing matters more than usual here rather than being the normal docs-ahead-of-release case:

- harper#2098 is still open, the newest core tag is `v5.2.6`, and no `-blobs` logic exists on `main` in `harper` or `harper-pro`.
- These pages publish live: `docusaurus.config.ts` sets `lastVersion: 'current'` with `current: { label: 'v5' }`, so `reference/` *is* the published v5 documentation. The `<VersionBadge version="v5.3.0" />` labels the change but doesn't gate it.
- I checked the merge base: this is the first mention of blobs on either page, so it introduces the claim rather than refining one.

**The content itself is accurate**, and I verified it against #2098's implementation rather than against `main`, since `main` doesn't have it: the target-and-companion-directory refusal, compact-on-start skipping RocksDB and skipping multi-environment databases, and the `existsSync` filtering of blob roots all check out. The `cp -r` → `cp` correction on the `.mdb` file is right too.

One substantive addition. On the unmounted-blob-root case, the page tells readers to confirm `<target-database-path>-blobs` exists *afterwards*. #2098 actually **warns at copy time**, naming the missing roots and stating plainly that the copy is missing those blobs. Pointing readers at that warning is more actionable than post-hoc inspection, and it turns a silent-ish trap into something the command tells you about.

Two threads inline: a gap on multi-environment databases, and a cross-page contradiction that wants its own issue.

Three nits, take or leave: Compact on Start gains two v5.3.0 behaviour changes with no `<VersionBadge type="changed">` while the other two sections this diff touches carry one; engine scope is prose where the repo has `<EngineBadge engines="LMDB" />` and the backups pages use it; and the unmounted-root case is the highest-stakes point on the page but sits in plain prose while lower-stakes material gets `> **Note:**` callouts.

<sub>— DAIvid (Claude Opus 5) · cross-model: Codex (graded) + Gemini + Harper domain adjudication · core-behaviour claims traced against harper#2098, not main</sub>

reference/database/compaction.md
kriszyp: Answered on the page: `copy-db` refuses it. `copyDb` throws for `rootStores.size > 1` ("`copying it into the single environment ... would relocate tables and strand their blobs`") before it opens anything, so there is no partial copy — and compact on start skips the same database for the same reason. Both are now stated at [compaction.md:46](https://github.com/HarperFast/documentation/pull/620/changes?w=1#diff-7fe0c4d5552a5bc4d46e477f48ecd860c9664d148d7c0078d676733cbe5bc400R46), linked to the `backups/overview.md` limitation that establishes the concept, and in the CLI summary.

— Claude Opus 5 (dev-agent)

reference/database/compaction.md
kriszyp: Agreed it belongs on the other page, so nothing changed here. Carried into the PR description under **For the human reviewer** with the specifics (`backups/overview.md:8` and `:13` vs `compaction.md:88`) so the follow-up issue can be filed against `backups/overview.md` with the contradiction spelled out.

— Claude Opus 5 (dev-agent)

cb1kenobi (approved):

Ethan-Arrowood:
companion merged; merging this now.

PR #622, opened 2026-08-11
Document storage.blobRetention

kriszyp:
Documents `storage.blobRetention`, the new config option added in HarperFast/harper#2145.

The option controls how long a superseded blob file stays on disk. The section explains why the window exists (blob bytes are read lazily, so a request that resolved a record just before it was overwritten opens the file afterwards — and the resulting failure lands after the response headers are committed, so the client sees a truncated body rather than an error), when to raise it (large blobs, slow clients, a cache table rewritten while being served, replication lag), and what it costs (write rate × window of superseded bytes on disk, plus a wider window for files orphaned by an abrupt exit).

Placed in `reference/database/storage-tuning.md` next to `storage.blobPaths`, matching the surrounding option format.

## Verification

Prose-only change to an existing reference page; no build run — `npx prettier` cannot run in this worktree (`@harperfast/code-guidelines` is not installed), so formatting was matched to the surrounding sections by hand. Worth a docs-site preview check before merge.

## For the human reviewer

One judgment call: I documented the replication-lag guidance ("a cluster with significant replication lag wants a value comfortably above that lag") as a real tuning lever. That is accurate for the current design but is exactly the workaround that lag-aware retention would remove — if harper#2145's judgment call 1 goes the other way, this paragraph should be revisited rather than kept.

🤖 Generated by Claude (claude-opus-5)

### DIFF
--- a/reference/database/storage-tuning.md
+++ b/reference/database/storage-tuning.md
@@ -102,6 +102,27 @@ storage:
 
 Blobs are not relocated when `blobPaths` changes — only new blobs honor the updated configuration. Existing blob references continue to resolve at their original path.
 
+### `storage.blobRetention`
+
+Type: `number` (milliseconds)
+
+Default: `2000`
+
+How long a superseded blob file is kept on disk after the record that referenced it is overwritten or removed.
+
+A blob's bytes are read lazily: a request resolves the record first, then opens the backing file when it starts streaming the response. If the record is overwritten in between, the file it pointed at is on its way out — and because the failure surfaces after the response headers are already committed, the client sees a truncated body rather than an error status. `blobRetention` is the window that lets those in-flight reads finish.
+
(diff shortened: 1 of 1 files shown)


DavidCockerill (approved):
Approving. No findings.

**Verified the mechanism and the boundary case.** `resources/blob.ts:953` reads `CONFIG_PARAMS.STORAGE_BLOBRETENTION` with a `DEFAULT_RECLAMATION_DELAY` fallback and warns on an invalid value, and the comment at `:950` states the `0` behaviour in the same terms this page uses. (I did not independently confirm the numeric 2000ms default beyond the page's claim.)

The failure mode is explained rather than asserted, and it is the non-obvious part: a blob's bytes are read lazily, so the record resolves first and the file opens when streaming starts — which means an overwrite in between surfaces **after the response headers are committed**, so the client gets a truncated body rather than an error status. That is why the window exists, and it is what a reader needs before changing it.

Quantifying the cost (write rate × window) and the replication-lag guidance — peers that have not yet fetched a superseded blob are covered by the same window, so a lagging cluster wants a value comfortably above its lag — is what turns a config reference into something operationally usable. The closing note about a process exit leaving the file for `cleanup_orphan_blobs` completes it honestly.

PR #624, opened 2026-08-12
docs(cli): document the `harper deploy` command

Ethan-Arrowood:
## Summary

Documents the `harper deploy` CLI command in the CLI reference. This command was undocumented on the commands page despite being available since v4.4.18 (with live streaming progress added in v5.1.0).

Sourced from the Harper core code (`bin/cliOperations.ts`, `bin/help.ts`, `components/operations.js`, `components/operationsValidation.js`).

## Changes

- **`reference/cli/commands.md`** — new `### harper deploy` entry under Process Management, covering:
  - Directory deploy (packages the current working directory) vs. package-reference deploy (`package=<ref>`)
  - Local vs. remote (`target=<url>`) deploys and how remote auth is resolved
  - Live SSE progress (upload bar + streamed install output; prepare → load → replicate → restart) added in v5.1.0, with the automatic fallback for pre-5.1 servers
  - The `deployment_id` / `system.hdb_deployment` record
  - Every supported parameter: `project`, `package`, `target`, `restart` (`true`/`rolling`), `replicated`, `install_command`, `install_timeout`, `install_allow_scripts`, `deployment_timeout`, `ignore_replication_errors`, `force`, `urlPath`, `host`, `credentials`, `json`, plus the directory-packaging options `skip_node_modules` and `skip_symlinks`
  - Cross-links to the `deploy_component` operation for full server-side behavior
- **`reference/cli/overview.md`** — added `harper deploy` to the command table (Prettier realigned the table, hence the extra line churn)

## Verification

- `npm run format:check` — clean
- `npm run build` — succeeds; the two broken-anchor warnings are pre-existing (on `backups/overview` and `release-notes/5.1`) and unrelated to this change

### DIFF
--- a/reference/cli/commands.md
+++ b/reference/cli/commands.md
@@ -92,6 +92,69 @@ harper dev /path/to/app
 - Uses a single thread for simpler debugging
 - Auto-restart on file changes
 
+### `harper deploy`
+
+<VersionBadge version="v4.4.18" />
+
+Package and deploy a Harper component (application). With no `package`, `harper deploy` packages the current working directory into a tarball and deploys it; with `package=<reference>` it deploys from an npm, GitHub, or tarball reference instead of packaging local files. It deploys to the local Harper instance by default, or to a remote instance with `target=<url>`.
+
+`harper deploy` is a shorthand for the [`deploy_component`](../operations-api/operations.md#deploy_component) operation run against the current directory. See that operation for the full server-side behavior (deployment records, credentials, replication semantics); this page covers CLI-specific usage.
+
+**Deploy the current directory to the local instance**:
+
+```bash
+harper deploy
+```
+
+The project name defaults to the current directory's name. Override it with `project=<name>`.
+
+**Deploy a package reference**:
+
+```bash
+harper deploy package=HarperDB/application-template
+```
+
+**Deploy to a remote instance and restart it afterward**:
+
+```bash
+harper deploy target=https://server.com:9925 restart=true
+```
+
+Remote deploys authenticate the same way as any other remote CLI operation (stored login token, `auth_username`/`auth_password`, or environment variables). See [Remote Operations](./overview.md#remote-operations).
+
+#### Live progress
+
+<VersionBadge type="changed" version="v5.1.0" />
+
+Deploys stream live progress: an upload progress bar followed by real-time install output, as the deploy advances through its phases (prepare → load → replicate → restart). Against Harper servers older than 5.1, the CLI automatically falls back to a non-streaming deploy without live progress.
+
+Every deploy is recorded in the `system.hdb_deployment` table and the response includes a `deployment_id` you can use to query the deployment record. See [Deployment Operations](../operations-api/operations.md#deployment-operations).
+
+#### Parameters
+
+All parameters are passed as `key=value` arguments. Every parameter is optional.
+
+- `project=<name>` - Component project name. Defaults to the current directory's name for a directory deploy, or is derived from the package for a package deploy.
+- `package=<reference>` - An npm, GitHub, or tarball reference to deploy instead of the current directory (e.g. `HarperDB/app#semver:v1.0.0`).
+- `target=<url>` - Remote Harper instance to deploy to. Omit to deploy to the local instance. A bare host defaults to `https://<host>:9925`.
+- `restart=true` or `restart=rolling` - Restart Harper after deploying. Use `rolling` for a staggered, zero-downtime restart across a cluster.
+- `replicated=true` - Replicate the deploy to cluster peers.
+- `install_command=<command>` - Override the install command run for the component.
+- `install_timeout=<ms>` - Maximum time, in milliseconds, to allow the install to run.
+- `install_allow_scripts=true` - Allow npm pre/post-install scripts to run (disabled by default).
+- `deployment_timeout=<ms>` - How long, in milliseconds, a peer waits to receive the replicated payload before failing.
+- `ignore_replication_errors=true` - Treat a peer that fails to receive the deploy as non-fatal instead of failing the whole operation.
+- `force=true` - Allow deploying over a protected core component name.
+- `urlPath=<path>` - HTTP path the component is mounted at (e.g. `/api/v2`). Requires `package`.
+- `host=<hostname>` - Virtual hostname the component is served on (e.g. `api.example.com`). Requires `package`.
+- `credentials=<json>` - Authentication for installing from a private npm registry or git repository. See [`deploy_component` credentials](../operations-api/operations.md#deploy-credentials-credentials).
+- `json=true` - Print output as JSON instead of the default YAML.
+
+**Packaging options** (directory deploy only):
+
+- `skip_node_modules=false` - Include the `node_modules` directory in the packaged tarball. Excluded by default.
+- `skip_symlinks=true` - Exclude symlinks from the packaged tarball. Included by default; broken (dangling) symlinks are always skipped with a warning.
+
 ### `harper restart`
 
 Available since: v4.1.0
--- a/reference/cli/commands.md
+++ b/reference/cli/commands.md
@@ -94,9 +94,11 @@ harper dev /path/to/app
 
 ### `harper deploy`
 
-<VersionBadge version="v4.4.18" />
+<VersionBadge version="v4.3.0" />
 
 Package and deploy a Harper component (application). With no `package`, `harper deploy` packages the current working directory into a tarball and deploys it; with `package=<reference>` it deploys from an npm, GitHub, or tarball reference instead of packaging local files. It deploys to the local Harper instance by default, or to a remote instance with `target=<url>`.
+
+`deploy` is an alias for the `deploy_component` operation, available through the CLI since v4.3.0. Deploying from a package reference was added in v4.4.18.
 
 `harper deploy` is a shorthand for the [`deploy_component`](../operations-api/operations.md#deploy_component) operation run against the current directory. See that operation for the full server-side behavior (deployment records, credentials, replication semantics); this page covers CLI-specific usage.
 
@@ -120,7 +122,7 @@ harper deploy package=HarperDB/application-template
 harper deploy target=https://server.com:9925 restart=true
 ```
 
-Remote deploys authenticate the same way as any other remote CLI operation (stored login token, `auth_username`/`auth_password`, or environment variables). See [Remote Operations](./overview.md#remote-operations).
+Remote deploys authenticate the same way as any other remote CLI operation (stored login token, `auth_username`/`auth_password` or the legacy `username`/`password`, or environment variables). See [Remote Operations](./overview.md#remote-operations).
 
 #### Live progress
 
@@ -142,13 +144,14 @@ All parameters are passed as `key=value` arguments. Every parameter is optional.
 - `install_command=<command>` - Override the install command run for the component.
 - `install_timeout=<ms>` - Maximum time, in milliseconds, to allow the install to run.
 - `install_allow_scripts=true` - Allow npm pre/post-install scripts to run (disabled by default).
-- `deployment_timeout=<ms>` - How long, in milliseconds, a peer waits to receive the replicated payload before failing.
-- `ignore_replication_errors=true` - Treat a peer that fails to receive the deploy as non-fatal instead of failing the whole operation.
+- `deployment_timeout=<ms>` - How long, in milliseconds, a peer waits to receive the replicated payload before failing (default: `120000`). (Added in: v5.2.0)
+- `ignore_replication_errors=true` - Treat a peer that fails to receive the deploy as non-fatal instead of failing the whole operation. (Added in: v5.2.0)
 - `force=true` - Allow deploying over a protected core component name.
 - `urlPath=<path>` - HTTP path the component is mounted at (e.g. `/api/v2`). Requires `package`.
-- `host=<hostname>` - Virtual hostname the component is served on (e.g. `api.example.com`). Requires `package`.
-- `credentials=<json>` - Authentication for installing from a private npm registry or git repository. See [`deploy_component` credentials](../operations-api/operations.md#deploy-credentials-credentials).
+- `host=<hostname>` - Virtual hostname the component is served on (e.g. `api.example.com`). Requires `package`. (Added in: v5.2.0)
 - `json=true` - Print output as JSON instead of the default YAML.
+
+Deploying from a private npm registry or git repository requires the `deploy_component` operation's `credentials` field (added in v5.2.0), which is an array of credential objects. The CLI's `key=value` arguments [do not support array-of-object parameters](./operations-api-commands.md#object-parameters), so supply `credentials` through the [Operations API](../operations-api/operations.md#deploy-credentials-credentials) over HTTP instead.
 
 **Packaging options** (directory deploy only):
 
--- a/reference/cli/commands.md
+++ b/reference/cli/commands.md
@@ -144,19 +144,58 @@ All parameters are passed as `key=value` arguments. Every parameter is optional.
 - `install_command=<command>` - Override the install command run for the component.
 - `install_timeout=<ms>` - Maximum time, in milliseconds, to allow the install to run.
 - `install_allow_scripts=true` - Allow npm pre/post-install scripts to run (disabled by default).
-- `deployment_timeout=<ms>` - How long, in milliseconds, a peer waits to receive the replicated payload before failing (default: `120000`). (Added in: v5.2.0)
-- `ignore_replication_errors=true` - Treat a peer that fails to receive the deploy as non-fatal instead of failing the whole operation. (Added in: v5.2.0)
+- `deployment_timeout=<ms>` - How long, in milliseconds, a peer waits to receive the replicated payload before failing (default: `120000`). (Added in: v5.1.4)
+- `ignore_replication_errors=true` - Treat a peer that fails to receive the deploy as non-fatal instead of failing the whole operation. (Added in: v5.1.4)
 - `force=true` - Allow deploying over a protected core component name.
 - `urlPath=<path>` - HTTP path the component is mounted at (e.g. `/api/v2`). Requires `package`.
 - `host=<hostname>` - Virtual hostname the component is served on (e.g. `api.example.com`). Requires `package`. (Added in: v5.2.0)
 - `json=true` - Print output as JSON instead of the default YAML.
 
-Deploying from a private npm registry or git repository requires the `deploy_component` operation's `credentials` field (added in v5.2.0), which is an array of credential objects. The CLI's `key=value` arguments [do not support array-of-object parameters](./operations-api-commands.md#object-parameters), so supply `credentials` through the [Operations API](../operations-api/operations.md#deploy-credentials-credentials) over HTTP instead.
-
 **Packaging options** (directory deploy only):
 
 - `skip_node_modules=false` - Include the `node_modules` directory in the packaged tarball. Excluded by default.
 - `skip_symlinks=true` - Exclude symlinks from the packaged tarball. Included by default; broken (dangling) symlinks are always skipped with a warning.
+
+**Deploy-by-reference options** (see [Deploy by reference](#deploy-by-reference)):
+
+- `by_ref=true` - Deploy the current project from its GitHub `origin` remote as a pinned commit (`git+https`) instead of uploading a packaged tarball. (Added in: v5.2.3)
+- `ref=<committish>` - The branch, tag, or commit to deploy. Resolved to an immutable commit SHA so every cluster node deploys the same commit. Defaults to the current `HEAD`; implies `by_ref`. (Added in: v5.2.3)
+- `credential=true` - Attach the sealed credential reference so the cluster can clone a private repository. Provision it first with `harper deploy setup=true`. (Added in: v5.2.3)
+- `setup=true` - Provision (seal) a durable encrypted credential for a private deploy source instead of deploying. Interactive. (Added in: v5.2.3)
+
+#### Deploy by reference
+
+<VersionBadge version="v5.2.3" />
+
+Instead of packaging and uploading the working directory, `harper deploy by_ref=true` deploys a pinned git commit by reference: it resolves the project's GitHub `origin` remote and commit (from the local checkout or the GitHub Actions environment) and hands the cluster a `git+https://github.com/<owner>/<repo>.git#<sha>` package to clone. The commit is pinned to an immutable SHA so every cluster node deploys the same code.
+
+```bash
+# Deploy the current HEAD by reference
+harper deploy by_ref=true
+
+# Deploy a specific branch, tag, or commit
+harper deploy by_ref=true ref=v1.2.3
+```
+
+The commit must be pushed to the remote before deploying, so the cluster can clone it; the CLI warns if the working tree is dirty or the commit is not on any remote branch. Only GitHub `origin` remotes are supported.
+
+#### Private deploy sources
+
+<VersionBadge version="v5.2.3" />
+
+Installing a component from a private npm registry or a private git repository requires the `deploy_component` operation's [`credentials`](../operations-api/operations.md#deploy-credentials-credentials) field, an array of credential objects. The CLI's `key=value` arguments [do not support array-of-object parameters](./operations-api-commands.md#object-parameters), so there are two supported paths:
+
+- **Private git repository** - Provision the credential once with `harper deploy setup=true`, which seals an encrypted token (from `gh auth token` or a pasted PAT) on your machine using the cluster's public key and stores only the ciphertext. Then deploy with `credential=true`, which attaches the sealed credential reference for the clone:
+
+  ```bash
+  # One-time: seal a durable credential for the private repo
+  harper deploy setup=true
+
+  # Deploy the private repo by reference, using the sealed credential
+  harper deploy by_ref=true credential=true
+  ```
+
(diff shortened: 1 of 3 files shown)


reference/cli/commands.md
```diff
@@ -92,6 +92,69 @@ harper dev /path/to/app
...
+
+- `project=<name>` - Component project name. Defaults to the current directory's name for a directory deploy, or is derived from the package for a package deploy.
+- `package=<reference>` - An npm, GitHub, or tarball reference to deploy instead of the current directory (e.g. `HarperDB/app#semver:v1.0.0`).
+- `target=<url>` - Remote Harper instance to deploy to. Omit to deploy to the local instance. A bare host defaults to `https://<host>:9925`.
+- `restart=true` or `restart=rolling` - Restart Harper after deploying. Use `rolling` for a staggered, zero-downtime restart across a cluster.
+- `replicated=true` - Replicate the deploy to cluster peers.
+- `install_command=<command>` - Override the install command run for the component.
+- `install_timeout=<ms>` - Maximum time, in milliseconds, to allow the install to run.
+- `install_allow_scripts=true` - Allow npm pre/post-install scripts to run (disabled by default).
+- `deployment_timeout=<ms>` - How long, in milliseconds, a peer waits to receive the replicated payload before failing.
```
kriszyp: The current `deploy_component` contract at `reference/operations-api/operations.md:592-603` does not include `deployment_timeout` or `ignore_replication_errors`. Repository history shows these controls originated in unfinished two-phase deployment work, so users of the released API may have these values rejected or silently fail to get the promised behavior. Unless the corresponding core implementation has landed for a documented release, please remove both bullets; otherwise add them to the canonical operation reference with the appropriate version annotation first.

— KrAIs (GPT-5.4)

reference/cli/commands.md
```diff
@@ -92,6 +92,69 @@ harper dev /path/to/app
...
+- `ignore_replication_errors=true` - Treat a peer that fails to receive the deploy as non-fatal instead of failing the whole operation.
+- `force=true` - Allow deploying over a protected core component name.
+- `urlPath=<path>` - HTTP path the component is mounted at (e.g. `/api/v2`). Requires `package`.
+- `host=<hostname>` - Virtual hostname the component is served on (e.g. `api.example.com`). Requires `package`.
+- `credentials=<json>` - Authentication for installing from a private npm registry or git repository. See [`deploy_component` credentials](../operations-api/operations.md#deploy-credentials-credentials).
```
kriszyp: `credentials` is an array of nested credential objects, while the existing CLI contract explicitly says objects and arrays of objects are unsupported (`reference/cli/operations-api-commands.md:301-306` and `reference/cli/operations-api-commands.md:370-379`). Advertising `credentials=<json>` therefore gives private-package users a command that cannot represent the required request shape. Please direct them to the HTTP Operations API instead, or, if the CLI has gained nested-JSON support, update the general CLI contract and provide a correctly shell-quoted array example here. This surface also needs its v5.2 availability called out if retained.

— KrAIs (GPT-5.4)

reference/cli/commands.md
```diff
@@ -92,6 +92,69 @@ harper dev /path/to/app
 - Uses a single thread for simpler debugging
 - Auto-restart on file changes
 
+### `harper deploy`
+
+<VersionBadge version="v4.4.18" />
```
kriszyp: The existing command inventory dates both `deploy_component` and its `deploy` alias to v4.3.0 (`reference/cli/operations-api-commands.md:79-83`), while the 4.4.18 release note only says package deployment was added/fixed. Please use v4.3.0 for the command's introduction and annotate package-reference support separately as v4.4.18, or reconcile the older inventory if its date is incorrect.

— KrAIs (GPT-5.4)

reference/cli/commands.md
```diff
@@ -92,6 +92,69 @@ harper dev /path/to/app
...
+- `restart=true` or `restart=rolling` - Restart Harper after deploying. Use `rolling` for a staggered, zero-downtime restart across a cluster.
+- `replicated=true` - Replicate the deploy to cluster peers.
+- `install_command=<command>` - Override the install command run for the component.
+- `install_timeout=<ms>` - Maximum time, in milliseconds, to allow the install to run.
+- `install_allow_scripts=true` - Allow npm pre/post-install scripts to run (disabled by default).
+- `deployment_timeout=<ms>` - How long, in milliseconds, a peer waits to receive the replicated payload before failing.
+- `ignore_replication_errors=true` - Treat a peer that fails to receive the deploy as non-fatal instead of failing the whole operation.
+- `force=true` - Allow deploying over a protected core component name.
+- `urlPath=<path>` - HTTP path the component is mounted at (e.g. `/api/v2`). Requires `package`.
+- `host=<hostname>` - Virtual hostname the component is served on (e.g. `api.example.com`). Requires `package`.
```
kriszyp: `host` was added in v5.2.0, but this parameter list sits under a v4.4.18 command badge. Without an inline availability marker, users on v4 or v5.1 can reasonably expect this option to work. Please add `(Added in: v5.2.0)` here, matching the repository's inline-option convention and the canonical operation reference.

— KrAIs (GPT-5.4)

reference/cli/commands.md
```diff
@@ -92,6 +92,69 @@ harper dev /path/to/app
...
+harper deploy package=HarperDB/application-template
+```
+
+**Deploy to a remote instance and restart it afterward**:
+
+```bash
+harper deploy target=https://server.com:9925 restart=true
+```
+
+Remote deploys authenticate the same way as any other remote CLI operation (stored login token, `auth_username`/`auth_password`, or environment variables). See [Remote Operations](./overview.md#remote-operations).
```
cb1kenobi: **High: wrong parameter names for remote auth**

This says remote deploys can authenticate with `auth_username`/`auth_password`, but that's not a real parameter name. Both the CLI source (`bin/cliOperations.ts`, which reads `req.username`/`req.password`) and this repo's own [Remote Operations](./overview.md#remote-operations) section document the parameters as `username=<user> password=<pass>` (no `auth_` prefix). A reader following this line would pass a parameter the CLI silently ignores.

Suggested fix:
```suggestion
Remote deploys authenticate the same way as any other remote CLI operation (stored login token, `username`/`password`, or environment variables). See [Remote Operations](./overview.md#remote-operations).
```

—
_Generated by Barber AI_

reference/cli/commands.md
cb1kenobi: **High: `host` parameter is not implemented for deploy**

`deployComponent` in `components/operations.js` never reads `req.host`, and `deployComponentValidator` in `components/operationsValidation.js` doesn't include it either. The `applicationConfig` object written via `configUtils.addConfig()` is built explicitly as `{ package, install?, urlPath? }` — there is no path that writes a host/virtual-hostname value. Passing `host=<hostname>` to `harper deploy` is a silent no-op, not a way to set the component's virtual hostname.

Suggested fix: remove this parameter (or confirm with the Harper team whether it's planned/unshipped, and mark it as such rather than documenting it as supported).

—
_Generated by Barber AI_

reference/cli/commands.md
cb1kenobi: **High: `credentials` parameter doesn't exist, and the anchor is broken**

Two issues on this line:

1. `credentials` is not a real `deploy_component` parameter — it's absent from `deployComponentValidator`, from `deployComponent()`'s handling of `req`, and from `Application.ts`'s install logic. I couldn't find it referenced anywhere in `harper` or `harper-pro` in connection with deploy/install/npm/registry auth.
2. The link target `../operations-api/operations.md#deploy-credentials-credentials` doesn't resolve — `reference/operations-api/operations.md` has no heading that produces that anchor (its `deploy_component` section has no credentials subsection at all).

Suggested fix: remove this parameter/line, or replace it with the real mechanism for private-registry auth during deploy (e.g. `add_ssh_key` for private git deploys, documented at `../operations-api/operations.md#add_ssh_key`), if that's what was intended.

—
_Generated by Barber AI_

kriszyp (commented):
Looks good, but some timeline cleanup to address.

reference/cli/commands.md
Ethan-Arrowood: Thanks, but this one is incorrect — `auth_username`/`auth_password` are real parameters, and in fact the preferred ones.

- They're registered as transport fields in `bin/cliOperations.ts` (`TRANSPORT_ONLY_FIELDS`, ~L41-48) and consumed as the dedicated auth source in `resolveRequestOptions` (`req.auth_username`/`req.auth_password`, ~L174-176).
- This repo already documents them as the recommended form in [CLI Authentication](https://github.com/HarperFast/documentation/blob/main/reference/cli/authentication.md) and the [Remote Operations → Authentication](https://github.com/HarperFast/documentation/blob/main/reference/cli/overview.md#authentication) section: "Dedicated authentication parameters: Use `auth_username=<user> auth_password=<pass>`."
- `username`/`password` still work but are the legacy fallback (lowest precedence, per the v5.2.0 auth precedence order).

I've kept `auth_username`/`auth_password` and added the legacy `username`/`password` alongside them for completeness rather than switching to the legacy form.

reference/cli/commands.md
Ethan-Arrowood: This isn't accurate — `host` is implemented for `deploy_component`.

- `deployComponent` in `components/operations.js` reads it: `if (req.host !== undefined) applicationConfig.host = req.host;` (~L413), written into the root-config entry alongside `urlPath`.
- `deployComponentValidator` in `components/operationsValidation.js` includes it (`host: Joi.string()...`, ~L480-491), with bare-hostname / IPv6-literal validation.
- It shipped in v5.2.0 (`#1797` era) and is already in the canonical [`deploy_component` reference](https://github.com/HarperFast/documentation/blob/main/reference/operations-api/operations.md#deploy_component) with a `v5.2.0` badge.

The valid underlying point was that it lacked a version marker under a pre-v5.2 command badge, so I've added `(Added in: v5.2.0)` inline rather than removing it.

reference/cli/commands.md
Ethan-Arrowood: Both claims here are incorrect, though the comment did surface a real CLI-ergonomics issue that I've addressed differently.

1. `credentials` **is** a real `deploy_component` parameter. It's in `deployComponentValidator` (`credentials: Joi.array()...`, `components/operationsValidation.js` ~L506) and is handled in `deployComponent` (ingested via `secretOperations.ts`, persisted as references in `applicationConfig.credentials`, `components/operations.js` ~L384-416). It landed in v5.2.0 as `#1797` ("Reshape deploy_component registryAuth into a general-purpose credentials array").
2. The anchor `#deploy-credentials-credentials` **does** resolve — `reference/operations-api/operations.md` has a `#### Deploy credentials (credentials)` heading, and the Docusaurus build reports no broken anchor for it (the only broken anchors in the build are pre-existing ones on `backups/overview` and `release-notes/5.1`).

The genuine issue: `credentials` is an array of objects, and the CLI's `key=value` args [don't support array-of-object params](https://github.com/HarperFast/documentation/blob/main/reference/cli/operations-api-commands.md#object-parameters). So rather than advertising `credentials=<json>` on the CLI, I've reworked that line to direct private-registry/git deploys to the HTTP Operations API (where the `credentials` array is fully expressible), noting the v5.2.0 availability.

kriszyp (approved):

reference/cli/commands.md
```diff
@@ -92,6 +92,72 @@ harper dev /path/to/app
...
+- `package=<reference>` - An npm, GitHub, or tarball reference to deploy instead of the current directory (e.g. `HarperDB/app#semver:v1.0.0`).
+- `target=<url>` - Remote Harper instance to deploy to. Omit to deploy to the local instance. A bare host defaults to `https://<host>:9925`.
+- `restart=true` or `restart=rolling` - Restart Harper after deploying. Use `rolling` for a staggered, zero-downtime restart across a cluster.
+- `replicated=true` - Replicate the deploy to cluster peers.
+- `install_command=<command>` - Override the install command run for the component.
+- `install_timeout=<ms>` - Maximum time, in milliseconds, to allow the install to run.
+- `install_allow_scripts=true` - Allow npm pre/post-install scripts to run (disabled by default).
+- `deployment_timeout=<ms>` - How long, in milliseconds, a peer waits to receive the replicated payload before failing (default: `120000`). (Added in: v5.2.0)
+- `ignore_replication_errors=true` - Treat a peer that fails to receive the deploy as non-fatal instead of failing the whole operation. (Added in: v5.2.0)
```
cb1kenobi: **Medium: `deployment_timeout` and `ignore_replication_errors` shipped in v5.1.4, not v5.2.0**

Both parameters were added on 2026-06-17 and the earliest release tag containing either commit is `v5.1.4`:

- `deployment_timeout` — `c4800b0e0` "fix(deploy): make peer deployment-row wait configurable, default 120s (#1338)"
- `ignore_replication_errors` — `65d20585c` "deploy_component: fail the deploy on peer replication failures (non-zero exit) (#1334)"

(`git tag --contains` on both returns `v5.1.4` as the first release.) Badging them v5.2.0 tells anyone on a 5.1.x server that a parameter they already have is unavailable. The `host` (v5.2.0, `79325af60`) and `credentials` (v5.2.0, `3dbcf7b9e`) annotations in this same commit are correct — it is only these two that are off.

Suggested fix:
```suggestion
- `deployment_timeout=<ms>` - How long, in milliseconds, a peer waits to receive the replicated payload before failing (default: `120000`). (Added in: v5.1.4)
- `ignore_replication_errors=true` - Treat a peer that fails to receive the deploy as non-fatal instead of failing the whole operation. (Added in: v5.1.4)
```

—
_Generated by Barber AI_

reference/operations-api/operations.md
```diff
@@ -599,6 +599,8 @@ Additional parameters:
 - `host` <VersionBadge version="v5.2.0" /> — the virtual hostname the component is served on (e.g. `"api.example.com"`). Must be a bare hostname or IPv6 literal — no scheme, port, path, or brackets. Persisted alongside `urlPath`.
 - `install_allow_scripts` — set to `true` to allow npm pre/post install scripts (disabled by default)
 - `credentials` — credentials for installing a component from a private npm registry or private git repository (see below)
+- `deployment_timeout` <VersionBadge version="v5.2.0" /> — how long, in milliseconds, a peer waits to receive the replicated deployment payload before failing (default: `120000`)
+- `ignore_replication_errors` <VersionBadge version="v5.2.0" /> — set to `true` to treat a peer that fails to receive the deploy as non-fatal instead of failing the whole operation. By default a failed peer causes `deploy_component` to return a non-2xx status; the component is still deployed (and, if requested, restarted) on the origin node.
```
cb1kenobi: **Medium: same version error as the CLI page — these are v5.1.4 parameters**

`deployment_timeout` (`c4800b0e0`, #1338) and `ignore_replication_errors` (`65d20585c`, #1334) both landed 2026-06-17 and first shipped in `v5.1.4`, not v5.2.0. Since this is the canonical operation reference the CLI page links to, the wrong badge here propagates.

Suggested fix:
```suggestion
- `deployment_timeout` <VersionBadge version="v5.1.4" /> — how long, in milliseconds, a peer waits to receive the replicated deployment payload before failing (default: `120000`)
- `ignore_replication_errors` <VersionBadge version="v5.1.4" /> — set to `true` to treat a peer that fails to receive the deploy as non-fatal instead of failing the whole operation. By default a failed peer causes `deploy_component` to return a non-2xx status; the component is still deployed (and, if requested, restarted) on the origin node.
```

—
_Generated by Barber AI_

reference/cli/commands.md
```diff
@@ -92,6 +92,72 @@ harper dev /path/to/app
...
+- `force=true` - Allow deploying over a protected core component name.
+- `urlPath=<path>` - HTTP path the component is mounted at (e.g. `/api/v2`). Requires `package`.
+- `host=<hostname>` - Virtual hostname the component is served on (e.g. `api.example.com`). Requires `package`. (Added in: v5.2.0)
+- `json=true` - Print output as JSON instead of the default YAML.
+
+Deploying from a private npm registry or git repository requires the `deploy_component` operation's `credentials` field (added in v5.2.0), which is an array of credential objects. The CLI's `key=value` arguments [do not support array-of-object parameters](./operations-api-commands.md#object-parameters), so supply `credentials` through the [Operations API](../operations-api/operations.md#deploy-credentials-credentials) over HTTP instead.
```
cb1kenobi: **Medium: the CLI *does* have a path for private-git deploy credentials**

The array-of-objects reasoning is right, but the conclusion ("use the Operations API over HTTP instead") is only true for a private **npm registry**. For a private **git** source the CLI builds the `credentials` array for you: `bin/cliOperations.ts` accepts `by_ref` / `ref` / `credential` as client-side-only args, and `prepareDeployByRef()` does

```js
if (credentialHost && req.credentials === undefined) {
	req.credentials = [{ host: credentialHost, secret: deriveGitSecretName(req.project, credentialHost) }];
}
```

so `harper deploy by_ref=true credential=true` deploys a pinned private git commit with a sealed credential reference — no HTTP call needed. As written this line sends a reader to the Operations API for a case the CLI handles natively.

Suggested fix: scope the sentence to npm registries and point git users at the `by_ref`/`credential` args, e.g. "Deploying from a **private npm registry** requires the `deploy_component` operation's `credentials` field (added in v5.2.0) … supply it through the Operations API over HTTP instead. For a **private git repository**, use `harper deploy by_ref=true credential=true`, which builds the credential reference for you." (Note that `by_ref`, `ref`, and `credential` are currently undocumented anywhere in this repo — worth their own entry in the parameter list above.)

—
_Generated by Barber AI_

reference/cli/commands.md
cb1kenobi: Withdrawing this — it was wrong. `bin/cliOperations.ts` lists `auth_username`/`auth_password` in `TRANSPORT_ONLY_FIELDS`, with an explicit comment that plain `username`/`password` are payload fields and the `auth_`-prefixed pair is the way to authenticate as a different user. Your updated line (both pairs, legacy noted) matches the source.

—
_Generated by Barber AI_

reference/cli/commands.md
cb1kenobi: Withdrawing this — it was wrong. `host` was added to `deployComponentValidator` by `79325af60` ("feat: route applications by host/urlPath declared in the root config", 2026-07-27), first shipped in `v5.2.0`. The v5.2.0 annotation you added is correct.

—
_Generated by Barber AI_

reference/cli/commands.md
cb1kenobi: Withdrawing this — it was wrong on both counts. `credentials` became a real `deploy_component` field in `3dbcf7b9e` ("Reshape deploy_component registryAuth into a general-purpose credentials array (#1797)", first shipped `v5.2.0`), and `reference/operations-api/operations.md` does have the `#### Deploy credentials (`credentials`)` heading the link targets. One follow-up on the replacement paragraph, posted separately.

—
_Generated by Barber AI_

reference/operations-api/operations.md
cb1kenobi: I confirmed that it should be 5.1.4.

reference/cli/commands.md
cb1kenobi: I confirmed that it should be 5.1.4.

cb1kenobi (changes requested):

reference/cli/commands.md
Ethan-Arrowood: Fixed in 038d7c4f — you're right, and my earlier check was buggy (I filtered `git tag` to `^v5.\d+.0$`, so I only saw minor releases and missed the v5.1.4 patch). `git tag --contains` on both `c4800b0e0` (#1338) and `65d20585c` (#1334) returns v5.1.4 as the first release. Both the CLI page and the canonical operation reference now read `v5.1.4`.

reference/operations-api/operations.md
Ethan-Arrowood: Fixed in 038d7c4f — both badges on the canonical `deploy_component` reference now read `v5.1.4`.

reference/cli/commands.md
Ethan-Arrowood: Good catch — and it's now released, so I've documented it. `by_ref` landed in `92307c745` (first shipped v5.2.3) and `harper deploy setup=true` in `d46ea1f30` / #1851 (also v5.2.3); both are after my original local checkout, which is why I'd steered everything to HTTP.

In 038d7c4f I:

- Added `by_ref`, `ref`, `credential`, and `setup` to the parameter list (v5.2.3).
- Added a **Deploy by reference** section covering the `git+https://…#<sha>` pinned-commit flow.
- Reworked the private-source guidance into a **Private deploy sources** section that splits the two paths: private **git** uses `harper deploy setup=true` then `harper deploy by_ref=true credential=true`; private **npm registry** still goes through the HTTP Operations API (the CLI can't express the nested `credentials` array).

reference/cli/commands.md
```diff
@@ -92,6 +92,111 @@ harper dev /path/to/app
...
+harper deploy package=HarperDB/application-template
+```
+
+**Deploy to a remote instance and restart it afterward**:
+
+```bash
+harper deploy target=https://server.com:9925 restart=true
+```
+
+Remote deploys authenticate the same way as any other remote CLI operation (stored login token, `auth_username`/`auth_password` or the legacy `username`/`password`, or environment variables). See [Remote Operations](./overview.md#remote-operations).
```
dawsontoth, line 125: If we want to get real spicy, we can document the OIDC stuff that just landed. Your Claude can see from the merged PR.

reference/cli/commands.md
```diff
@@ -92,6 +92,111 @@ harper dev /path/to/app
...
+harper deploy by_ref=true ref=v1.2.3
+```
+
+The commit must be pushed to the remote before deploying, so the cluster can clone it; the CLI warns if the working tree is dirty or the commit is not on any remote branch. Only GitHub `origin` remotes are supported.
+
+#### Private deploy sources
+
+<VersionBadge version="v5.2.3" />
+
+Installing a component from a private npm registry or a private git repository requires the `deploy_component` operation's [`credentials`](../operations-api/operations.md#deploy-credentials-credentials) field, an array of credential objects. The CLI's `key=value` arguments [do not support array-of-object parameters](./operations-api-commands.md#object-parameters), so there are two supported paths:
```
cb1kenobi: **Medium: the CLI array-of-object claim is false, so the npm-only-via-HTTP guidance is stricter than it needs to be**

This line (and the "Private npm registry" bullet below at line 198) says "The CLI's `key=value` arguments do not support array-of-object parameters," and routes private npm registry credentials to the HTTP Operations API on that basis. That's contradicted by the current `harper` source:

- `bin/cliOperations.ts` `buildRequest()` (~L602-627) `JSON.parse`s every arg value except `ref` (the sole `RAW_STRING_FIELDS` entry) — the comment at L81-83 says so explicitly, and `credentials='[{"host":"github.com","secret":"x"}]'` parses to a real array of objects (verified locally).
- `bin/deploySetup.ts` itself prints a working example using exactly this shape for git (L286-292): `harper deploy_component ... credentials='[{"host":"github.com","secret":"..."}]' ...`.
- `components/operationsValidation.js`'s `credentials: Joi.array().items(...)` (~L513-520) accepts both git (`host`) and npm (`registry`) shaped entries — nothing in the schema or the CLI parser special-cases npm.
- `deploySetup.ts` already prompts for `provider: 'npm'` (~L164-169) and seals an npm token the same way as git, printing the same `credentials='[...]'` line regardless of provider (L284-285) — the CLI-native flow half-exists for npm already.

This traces back to a pre-existing claim in `reference/cli/operations-api-commands.md`'s "Object Parameters"/"Limitations" sections (out of this diff), which this PR's new text inherits and restates. Worth reconciling rather than propagating further.

Suggested fix: drop the "CLI can't express array-of-objects" framing, and either (a) document that `harper deploy setup=true` (provider: npm) + manually passing `credentials='[...]'` on `harper deploy` works today for private npm registries too, or (b) if there's a reason it's intentionally not supported/recommended for npm (e.g. it's untested/unsupported in practice), say that instead of citing a CLI limitation that doesn't exist.

—
_Generated by Barber AI_

Ethan-Arrowood: You're right — fixed in 6933f14a. I confirmed against `origin/main`:

- `buildRequest()` JSON-parses every arg value except `ref` (`RAW_STRING_FIELDS`), so `credentials='[{"registry":"...","secret":"..."}]'` parses to a real array of objects.
- `deploySetup.ts` supports `provider: 'npm'` (`deriveRegistrySecretName`) as well as github, and prints the same `credentials='[...]'` reference line for both (L285).

So the "CLI can't express array-of-objects" framing was wrong, and I've dropped it. The rewritten **Private deploy sources** section now centers on `harper deploy setup=true` for both private git and private npm — sealing the token client-side rather than putting a raw token on the command line (shell history / process listings / CI logs) — and shows the CLI-native npm deploy passing the printed `credentials='[...]'` array directly, no HTTP detour.

On the root cause: the stale claim lives in `reference/cli/operations-api-commands.md`'s "Object Parameters" / "Limitations" sections (which even contradict that page's own working `ids='[...]'` array example). That's outside this PR's diff — I've stopped restating it here, and I'll file a separate issue to reconcile that page rather than expand this PR's scope.

Ethan-Arrowood: Filed the root-cause follow-up: https://github.com/HarperFast/documentation/issues/639 — reconciling the "Object Parameters" / "Limitations" sections in `reference/cli/operations-api-commands.md` with the actual `buildRequest()` behavior (it JSON-parses every arg value except `ref`, so array-of-objects works and even contradicts that page's own `ids='[...]'` example). Kept it out of this PR to stay scoped to `harper deploy`.

cb1kenobi (approved):

dawsontoth (approved):

PR #627, opened 2026-08-14
Document scoped authentication tokens (inline role)

kriszyp:
Documents **scoped tokens** — `create_authentication_tokens` with an inline `role` object — introduced in HarperFast/harper#2176.

- `reference/security/jwt-authentication.md`: new "Scoped Tokens (Inline Role)" section covering the request/response shape, the super_user gate, attribution-only (non-colliding) username, the operations-API-vs-REST allowlist scope, and the expiry-only (irrevocable) trust model.
- `reference/operations-api/operations.md`: notes the inline-`role` variant on the `create_authentication_tokens` entry, cross-linked to the JWT auth page.

Companion to HarperFast/harper#2176. Core semantics match the reviewer notes there (attribution collisions rejected at mint; allowlist governs the operations API + SQL, table CRUD governs REST; short expiries recommended since scoped tokens can't be revoked before they expire).

### DIFF
--- a/reference/operations-api/operations.md
+++ b/reference/operations-api/operations.md
@@ -539,13 +539,24 @@ Detailed documentation: [JWT Authentication](../security/jwt-authentication.md)
 
 ### `create_authentication_tokens`
 
-Does not require prior authentication. Returns `operation_token` (short-lived JWT) and `refresh_token` (long-lived JWT).
+Does not require prior authentication when called with `username`/`password`. Returns `operation_token` (short-lived JWT) and `refresh_token` (long-lived JWT).
 
 ```json
 {
 	"operation": "create_authentication_tokens",
 	"username": "my-user",
 	"password": "my-password"
+}
+```
+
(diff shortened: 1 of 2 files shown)


DavidCockerill (approved):
Approving. No findings — accurate against the implementation, including the uncomfortable part.

I reviewed and approved harper#2176, so I checked these claims against that code rather than its description: the super_user gate is real and a scoped-token bearer is explicitly blocked from self-minting; `username` is attribution-only with the collision check at mint (the self-mint block exists precisely so the name cannot later resolve to a real user); and **"no refresh token is issued and the token cannot be revoked before expiry"** is correct.

That last sentence is the one that matters most and the one it would have been easiest to leave out. There is no per-token kill switch and no stored record to delete, so the only remedy for a leaked scoped token is rotating the JWT keypair for the entire cluster. Putting that in the reference, in plain terms, is what lets a reader size `expires_in` responsibly.

One optional note: `expires_in` is currently unbounded, so "size it accordingly" is the only limit that exists. If a cap lands upstream, this page is where the ceiling belongs.

PR #628, opened 2026-08-14
Document replication.blobGapReconnectMs and bulk-copy cursor flush options

kriszyp:
Documents the blob-gap watchdog interval added in [harper#2177](https://github.com/HarperFast/harper/pull/2177) (consumed by [harper-pro#701](https://github.com/HarperFast/harper-pro/pull/701)) and the existing bulk-copy cursor flush cadence knobs (`copyCursorFlushBytes`/`copyCursorFlushIntervalMs`, harper-pro#480) that became configurable with the same change. Merge together with those PRs.

### DIFF
--- a/reference/configuration/options.md
+++ b/reference/configuration/options.md
@@ -226,6 +226,9 @@ replication:
 - `securePort` — Secure replication port; _Default_: `9933` (changed from `9925` in v4.5.0)
 - `enableRootCAs` — Verify against Node.js Mozilla CA store; _Default_: `true`
 - `blobTimeout` — Blob transfer timeout (ms); _Default_: `120000`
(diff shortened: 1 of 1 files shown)


DavidCockerill (approved):
Approving. Three lines, and the non-obvious default is right.

**Verified:** `blobGapReconnectMs` does default to the `blobTimeout` value — `replication/replicationConnection.ts:3494` reads `positiveMsOr(env.get('replication_blobGapReconnectMs'), effectiveBlobTimeoutMs)`. A derived default is the easiest thing for a config reference to get wrong, usually by writing the literal it happens to equal today; stating it as "the `blobTimeout` value" is correct now and stays correct after someone changes `blobTimeout`.

Pairing the knob with its trade-off in the same line — lower values heal gaps faster at the cost of more reconnects on a link whose faults never heal — is what someone reaching for this during an incident actually needs. Naming the bulk-copy resume behaviour matters too, since that is the case where the reconnect is not merely a retry but resumes from the last banked cursor. RocksDB scoping on the two `copyCursorFlush*` options is noted rather than left to be discovered.

PR #630, opened 2026-08-17
docs(cli): token credentials in the canonical auth precedence (v5.2.0)

dawsontoth:
Documents the CI/CD token credentials that shipped in **v5.2.0** via [harper#1876](https://github.com/HarperFast/harper/pull/1876) and never reached the reference docs.

Split out of [#599](https://github.com/HarperFast/documentation/pull/599), where this content was blocked behind four unmerged upstream PRs. It has no upstream dependency of its own — the behavior is live in v5.2.0 and has been since 2026-07-31 — so it can be reviewed and merged independently. That answers the first of @Ethan-Arrowood's two sequencing questions on #599.

## Why this was worth separating

`HARPER_CLI_REFRESH_TOKEN` and `HARPER_CLI_OPERATION_TOKEN` are the credentials the CLI is meant to use in a pipeline, and the **Authentication Precedence** list — the one place that states which credential wins — stopped at the saved `harper login` token. A reader following the docs to set up CI had no way to learn the two variables exist, let alone where they rank.

## What changed

**`reference/cli/authentication.md`**

- Bearer tokens slotted into the existing numbered precedence list at **position 5**, above the saved login token and above the legacy `username=`/`password=` fallback. Folded into the canonical list rather than stated as a second, competing paragraph, so the order is given once.
- Added to Method 2's supported-variable list, with the namespace rule: whichever prefix supplies a token owns both halves of it.
- New **Token credentials for CI/CD** subsection covering `harper login --for-ci`, the GitHub Actions step shape, refresh behavior, lifetimes, and the one-refresh-token-per-user warning.

**`reference/cli/commands.md`** — `--for-ci` on `harper login`, including the stdout/stderr split that makes `| gh secret set --env-file -` work without displaying the token.

## Behavior documented that was only visible in the source

These are the parts a reader cannot infer, and the reason the section runs longer than a variable list:

- **Token variables are ignored for local operations.** A local command goes over the domain socket, which the server already trusts via `bypassLocalAuth` — and that bypass is an `else if` on "no Authorization header present". Attaching a bearer token to a local request opts out of the trust and gets validated instead, 401ing on a token minted for another cluster. Since these variables are meant to persist across a whole CI job or a developer's shell, reading them unconditionally would break every local `harper` command in that environment.
- **A token namespace is selected as a unit.** If `HARPER_CLI_OPERATION_TOKEN` could pair with `CLI_TARGET_REFRESH_TOKEN`, commands would run as the first identity until its operation token expired and then silently continue as the second.
- **A set-but-blank namespace reports and falls back** to saved login credentials rather than silently running as whoever last logged in on that machine — the usual shape of a misconfigured CI secret.
- **An env-sourced refreshed token stays in memory.** There is no credentials-file entry to persist it to, unlike a token refreshed from `~/.harperdb/credentials.json`.
- **One refresh-token hash per user.** `createTokens` overwrites `hdb_user.refresh_token` and refresh validation accepts only that hash, so minting a token for CI revokes whatever token that user already held — invisibly, until the other holder's next refresh 401s. This is why the section pushes a dedicated CI user rather than mentioning it as a nicety. Named, independently revocable credentials are [harper#2018](https://github.com/HarperFast/harper/issues/2018).

## Verification

Checked against harper **v5.2.2**, not against the PR that introduced the feature:

| Claim | Source |
| --- | --- |
| Precedence order, local-target gate, namespace unit, blank-namespace fallback | `bin/cliOperations.ts` |
| In-memory vs. persisted refresh | `refreshExpiredOperationToken`, `persistKey` |
| `--for-ci` emits two lines on stdout; fails rather than emitting a half-block | `bin/login.ts` |
| `1d` / `30d` lifetime defaults | `config/configUtils.ts` |
| One refresh-token hash per user | `security/tokenAuthentication.ts` |

`npm run build` and `npm run format:check` are clean. The build reports two broken anchors; both are pre-existing on `main` (verified by building `main` with these changes stashed) and neither is in a file this PR touches.

## Notes for review

- No companion marker: nothing here is gated on an unmerged PR, so it gets an immediate pass once [#629](https://github.com/HarperFast/documentation/pull/629)'s check becomes required.
- Opened as a draft only to match the rest of this stack. Unlike the others there is nothing to wait for — promote it whenever you want it reviewed.
- The v5.2.0 badge on the `--for-ci` heading is a real release, not a placeholder. The rest of the deploy stack has moved to v5.3.0 (see #599).

### DIFF
--- a/reference/cli/authentication.md
+++ b/reference/cli/authentication.md
@@ -208,7 +208,7 @@ Expose the two values to the deploy step and no other credentials are needed:
 
 **Refresh behavior.** The CLI mints an operation token from the refresh token when none is supplied, and again whenever the supplied one has expired. A token refreshed from an environment variable is held in memory for that invocation only — nothing is written to `~/.harperdb/credentials.json`, because there is no file entry for an environment-supplied credential. If the refresh token itself is rejected, the command reports that and exits non-zero rather than falling back to another identity.
 
-**A blank token variable is an error, not a fallback.** If a namespace is set but empty — the usual shape of a misconfigured CI secret — the CLI says so and falls back to saved login credentials. It does not silently run as whoever last logged in on that machine.
+**A blank token variable is reported, then skipped.** If a namespace is set but empty — the usual shape of a misconfigured CI secret — the CLI warns and continues down the precedence list, so the run proceeds under the saved `harper login` token if that machine has one. The warning is the only thing separating this from silently deploying as whoever last logged in, so treat it as a failure signal in CI rather than assuming a blank secret stops the run.
 
 **Lifetimes.** Operation tokens expire after `authentication.operationTokenTimeout` (default `1d`) and refresh tokens after `authentication.refreshTokenTimeout` (default `30d`). The pipeline needs a new refresh token when that window closes.
 
--- a/reference/cli/authentication.md
+++ b/reference/cli/authentication.md
@@ -42,10 +42,13 @@ For remote Operations API commands, the CLI uses the first complete authenticati
 2. Credentials embedded in the `target` URL
 3. `HARPER_CLI_USERNAME` and `HARPER_CLI_PASSWORD` environment variables
 4. Legacy `CLI_TARGET_USERNAME` and `CLI_TARGET_PASSWORD` environment variables
-5. A token saved by `harper login`
-6. `username=` and `password=` operation parameters (legacy fallback)
+5. `HARPER_CLI_OPERATION_TOKEN` and `HARPER_CLI_REFRESH_TOKEN` environment variables, or their legacy `CLI_TARGET_` equivalents — see [Token credentials for CI/CD](#token-credentials-for-cicd)
+6. A token saved by `harper login`
+7. `username=` and `password=` operation parameters (legacy fallback)
 
 Credentials are resolved as a pair and are never combined across sources. An incomplete pair supplied with dedicated authentication parameters or in the target URL causes the command to fail. An incomplete environment-variable pair is skipped with a warning so that a saved login token can still be used.
+
+Entries 5 and 6 authenticate with a bearer token and apply to **remote targets only**. A local operation goes over the domain socket, which the server already trusts, so token environment variables are deliberately ignored there — a token minted for one instance would otherwise be attached to every local `harper` command in that shell and rejected.
 
 Before v5.2.0, `username=` and `password=` operation parameters took precedence over environment variables and saved login tokens. This could authenticate an operation as the wrong user when those fields were part of the operation payload, such as the user being created by `add_user`.
 
@@ -94,8 +97,14 @@ Starting in v5.2.0, a complete environment-variable credential pair takes preced
 - `HARPER_CLI_TARGET` - Sets the default `target` for CLI commands. `CLI_TARGET` is the legacy equivalent.
 - `HARPER_CLI_USERNAME` and `HARPER_CLI_PASSWORD` - Preferred credential pair for the target.
 - `CLI_TARGET_USERNAME` and `CLI_TARGET_PASSWORD` - Lower-priority legacy credential pair.
+- `HARPER_CLI_REFRESH_TOKEN` - Long-lived token the CLI exchanges for a fresh operation token on each run. `CLI_TARGET_REFRESH_TOKEN` is the legacy equivalent.
+- `HARPER_CLI_OPERATION_TOKEN` - A short-lived operation token supplied directly, for callers that mint their own.
 
 Each credential namespace is independent. For example, the CLI never combines `HARPER_CLI_USERNAME` with `CLI_TARGET_PASSWORD`. If either namespace supplies only a username or only a password, that incomplete pair is skipped with a warning.
+
+The same rule holds for tokens, and for the same reason: whichever namespace supplies a token owns both halves of it. If `HARPER_CLI_OPERATION_TOKEN` were allowed to pair with `CLI_TARGET_REFRESH_TOKEN`, commands would run as the first identity until its operation token expired and then silently continue as the second.
+
+For a pipeline, prefer a token over `HARPER_CLI_PASSWORD`: it is scoped to authentication, it can be revoked without changing the account password, and it cannot be used to log in interactively. See [Token credentials for CI/CD](#token-credentials-for-cicd).
 
 **Example `.env` file**:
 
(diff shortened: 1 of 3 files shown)


reference/cli/authentication.md
```diff
@@ -162,6 +171,53 @@ harper add_user \
...
+- name: Deploy
+  run: harper deploy project=my-app restart=true replicated=true
+  env:
+    HARPER_CLI_TARGET: ${{ secrets.HARPER_CLI_TARGET }}
+    HARPER_CLI_REFRESH_TOKEN: ${{ secrets.HARPER_CLI_REFRESH_TOKEN }}
+```
+
+**Refresh behavior.** The CLI mints an operation token from the refresh token when none is supplied, and again whenever the supplied one has expired. A token refreshed from an environment variable is held in memory for that invocation only — nothing is written to `~/.harperdb/credentials.json`, because there is no file entry for an environment-supplied credential. If the refresh token itself is rejected, the command reports that and exits non-zero rather than falling back to another identity.
+
+**A blank token variable is an error, not a fallback.** If a namespace is set but empty — the usual shape of a misconfigured CI secret — the CLI says so and falls back to saved login credentials. It does not silently run as whoever last logged in on that machine.
```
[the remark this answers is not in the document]
dawsontoth: Fixed in 613c2e86 — you were right, and a separate reliability pass over this branch reached the same conclusion independently, so this had two lenses on it.

I went further than the suggested wording. Your replacement keeps the closing sentence "It does not silently run as whoever last logged in on that machine", and that part is also wrong: it *does* run as whoever last logged in, because falling back to the saved `harper login` token is exactly that. The only thing separating it from silent is the warning. Keeping that sentence would have left a reader believing a blank CI secret halts the run.

The paragraph now reads:

**A blank token variable is reported, then skipped.** If a namespace is set but empty — the usual shape of a misconfigured CI secret — the CLI warns and continues down the precedence list, so the run proceeds under the saved `harper login` token if that machine has one. The warning is the only thing separating this from silently deploying as whoever last logged in, so treat it as a failure signal in CI rather than assuming a blank secret stops the run.

Verified against `bin/cliOperations.ts` in harper v5.2.4: the empty-namespace branch logs "Ignoring empty ..._OPERATION_TOKEN/..._REFRESH_TOKEN; falling back to saved login credentials." and then continues to the credentials-file lookup.

_🤖 Addressed by Claude Code (Opus 5)._

reference/cli/authentication.md
```diff
@@ -94,9 +97,17 @@ Starting in v5.2.0, a complete environment-variable credential pair takes preced
...
 
 Each credential namespace is independent. For example, the CLI never combines `HARPER_CLI_USERNAME` with `CLI_TARGET_PASSWORD`. If either namespace supplies only a username or only a password, that incomplete pair is skipped with a warning.
 
+The same rule holds for tokens, and for the same reason: whichever namespace supplies a token owns both halves of it. If `HARPER_CLI_OPERATION_TOKEN` were allowed to pair with `CLI_TARGET_REFRESH_TOKEN`, commands would run as the first identity until its operation token expired and then silently continue as the second.
+
+The namespace is chosen by which one is **set**, not by which one has a usable value — so `HARPER_CLI_REFRESH_TOKEN=` (present but empty) claims the choice and shadows a perfectly good `CLI_TARGET_REFRESH_TOKEN`, which is never consulted. The run then falls through to the saved login token. Unset the preferred variable rather than blanking it.
+
+For a pipeline, prefer a token over `HARPER_CLI_PASSWORD`: it is scoped to authentication, it can be revoked without changing the account password, and it cannot be used to log in interactively. See [Token credentials for CI/CD](#token-credentials-for-cicd).
```
kriszyp: `reference/cli/overview.md:160-168` still tells CI users to use `HARPER_CLI_USERNAME`/`HARPER_CLI_PASSWORD`, omits both token variables, and says the displayed sources are resolved "in the order shown." That leaves the overview inconsistent with this recommendation and the canonical precedence list. Please add token credentials (ideally including the `--for-ci` path) there, or reword the overview as an intentionally abbreviated list without the ordering claim.

reference/cli/authentication.md
```diff
@@ -94,9 +97,17 @@ Starting in v5.2.0, a complete environment-variable credential pair takes preced
 - `HARPER_CLI_TARGET` - Sets the default `target` for CLI commands. `CLI_TARGET` is the legacy equivalent.
 - `HARPER_CLI_USERNAME` and `HARPER_CLI_PASSWORD` - Preferred credential pair for the target.
 - `CLI_TARGET_USERNAME` and `CLI_TARGET_PASSWORD` - Lower-priority legacy credential pair.
+- `HARPER_CLI_REFRESH_TOKEN` - Long-lived token the CLI exchanges for a fresh operation token on each run. `CLI_TARGET_REFRESH_TOKEN` is the legacy equivalent.
+- `HARPER_CLI_OPERATION_TOKEN` - A short-lived operation token supplied directly, for callers that mint their own.
```
kriszyp: The v5.2 CLI also supports `CLI_TARGET_OPERATION_TOKEN`, but this supported-variable list mentions only the legacy refresh-token alias. Could this line name `CLI_TARGET_OPERATION_TOKEN` as the legacy equivalent, matching the preceding entry and the precedence section?

reference/cli/authentication.md
```diff
@@ -94,9 +111,17 @@ Starting in v5.2.0, a complete environment-variable credential pair takes preced
...
 
 Each credential namespace is independent. For example, the CLI never combines `HARPER_CLI_USERNAME` with `CLI_TARGET_PASSWORD`. If either namespace supplies only a username or only a password, that incomplete pair is skipped with a warning.
 
+The same rule holds for tokens: whichever namespace supplies a token owns both halves of it, so an operation token from one namespace is never paired with a refresh token from the other.
+
+The namespace is chosen by which one is **set**, not by which one has a usable value — so `HARPER_CLI_REFRESH_TOKEN=` (present but empty) claims the choice and shadows a perfectly good `CLI_TARGET_REFRESH_TOKEN`, which is never consulted. The run then falls through to the saved login token. Unset the preferred variable rather than blanking it.
+
+For a pipeline, a token is the right style: it is scoped to authentication, it can be revoked without changing the account password, and it cannot be used to log in interactively. Use it _instead of_ a password or payload credentials, not alongside them — see [Token credentials for CI/CD](#token-credentials-for-cicd).
```
kriszyp, line 123: `reference/cli/overview.md:160-168` still tells CI users to use `HARPER_CLI_USERNAME`/`HARPER_CLI_PASSWORD`, omits both token variables, and says the displayed sources are resolved "in the order shown." That leaves the overview inconsistent with this recommendation and the canonical precedence list. Please add token credentials (ideally including the `--for-ci` path) there, or reword the overview as an intentionally abbreviated list without the ordering claim.

reference/cli/authentication.md
```diff
@@ -162,6 +187,72 @@ harper add_user \
...
+
+- **Refresh token only** (what `--for-ci` provisions): no bearer token is attached. Against a remote target the command fails as unauthenticated — unless it also carries `username=` and `password=` operation parameters, in which case it authenticates as that pair instead, a different identity than the one you configured.
+- **Refresh token only, against a loopback target**: it does not fail at all. `authentication.authorizeLocal` defaults to `true`, so a request from `127.0.0.1` or `::1` is authenticated as **superuser** with no credential checked. On a self-hosted runner pointed at its own node, an expired token therefore produces a green, fully privileged run rather than an error. This is the failure mode least likely to be noticed.
+- **An expired operation token as well**: that expired token is still attached, so the request goes out carrying it and the server rejects it with `403` — token expiry answers `403` on the operations path too, since both token types are checked by the same validator. You get a rejection rather than a silent identity switch, but do not alarm on `401` for it.
```
kriszyp: The preceding loopback guidance says `authorizeLocal` authenticates any loopback request as superuser without checking credentials, but this unqualified bullet says attaching an expired operation token makes the server reject the request with 403. Both cannot be true for a loopback target. Please either scope this bullet to non-loopback targets or revise the loopback guidance to explain when an `Authorization` header bypasses local authorization; otherwise self-hosted runners can get the opposite outcome from the documented one.

kriszyp (approved):
Deploys are getting so much better! Thank you!

reference/cli/authentication.md
dawsontoth: Fixed in 1a8dde1a — added token credentials with the `--for-ci` pointer, moved the CI/CD recommendation onto them, and replaced the ordering claim with an explicit note that the list is abbreviated and **not** in precedence order (pointing at the canonical list for the real order).

Thank you for catching this — it is a symmetry miss on my part. I updated `authentication.md` and `commands.md` and never grepped for other pages making precedence claims, which is exactly the check I should have run when the same pattern spans several surfaces. Worth noting the overview also still recommended username/password for CI, which now contradicts the one-style guidance this PR adds, so it was two inconsistencies rather than one.

reference/cli/authentication.md
dawsontoth: You were right that both could not be true, and tracing it produced the fact the whole section was missing. Fixed in 1a8dde1a.

`authentication()` in `security/auth.ts` is an `else if` chain: the local-authorization branch is only reached when a request arrives with **no** `Authorization` header. A request carrying a token — even an expired one — is validated in the first branch and rejected on its merits, loopback or not. So both statements were true, for different cases, and the page never supplied the hinge between them.

Stated explicitly now, and it makes the section coherent rather than just non-contradictory: the dangerous shape is precisely refresh-token-only, because a failed refresh leaves nothing to validate and the request arrives bare; an expired *operation* token is comparatively safe because it still produces a header.

That also forced a third change you did not ask for. The one-style tip said "against a loopback target Harper authenticates by address, so no choice of style protects you there" — which overstates once the real distinction is whether a credential is attached rather than which style it is. Left alone it would have reintroduced your contradiction one paragraph earlier.

Fourteen rounds of automated review circled this distinction without landing it, so: genuinely useful catch.

PR #631, opened 2026-08-18
docs(roles): state the limits of the operations allowlist

dawsontoth:
Companion to HarperFast/studio#1628, which surfaces `permission.operations` in the Studio roles UI. Building that UI meant establishing the field's real semantics against the Harper source, and several of them contradict a natural reading of this page — I got each one wrong first and had them caught in review.

## What changes

The two-gate description reads as universal ("any unlisted operation is denied"), so the obvious next thought is a restricted super user. Three limits are now stated explicitly:

- **`super_user`/`cluster_user` + `operations` is a validation error.** `validateNoSUPerms` rejects any multi-key permission setting either flag. Authorization also returns early for a `super_user` role before reaching the allowlist gate, so even a stored one would not narrow it.
- **`structure_user` bypasses the allowlist for DDL only** — the four table/attribute operations, plus create/drop database for the boolean form; the array form scopes that to its listed databases. Everything else is still gated.
- **`sql` is authorized on its own path** (`checkASTPermissions`/`verifyPermsAST`), which never consults the allowlist. Worth stating plainly because `sql` appears in both `read_only` and `standard_user`, so the groups imply the opposite.

Plus a pointer to HarperFast/harper#2194 for the non-array case, since a role carrying one can stop the instance loading its user cache.

## Verification

Read against harper `v5.2.2` (`utility/operation_authorization.ts`, `validation/role_validation.ts`, `security/user.ts`, `server/serverHelpers/serverUtilities.ts`); the `sql` and `structure_user` claims were also exercised against the compiled `dist/`. No behavior change here — this documents what 5.x already does.

## Review coverage

Written by an LLM (Claude Fable 5). Prose-only change to one file; `prettier --check` clean. The underlying findings were reviewed by @cb1kenobi and @kriszyp on studio#1628.

### DIFF
--- a/reference/users-and-roles/overview.md
+++ b/reference/users-and-roles/overview.md
@@ -79,6 +79,14 @@ The `operations` field in a permission object restricts which Operations API cal
 
 Operations normally restricted to `super_user` can be selectively granted by including them in the list. If `operations` is not set, the role can call any non-`super_user` operation, subject to table CRUD permissions.
 
+The field scopes an ordinary role; it is not a way to narrow an administrator. Three limits follow from that, and each one surprises people:
+
+- **`super_user` and `cluster_user` roles cannot carry an allowlist at all.** `add_role` and `alter_role` reject any permission that sets either flag alongside other keys, so the combination is a validation error rather than a restricted admin. Authorization also clears a `super_user` role before the allowlist is consulted.
+- **`structure_user` roles bypass the allowlist for DDL only.** `create_table`, `create_attribute`, `drop_table`, and `drop_attribute` are reachable regardless of the list — plus `create_database`/`drop_database` when `structure_user` is `true`. When it is an array of database names, that carve-out applies only to those databases. Every other operation is still gated normally.
+- **`sql` is not gated by this field.** SQL statements are authorized against table CRUD permissions on their own path, so listing `sql` neither grants nor restricts them, and omitting it does not prevent a role from running SQL. This matters when reading the `read_only` and `standard_user` groups below, both of which include the name.
+
+The value must be an array of strings. A non-array value is rejected on write, and a role that already holds one (for example a pre-5.0 role that granted a database named `operations`) can prevent the instance from loading its user cache — see [HarperFast/harper#2194](https://github.com/HarperFast/harper/issues/2194).
+
 **Permission Groups**
 
 Groups expand to a predefined set of operations and can be mixed with individual operation names:


reference/users-and-roles/overview.md
```diff
@@ -79,6 +79,14 @@ The `operations` field in a permission object restricts which Operations API cal
 
 Operations normally restricted to `super_user` can be selectively granted by including them in the list. If `operations` is not set, the role can call any non-`super_user` operation, subject to table CRUD permissions.
 
+The field scopes an ordinary role; it is not a way to narrow an administrator. Three limits follow from that, and each one surprises people:
+
+- **`super_user` and `cluster_user` roles cannot carry an allowlist at all.** `add_role` and `alter_role` reject any permission that sets either flag alongside other keys, so the combination is a validation error rather than a restricted admin. Authorization also clears a `super_user` role before the allowlist is consulted.
```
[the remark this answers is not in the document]
dawsontoth: Agreed the wording was misleading, and the sentence is gone rather than reworded.

Two reasons. It described a state `add_role`/`alter_role` will not store — `validateNoSUPerms` rejects `super_user: true` alongside any other permission key — so what authorization would do with such a role was never reachable. And [harper#2176](https://github.com/HarperFast/harper/pull/2176) has since inverted it: `verifyOperationsAllowlist` now runs *before* the `super_user` early return, so the allowlist is consulted first rather than bypassed. Your suggested phrasing was the correct description of 5.2.x, but it would have been wrong in 5.3.0.

The bullet now states only the durable fact — the combination is a validation error — and says to grant the operations to an ordinary role instead.

reference/users-and-roles/overview.md
```diff
@@ -79,6 +79,14 @@ The `operations` field in a permission object restricts which Operations API cal
...
+- **`structure_user` roles bypass the allowlist for DDL only.** `create_table`, `create_attribute`, `drop_table`, and `drop_attribute` are reachable regardless of the list — plus `create_database`/`drop_database` when `structure_user` is `true`. When it is an array of database names, that carve-out applies only to those databases. Every other operation is still gated normally.
+- **`sql` is not gated by this field.** SQL statements are authorized against table CRUD permissions on their own path, so listing `sql` neither grants nor restricts them, and omitting it does not prevent a role from running SQL. This matters when reading the `read_only` and `standard_user` groups below, both of which include the name.
+
+The value must be an array of strings. A non-array value is rejected on write, and a role that already holds one (for example a pre-5.0 role that granted a database named `operations`) can prevent the instance from loading its user cache — see [HarperFast/harper#2194](https://github.com/HarperFast/harper/issues/2194).
```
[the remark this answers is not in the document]
dawsontoth: Right about the convention, but not the right version here — declining the `v5.0.0` badge and adding a `v5.3.0` one instead.

`reference/` is the v5 tree (`lastVersion: 'current'`, archived v4 lives in `reference_versioned_docs/version-v4/`), so a `v5.0.0` badge marks a change that is true of every statement on the page — this file carries no badges at all for that reason. I confirmed the premise, though: `OPERATIONS_KEY` is already reserved in `validation/role_validation.js` at `v5.0.0`, so the reservation did land there. That v4 provenance is now explicit in the prose instead of implied by a badge:

> A role that already holds one — for example a pre-5.0 role that granted a database named `operations`, before the key became reserved — can prevent the instance from loading its user cache

The badge is on the change that actually falls *within* v5: [harper#2176](https://github.com/HarperFast/harper/pull/2176) moved allowlist gate one ahead of the `structure_user` carve-out and gave the SQL path its own call into it, so `<VersionBadge type="changed" version="v5.3.0" />` now sits on that sentence. Derived from tags per the repo guideline — #2176 is merged and `git tag --contains` returns only `v5.3.0-alpha.1`, with `v5.2.13` the latest stable.

reference/users-and-roles/overview.md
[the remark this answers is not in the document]
dawsontoth: Dropped the sentence entirely rather than rewording it: it described a state `validateNoSUPerms` will not store, and harper#2176 has since inverted the ordering it claimed. Detail in my reply above.

reference/users-and-roles/overview.md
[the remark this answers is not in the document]
dawsontoth: Declined the `v5.0.0` badge (all of `reference/` is v5, so it marks nothing) and put the v4 provenance in the prose instead; added a `v5.3.0` badge for the change that does fall within v5. Detail in my reply above.

dawsontoth:
Ready for review — but **please do not merge this yet**, even with an approval.

`reference/` is the published `/reference/v5` site, and this documents the allowlist ordering from [harper#2176](https://github.com/HarperFast/harper/pull/2176), which is merged to `harper` `main` but tagged only in `v5.3.0-alpha.1`. Latest stable is `v5.2.13`, so merging now would publish behavior no released Harper has. Labelled `waiting-for-companion` for that, though the literal wait here is on the 5.3.0 **release** rather than on a companion PR merging — the companion-check gate would pass immediately, since #2176 has already merged.

Please also do not arm auto-merge. Happy to flip it back to draft if that is a safer place to park it while it waits.

Worth a careful look at two things:

1. The `v5.3.0` badge version. Derived from tags per the repo guideline: #2176 is merged, `git tag --contains` returns only `v5.3.0-alpha.1`, and `v5.2.13` still has the old ordering. If 5.3.0 slips such that this lands in a later minor, the badge and the release-note file both need moving.
2. The claim that listing `sql` admits any statement variant, with table permissions deciding what it may touch. That is from `DESIGN.md` — "a direct `sql` call has no outer job op, so there the `operations` allowlist alone is the operation-invocation check" — plus `serverUtilities.ts:362` passing `json.operation`. It is the load-bearing claim in the example, and the previous version of this PR had it backwards, so it is the one I would most like a second pair of eyes on.

reference/users-and-roles/overview.md
```diff
@@ -79,6 +79,40 @@ The `operations` field in a permission object restricts which Operations API cal
 
 Operations normally restricted to `super_user` can be selectively granted by including them in the list. If `operations` is not set, the role can call any non-`super_user` operation, subject to table CRUD permissions.
 
+<VersionBadge type="changed" version="v5.3.0" />
+
+Gate one admits nothing the list omits, whatever else the role carries. Earlier v5 releases let table DDL and SQL around it; both now go through it.
```
Devin-Holland: This universal rule conflicts with the MCP guidance in `reference/mcp/tools-and-resources.md:46-50`, which says `tools/list` exposes schema operations to `structure_user` in addition to those in `permission.operations` and claims the result contains only operations the user can actually call. In v5.3, a role with `structure_user` that omits `create_table` can therefore be shown that tool even though invocation is denied by this newly documented gate. Please update the MCP section to distinguish tool visibility from call authorization, or align its filtering contract with the v5.3 rule, so users are not given contradictory expectations.

reference/users-and-roles/overview.md
```diff
@@ -79,6 +79,40 @@ The `operations` field in a permission object restricts which Operations API cal
 
 Operations normally restricted to `super_user` can be selectively granted by including them in the list. If `operations` is not set, the role can call any non-`super_user` operation, subject to table CRUD permissions.
 
+<VersionBadge type="changed" version="v5.3.0" />
```
kriszyp: The first affected release is `v5.2.5`, not `v5.3.0`. I compared the tagged core sources: `v5.2.4` has neither the allowlist-before-`structure_user` ordering nor the direct-SQL allowlist check, while both are present in the annotated `v5.2.5` tag; the implementation commit is contained by that tag. With this badge and the warning under the 5.3 release notes, operators upgrading 5.2.4 -> 5.2.5 will not be told to add `sql`/DDL operations before access is revoked. Please change this badge to `v5.2.5` and remove or relocate the section at `release-notes/v5-lincoln/5.3.md:35` so the breaking warning is attached to the release that introduced it. I could not compare the local tag objects to origin because network access is blocked in this review environment.

reference/users-and-roles/overview.md
kriszyp: The invocation path enforces this, but MCP discovery does not yet follow the same invariant. In tagged and current core, both MCP visibility helpers return true for schema-structure operations as soon as `structure_user` is truthy, before consulting `permission.operations`. Thus `{ structure_user: ['orders_db'], operations: ['sql'] }` is advertised create/drop MCP tools that every call will reject. That contradicts `reference/mcp/tools-and-resources.md:46-50`, which promises `tools/list` shows only operations a user can call. Please fix the root cause in core by giving both visibility helpers the same allowlist-first ordering. If core intentionally keeps this behavior, update the MCP reference in this PR to describe the advertised-but-unusable tools and link the tracking issue.

ldt1996 (commented):
Verified the substantive claims against the Harper source on current main rather than taking the description's word: verifyOperationsAllowlist exists with exactly the gate-one semantics documented here, its own comment confirms the SQL path must call it directly (serverUtilities.ts does, ahead of checkASTPermissions), api_name resolution matches the "written in snake_case" framing, and validateNoSUPerms plus OPERATIONS_MUST_BE_ARRAY back the role-shape statements. The example role is consistent with those rules, and the release note's two audit cases are the right upgrade caveats. Content-wise this is accurate and I would approve it.

Two process items:

1. The description says this is held as a draft until 5.3.0 is tagged, but the PR is currently marked ready for review. Only v5.3.0-alpha.1 exists as of today, and reference/ publishes as the live v5 site, so as it stands this is one merge click away from documenting unshipped behavior as current. Please flip it back to draft (or land whatever hold mechanism replaces it) until the stable tag exists; I am deliberately not approving while it is in a mergeable state its own body forbids.

2. Small wording nit, take or leave: "Gate one admits nothing the list omits" leans on internal gate numbering that the published page never introduces. Readers of the reference site have no gate one and gate two; "The allowlist admits nothing it omits, whatever else the role carries" says the same thing without the internal vocabulary.

Ping me when it is re-drafted or 5.3.0 tags and I will approve.

Lavinia, via Claude

dawsontoth:
Thanks for verifying against the source rather than the description — that is the check that matters most here, since the first version of this PR got two of these three claims backwards.

**On the wording nit (2): taken, with a note on the premise.** The page does introduce the numbering — the paragraph above this one says "it acts as a two-gate check" and its second item says "for data operations that pass gate one" — so the vocabulary is not unestablished. But leaning on it in the sentence that states the rule still makes the reader carry a label forward to get the point, so I dropped it in `f5e6ad77`:

> An operation the list omits is denied whatever else the role carries — the list is checked ahead of every other permission on the role.

That also says *where* the check happens rather than which numbered gate it is, and matches the wording the 5.3 release note already uses.

**On the draft state (1): agreed on the substance, but that is not my call to reverse.** Marking it ready was an explicit instruction from the PR author after I had flagged exactly the risk you are describing, so I have surfaced your request to them rather than flipping it back myself. You are right that the label and the body warning are a weaker gate than draft state was.

Will ping you once that is resolved, or when 5.3.0 tags.

reference/users-and-roles/overview.md
kriszyp: The first affected release is `v5.2.5`, not `v5.3.0`. The tagged core sources establish that v5.2.4 lacks both the DDL/SQL allowlist enforcement and v5.2.5 contains them. Keeping this badge, with the corresponding warning at `release-notes/v5-lincoln/5.3.md:35`, leaves 5.2.4 -> 5.2.5 upgrades without an alert to list `sql` and DDL operations before access is revoked. Please badge this `v5.2.5` and relocate the release-note warning to the 5.2.5 notes.

reference/users-and-roles/overview.md
```diff
@@ -79,6 +79,40 @@ The `operations` field in a permission object restricts which Operations API cal
...
+
+An operation the list omits is denied whatever else the role carries — the list is checked ahead of every other permission on the role. Earlier v5 releases let table DDL and SQL around it; both now go through it.
```
kriszyp: `structure_user` schema operations remain an exception in MCP discovery: both visibility helpers advertise them whenever `structure_user` is truthy before consulting `permission.operations`. A role such as `{ structure_user: ['orders_db'], operations: ['sql'] }` therefore lists create/drop MCP tools but rejects every invocation. That conflicts with this stated invariant and with `reference/mcp/tools-and-resources.md:46-50`, which promises that discovery shows only callable operations. Please fix the root cause in core by making both visibility helpers require the operation allowlist too, rather than documenting around the discrepancy.

Devin-Holland: This universal rule conflicts with the MCP guidance in `reference/mcp/tools-and-resources.md:46-50`, which says `tools/list` exposes schema operations to `structure_user` in addition to those in `permission.operations` and claims the result contains only operations the user can actually call. In v5.3, a role with `structure_user` that omits `create_table` can therefore be shown that tool even though invocation is denied by this newly documented gate. Please update the MCP section to distinguish tool visibility from call authorization, or align its filtering contract with the v5.3 rule, so users are not given contradictory expectations.

dawsontoth:
@ldt1996 — withdrawing the hold, on the author's call. @dawsontoth: since the content is tagged `v5.3.0` and an alpha is available, this can merge once approved, which is what justifies the ready state.

The badge is what carries the disambiguation. A reader on 5.2.x sees "Changed in: v5.3.0" and knows the ordering described is not what their instance does, which is the same contract every other pre-tag 5.3 page on the site is already published under.

That precedent is the part I had wrong when I asked you to hold it. `main` already publishes a full 5.3 release-notes page and seven `reference/` pages carrying `<VersionBadge type="changed" version="v5.3.0" />`:

```
reference/cli/commands.md
reference/configuration/options.md
reference/database/compaction.md
reference/http/overview.md
reference/operations-api/operations.md
reference/security/certificate-verification.md
reference/security/overview.md
```

So the bar I proposed was stricter than the one this repo actually applies, and holding this one PR to it would have made it the outlier. The description is updated accordingly, and the `waiting-for-companion` label is removed.

Your content verification and the gate-one wording fix both still stand — nothing about the substance changed. Over to you for approval whenever you are ready.

To be explicit: I am not merging this myself, and auto-merge is not armed.

kriszyp (approved):
Looks good, couple things to address.

reference/users-and-roles/overview.md
dawsontoth: Not seeing this — `v5.2.4` and `v5.2.5` are byte-identical on both files that would carry the enforcement:

```
$ git diff --stat v5.2.4 v5.2.5 -- utility/operation_authorization.ts \
    server/serverHelpers/serverUtilities.ts
(no output)
```

And the enforcement is absent from the whole 5.2 line, not just 5.2.4. At `v5.2.13`:

- `verifyOperationsAllowlist` does not exist at any 5.2.x tag (`v5.2.3` through `v5.2.13` all absent).
- The SQL path calls `sql.checkASTPermissions(json, parsedSqlObject)` with no allowlist consultation — `expandOperationsPerms` appears zero times in `serverUtilities.ts`.
- In `verifyPerms`, the `structure_user` carve-out is still at line 604 and the `operations` gate at line 628, so DDL continues to bypass the list.

The change is [harper#2176](https://github.com/HarperFast/harper/pull/2176), merged to `main` 2026-08-21 and tagged only in `v5.3.0-alpha.1`. I checked for a back-port by file content rather than ancestry, since a cherry-pick would defeat `git tag --contains` — that is the failure mode your reading would fit, but the files show no cherry-pick happened.

So a 5.2.4 to 5.2.5 upgrade revokes nothing, and a warning in the 5.2.5 notes would tell operators to change roles for a tightening their release does not contain. Keeping the badge at `v5.3.0`. Happy to look again if you have a specific 5.2.5 behavior in mind that I am reading past.

reference/users-and-roles/overview.md
dawsontoth: Confirmed, and it reproduces exactly as you describe. Both helpers short-circuit on `structure_user` truthiness before reaching the list:

```js
// components/mcp/toolRegistry.ts:413 — canRoleInvokeOperation
if (perm.structure_user && SCHEMA_STRUCTURE_OPERATIONS.has(operation)) return true;
if (Array.isArray(perm.operations) && perm.operations.includes(operation)) return true;
```

`components/mcp/resources.ts:918` is the same two lines. `SCHEMA_STRUCTURE_OPERATIONS` holds all eight create/drop schema, table and attribute ops, and an array `structure_user` is truthy, so a role of `{ structure_user: ["orders_db"], operations: ["sql"] }` advertises all eight and gate one denies all eight at dispatch.

One correction to the framing: this does not conflict with the invariant on **this** page. The sentence here is that an omitted operation is *denied* — which is what happens, as your own "rejects every invocation" confirms. `canRoleInvokeOperation` is a discovery filter, and its docblock is explicit that it is "intentionally conservative" and defers per-target checks to `verifyPerms`. What #2176 broke is that the deferral is no longer only about per-target predicates, so the filter is now wrong at the role level too.

The promise it does contradict is on the other page — `reference/mcp/tools-and-resources.md` says `tools/list` is filtered "so each session sees only the operations its user can actually call". Note the bullet just below it currently reads that a `structure_user` sees schema ops "in addition to anything in `permission.operations`", which documents the over-advertising as intended, so that page needs a pass either way once core changes.

On the fix itself: that is a core change in `HarperFast/harper`, outside what a docs PR can carry, so I am not making it here and have surfaced it to @dawsontoth rather than acting on it myself. Worth its own issue — the invocation side is already correct and fails closed, so this is a discovery-accuracy bug rather than an access one.

dawsontoth: @kriszyp Would you consider this to block this PR?

kriszyp (approved):
Great!

PR #633, opened 2026-08-20
Document record-structure dictionary counts on describe_table

kriszyp:
Companion to the core change for [HarperFast/harper#2220 — Per-table msgpackr structure dictionaries grow unbounded and unobservably](https://github.com/HarperFast/harper/issues/2220), which adds four fields to `describe_table`: `typed_structures_enabled`, `typed_structure_count`, `typed_structure_limit`, `classic_structure_count`.

Documents what a record structure is, why the dictionary only ever grows, what reaching the bound means, and how an application keeps it small.

## For the human reviewer

- **The growth model is the part worth checking.** A structure is minted per distinct *shape* = ordered field list × each field's value width class — so `{a, b}` and `{b, a}` are different shapes, and so are `{v: 1}` and `{v: 70000}`. This is measured against the installed deps, not inferred: 4 key orders → 4 structures; one two-field shape across 11 value kinds → 5 structures. The issue that prompted this attributed growth to realized field subsets alone, which understates it.
- **`typed_structures_enabled` needs the surrounding sentence.** `storage.randomAccessFields` defaults off, so most tables show `typed_structure_count: 0` against a limit of 256. Without the explanation an operator reads that as spare headroom rather than the feature being disabled, so the paragraph saying so is load-bearing.
- **The warning's scope is stated honestly rather than simply.** It fires per encoder from the post-save path, so several workers can each warn for one table and a worker that never saves after loading a full dictionary stays silent. The text says the counts are the reliable signal and the log is the heads-up, rather than implying one warning per table.
- **No analytics metric is documented**, deliberately. The core change originally folded these counts onto the `table-size` metric; that metric turns out to have no RocksDB emitter at all (HarperFast/harper#2249), so documenting it here would point operators at a series that never appears.

## Verification

Prettier clean. Claims cross-checked against the core diff and against live measurements from the core PR's integration suite (a saturated table reports `typed 256/256`, `classic 32`).


reference/operations-api/operations.md
```diff
@@ -69,6 +69,35 @@ Returns the definition of a specific table.
 { "operation": "describe_table", "table": "dog", "database": "dev" }
 ```
 
+Alongside the schema, the response carries the size of the table's **record-structure dictionaries**
```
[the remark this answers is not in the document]
kriszyp: Good catch — added `<VersionBadge type="changed" version="v5.2.5" />` in 5ec5070.

On the two judgment calls: `type="changed"` is right per AGENTS.md since `describe_table` itself already exists and only its response gains fields (same treatment as the `get_backup` RocksDB format change). Patch granularity rather than `vX.Y.0` because `main` is already at 5.2.4, so this lands in 5.2.5 — and patch-level badges have precedent in `reference/` (v4.7.2, v5.1.5, v5.1.13, v5.1.15).

— KrAIs (Claude Opus 5)

reference/operations-api/operations.md
```diff
@@ -69,6 +69,37 @@ Returns the definition of a specific table.
...
 ```
 
+<VersionBadge type="changed" version="v5.2.5" />
+
+Alongside the schema, the response carries the size of the table's **record-structure dictionaries**
+— the physical record layouts Harper has seen for this table:
+
+| Field                      | Description                                                                          |
+| -------------------------- | ------------------------------------------------------------------------------------ |
+| `typed_structures_enabled` | Whether random-access (typed) encoding is enabled for this table                     |
```
dawsontoth, line 82: Do we document somewhere the use cases for having this turned on? This is my first time seeing this feature in Harper, and it reads very technically. I don't walk away from it with a clear understanding of the benefits, when to turn it on, what data sets would benefit from it, that kind of thing...

kriszyp: You're right, and the honest answer to "do we document somewhere the use cases" was no — nowhere. `storage.randomAccessFields` shipped in v5.1.0 and had no entry in the configuration option list, no mention on the `@table` directive, and nothing explaining what the encoding buys. This `describe_table` paragraph was a reader's first encounter with the feature, which is exactly why it read as technical with no way in.

Fixed in 458332b by documenting the feature itself:

- **[Storage Tuning → Record Encoding](https://github.com/HarperFast/documentation/blob/kris/2220-structure-metric/reference/database/storage-tuning.md#storagerandomaccessfields)** — the new home for the concept. What the two encodings actually do (both keep field names in a shared structure; the typed one additionally fixes each field's type and width so a record becomes fixed-width slots you can index into, instead of values you have to decode in order), what that buys, then an explicit *Enable it when* / *Leave it off when* pair. Short version: large records where requests touch only a few fields, mostly scalars, consistently shaped writes — and leave it off for wide, sparse, or variably typed schemas, small records, or whole-record reads.
- **[`storage.randomAccessFields`](https://github.com/HarperFast/documentation/blob/kris/2220-structure-metric/reference/configuration/options.md#storage)** in the config option list, and **[`@table(randomAccessFields:)`](https://github.com/HarperFast/documentation/blob/kris/2220-structure-metric/reference/database/schema.md#randomaccessfields)** on the schema directive, both linking to the section above.
- This paragraph now opens with a pointer to it, so the counts are read against a feature the reader has been introduced to.

Two things fell out of writing it that were worth fixing here as well: "shape" was defined as field order plus value *width*, but a structure keys on the encoded *type* too — a field written as `0` in one record and `"ok"` in the next mints a new one, and the remediation advice inherited the omission. And `classic_structure_count` shipped with no interpretation; it stops at 32 (msgpackr's shared named-record list defaults to `maxSharedStructures: 32`, confirmed empirically against the installed dep and matching the core PR's `classic 32` measurement), which is why the response carries no limit field for it.

One thing I did not do: `storage.randomAccessFields` and the `@table` argument are v5.1.0 surface with no entry in `release-notes/v5-lincoln/5.1.md`. Backfilling a shipped release-notes file seemed like a separate call rather than something to fold into this PR — happy to if you'd like it here.

— KrAIs (Claude Opus 5)

dawsontoth (approved):

